PR #1, opened 2024-05-10
Add CI

amaslenn:
1. Add initial CI using GitHub Actions.
2. Add `.gitignore` file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,55 @@
+name: CI
+
+on:
+  pull_request:
+
+jobs:
+
+  lint:
+    name: Linting
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
+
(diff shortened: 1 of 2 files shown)


x41lakazam:
Hey,

I think the names of the workflows should follow a convention

For example see [SSW rules](https://www.ssw.com.au/rules/workflow-naming-scheme/)

amaslenn:
> I think the names of the workflows should follow a convention

Interesting link, thanks for sharing!

Could you please suggest such a name? Personally, I see `ci.yml` as "this is the main pipeline for all PRs", but will be happy to make it easy to understand for everyone.

x41lakazam:
In my opinion there shouldn't be any "main" PR workflow as it should stay modular and workflows should have specific jobs and let the possibility to be trigger by any event independently
To me this file should be split into `lint.yml` and `test.yml`

amaslenn:
> In my opinion there shouldn't be any "main" PR workflow as it should stay modular and workflows should have specific jobs and let the possibility to be trigger by any event independently To me this file should be split into `lint.yml` and `test.yml`

At this point it looks like a complication without adding much value. Yet we might want to follow this modular approach if pipelines will grow in size and complexity.

TaekyungHeo (approved):

PR #5, opened 2024-05-14
Add DEV_GUIDE.md

TaekyungHeo:
## Summary
Add DEV_GUIDE.md

## Test Plan
Not needed.

### DIFF
--- a/DEV_GUIDE.md
+++ b/DEV_GUIDE.md
@@ -1,0 +1,145 @@
+## Introduction
+CloudAI allows users to package workloads as test templates to facilitate the automation of running experiments. This method involves packaging workloads as docker images, which is one of several approaches you can take with CloudAI. This guide details how to package a workload as a docker image and then create a test template in CloudAI.
+
+### Step 1: Creating a Docker Image
+1. **Set Up the GitLab Repository**
+   Start by setting up a repository on GitLab to host your docker image. For this example, use `gitlab.com/cloudai/nccl-test`.
+
+2. **Writing the Dockerfile**
+   The dockerfile needs to specify the base image and detail the steps to add and compile the new NCCL test:
+   ```dockerfile
+   FROM nvcr.io/nvidia/pytorch:24.02-py3
+   COPY nccl_bisection.c .
+   RUN mpicc -o /usr/local/bin/bisection_mpi -I${CUDA_HOME}/include -L${CUDA_HOME}/lib64 nccl_bisection.c -lnccl -lcudart
+   RUN rm nccl_bisection.c
+   ```
+
+3. **Build and Push the Docker Image**
+   Build the docker image with the dockerfile and upload it to the designated repository:
+   ```bash
+   docker build -t gitlab.com/cloudai/nccl-test .
+   docker push gitlab.com/cloudai/nccl-test
+   ```
+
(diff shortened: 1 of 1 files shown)


DEV_GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+        "NcclTest",
+        "NcclTestGradingStrategy",
+        "NcclTestReportGenerationStrategy",
+        "NcclTestSlurmCommandGenStrategy",
+        "NcclTestSlurmInstallStrategy",
+        ...
+    ]
+    ```
+
+### Testing Procedures
```
amaslenn, line 120: Do you think we can make it a `pytest`?

DEV_GUIDE.md
```diff
@@ -0,0 +1,145 @@
...
+      --agg_iters 1 \
+      --average 1 \
+      --parallel_init 0 \
+      --check 1 \
+      --blocking 0 \
+      --cudagraph 0 \
+      --stepfactor 2
+   ```
+
+## Setting Up the Test Template in CloudAI
```
amaslenn, line 50: IMO we should start with a higher level overview and focus on implementing custom flows similar to our CLI:

```markdown
# Implementing a custom execution flow

```py
# Create a System, TestTemplate and TestScenario objects
system, test_templates, test_scenario = System(), [TestTemplate()], TestScenario()

# One way to do that is to use the Parser class
parser = Parser(<system_config_path>, <test_template_path>, <test_path>, <test_scenario_path>)
system, test_templates, test_scenario = parser.parse()

# Update the system object using relevant SystemObjectUpdater. This is necessary to update the system object
# with the correct system configuration.
system_object_updater = SystemObjectUpdater()
system_object_updater.update(system)

# Check if test templates are installed
installer = Installer(system)
if not installer.is_installed(test_templates):
    # raise an exception or
    installer.install(test_templates)

# Create a Runner object and run the test scenario. Use asyncio.run() to run the async function.
runner = Runner(args.mode, system, test_scenario)
asyncio.run(runner.run())

# Generate a report...
generator = ReportGenerator(runner.runner.output_path)
generator.generate_report(test_scenario)

# ... and grade the test scenario
grader = Grader(runner.runner.output_path)
grader.grade(test_scenario)

```

We might also need a class diagrams, but let's postpone that till we figure our the interface of Core API.

TaekyungHeo: It makes perfect sense. I have written this DEV_GUIDE.md to inform others on how to add a new test template. I understsand this is not complete, and we need more detailed guidelines and documentation. Let's improve the document gradually.

cc: @srinivas212

DEV_GUIDE.md
TaekyungHeo: Maybe I am not experienced enough to set it up as a pytest. I have several concerns regarding testing these with pytest. Installation involves interaction with the file system and remote docker repositories. Running involves interaction with the cluster. Therefore, with pytest, it must be challenging to validate. Others will work with pytest.

DEV_GUIDE.md
TaekyungHeo: Sharing the discussion that I had with Andrei via a call. Let's postpone the developer documentation for a while. While it's important to have it, we need additional refactoring and design updates. Let's keep it high level.

PR #11, opened 2024-05-16
Rename default sbatch script to include project-specific prefix

TaekyungHeo:
## Summary
Rename default sbatch script to include project-specific prefix

## Test Plan
Pytest passing.


tests/schema/system/slurm/strategy/test_slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,44 @@
...
+        default_partition="main",
+        partitions={"main": [SlurmNode(name="node1", partition="main", state=SlurmNodeState.IDLE)]},
+    )
+    env_vars = {"TEST_VAR": "VALUE"}
+    cmd_args = {"test_arg": "test_value"}
+    strategy = SlurmCommandGenStrategy(slurm_system, env_vars, cmd_args)
+    return strategy
+
+
+@pytest.mark.usefixtures("strategy_fixture")
```
amaslenn, line 24: [minor] You don't need `@pytest.mark.usefixtures("strategy_fixture")` since you are explicitly referencing `strategy_fixture` in the test function args.

I think we can make this test simpler:
1. Do not mock `open` and `os.path.join`. Since we can pass `output_path`, we can utilise `pytest`'s internal fixture `tmppath: Path`. You can convert it into a string with `str(tmppath)`.
2. After calling `_write_sbatch_script`, check that required file exists under `tmppath`.
3. Additionally, you can also check its content. At least for the arguments you are passing into the function.

amaslenn (changes requested):

src/cloudai/schema/system/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -189,7 +189,7 @@ def _write_sbatch_script(self, args: Dict[str, Any], env_vars_str: str, srun_com
 
         batch_script_content.extend(["", env_vars_str, "", srun_command])
 
-        batch_script_path = os.path.join(output_path, "sbatch_script.sh")
+        batch_script_path = os.path.join(output_path, "cloudai_autogenerated_sbatch_script.sh")
```
srinivas212, line 192: remove _autogenerated? Just cloudai_sbatch_script.sh

tests/schema/system/slurm/strategy/test_slurm_command_gen_strategy.py
TaekyungHeo: That makes sense. Thank you! Updated. I wanted to check the file name only because this pull request concerns the filename.

src/cloudai/schema/system/slurm/strategy/slurm_command_gen_strategy.py
TaekyungHeo: Updated.

PR #14, opened 2024-05-16
Refactor SlurmSystem methods for bug fix and add unit tests

TaekyungHeo:
## Summary
Refactor SlurmSystem methods for bug fix and add unit tests

## Test Plan
1. pytest
2. Ran the nccl-test & nccl-miner on our internal servers.

### DIFF
--- a/src/cloudai/schema/system/slurm/slurm_system.py
+++ b/src/cloudai/schema/system/slurm/slurm_system.py
@@ -15,7 +15,7 @@
 import getpass
 import logging
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from cloudai.schema.core import System
 from cloudai.util import CommandShell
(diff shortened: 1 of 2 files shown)


src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -493,61 +493,99 @@ def scancel(self, job_id: int) -> None:
         """
         self.cmd_shell.execute(f"scancel {job_id}")
 
-    def update_node_states(self) -> None:  # noqa: C901
+    def update_node_states(self, squeue_output: Optional[str] = None, sinfo_output: Optional[str] = None) -> None:
```
amaslenn, line 496: I suggest we make arguments non-optional. Caller should prepare it in advance. This simplifies the function and the way we test it.

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -493,61 +493,99 @@ def scancel(self, job_id: int) -> None:
...
-            self.logger.error(f"Error querying node user information: {squeue_stderr}")
-            return
-
-        for line in squeue_stdout.split("\n"):
-            if line.strip():
-                node_list, user = line.split("|")
-                for node in self.parse_node_list([node_list]):
-                    node_user_map[node] = user
+        if squeue_output is None or sinfo_output is None:
+            squeue_output, _ = self.fetch_command_output("squeue -o '%N|%u' --noheader")
```
amaslenn, line 508: Do you think it make sense to introduce specific function like:
```py
self.get_squeue_output()
self.get_snifo_output()
```

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -493,61 +493,99 @@ def scancel(self, job_id: int) -> None:
...
+                state_enum = self.convert_state_to_enum(state)
+
+                for node_name in node_names:
+                    for part_name, nodes in self.partitions.items():
+                        if part_name != partition:
+                            continue
+                        for node in nodes:
+                            if node.name == node_name:
+                                node.state = state_enum
+                                node.user = node_user_map.get(node_name, "N/A")
```
amaslenn, line 588: Is it safe to have such value here?

TaekyungHeo: Which value do you mean?

src/cloudai/schema/system/slurm/slurm_system.py
TaekyungHeo: I introduced these primarily for testing the update_node_states function. My intention is to keep the interface simple. When you examine the caller sites of this function, you'll see it is used to ensure all node states are up-to-date. Mandating users to pass squeue and sinfo would complicate the calling process.

src/cloudai/schema/system/slurm/slurm_system.py
TaekyungHeo: Yes, it makes sense. I will do.

TaekyungHeo: Updated.

src/cloudai/schema/system/slurm/slurm_system.py
TaekyungHeo: Do you mean "N/A"?

amaslenn: Yes, the `"N/A"`. Should we raise an issue right here or maybe on a level higher if that is non-recoverable issue? We probably want to raise issues as early as possible, otherwise it will be harder to debug. And harder for our users to understand the issue itself.

src/cloudai/schema/system/slurm/slurm_system.py
amaslenn: Now you can mock `get_squeue()` and `get_sinfo()` outputs and remove this args to keep this interface simple.

TaekyungHeo: I see, I will remove the args then :)

TaekyungHeo: Updated.

src/cloudai/schema/system/slurm/slurm_system.py
TaekyungHeo: I understand this can be an issue and it can be difficult to debug, but I wanted to delay the decision because interfacing with Slurm is tricky. The node_user_map is responsible for maintaining a map between node names and user names. The idea is to use the squeue result to identify the user of a node. A corner case occurs when a node is allocated in the middle of execution and assigned to another user, making the node_user_map outdated. Do you think it's better to raise an exception right now rather than delaying the decision?

amaslenn: I think we need to have a discussion about error handling approach that we'd like to take. For now, IMO it is OK to have it like this.

amaslenn (approved):

PR #24, opened 2024-05-20
Add comparison to ground truth in acceptance test

x41lakazam:
Execute a scenario in dry-run mode and compare the results folder with an expected output folder, if any difference is foundin one of the files, the test fails. Note that the empty lines and commented lines (the ones that start with #) are not taken into account when computing the difference.


tests/test_acceptance.py
```diff
@@ -1,30 +1,104 @@
...
         test_template_path="conf/v0.6/general/test_template",
     )
     handle_dry_run_and_run(args)
 
     test_dir = list(tmp_path.glob("*"))[0]
-    for td in test_dir.iterdir():
-        assert td.is_dir(), "Invalid test directory"
-        assert "Tests." in td.name, "Invalid test directory name"
+
+    if expected_output is not None:
```
amaslenn, line 99: Let's split this function into two: one would verify output and other would only check that directories are created and names are expected.

For the second one let's also check the number of created `Test.*` folders, you can set the expected number in the test params.

tests/test_acceptance.py
```diff
@@ -1,30 +1,104 @@
...
 
 
-@pytest.mark.parametrize("test_scenario_path", SLURM_TEST_SCENARIOS, ids=lambda x: str(x))
-def test_slurm(tmp_path: Path, test_scenario_path: Path):
+def read_wout_comments(f):
+    with open(f) as fp:
+        return "\n".join(line.strip() for line in fp.readlines() if line.strip() and not line.strip().startswith("#"))
+
+
+def diff_dirs(dir1: Path, dir2: Path) -> bool:
```
amaslenn, line 46: Let's update this functions for the readability of those who'll run it and see failures:
1. See my comment below on splitting functions.
2. Let's check only if dirs exist right in the test function and say explicitly which one doesn't exit: test or reference.
3. For the file content, let's rely on `pytest` diff and cmp functionality: do `assert test_content == ref_content, f"file={file_name} has unexpected content"` and it will automatically highlight the diff. Let's not create own diff mechanisms. Again, let's keep it right in the test function as it is just a loop and an assert.

conf/v0.6/general/system/ci.toml
amaslenn, line 1: Can we please use an existing config instead of create a new one? Ideally, we should test all configs we have in all possible valid configuration. So to make it simple for us, we would likely want to keep the amount of configs to minimum yet providing the best possible coverage.

amaslenn (changes requested):

tests/test_acceptance.py
x41lakazam: Isn't it redundant ? As expected files numbers and names are included into ground truth ?

amaslenn: Right now, if there are no directories in the output, test won't fail, because of the empty loop. Considering this scenario, I think it is a useful check.

tests/test_acceptance.py
x41lakazam: Done in https://github.com/NVIDIA/cloudai/pull/24/commits/393e3581c8656269e84600ca4a51005440bd2e70
Can you check ?

tests/test_acceptance.py
amaslenn: Nope, it is under the different code path, we don't run diff for cases where there is no output.

conf/v0.6/general/system/ci.toml
amaslenn: @TaekyungHeo WDYT?

tests/test_acceptance.py
amaslenn: IMO it is OK. This should force users to run it locally and there one can fix it file by file.

tests/test_acceptance.py
```diff
@@ -1,30 +1,91 @@
...
+    """Recursively compare the two directories and check if the files are all the same
+    This function ignore empty lines and commented lines
+    """
+    ok = True
+    assert dir1.is_dir() and dir2.is_dir(), f"One of the directory is invalid ({dir1=}, {dir2=}"
```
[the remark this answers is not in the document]
x41lakazam: What do you mean ?

tests/test_acceptance.py
x41lakazam: I didn't get it, where is this condition ?

tests/test_acceptance.py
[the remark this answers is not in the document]
amaslenn:
```py
assert dir1.is_dir(), f"Reference directory doesn't exist ({dir1})"
assert dir2.is_dir(), f"Test directory doesn't exist ({dir2})"
```

`dir1` and `dir2` could be better names too.

tests/test_acceptance.py
amaslenn:
```py
if expected_output is not None:
    assert diff_dirs(test_dir, expected_output), "Output is not as expected"
else:
    for td in test_dir.iterdir():  # <- HERE if test_dir is empty, test will pass
        assert td.is_dir(), "Invalid test directory"
        assert "Tests." in td.name, "Invalid test directory name"
```

tests/test_acceptance.py
```diff
@@ -1,30 +1,113 @@
...
+"""
+
 import argparse
 from pathlib import Path
+from typing import Optional
 
 import pytest
 from cloudai.__main__ import handle_dry_run_and_run
 
 SLURM_TEST_SCENARIOS = [
```
amaslenn, line 25: Let's also split this into two to simplify parametrisation and highlight the difference between the tests.

conf/v0.6/general/system/ci.toml
TaekyungHeo: @srinivas212 and I agree that we should use the existing configurations. Please confirm, @srinivas212.

x41lakazam: Done in https://github.com/NVIDIA/cloudai/pull/24/commits/bfa2c3e77ab1d93983fa558e0b8fff88e471a1c5.

x41lakazam:
Moved to #60

PR #26, opened 2024-05-21
Enable pydocstyle

amaslenn:
## Summary
Enable `pydocstyle` to ensure single style of documentation. Missing pieces to be added later and checks to be enabled.

## Test Plan
CI

## Additional Notes
—


amaslenn:
> How did you update the docstring style of many functions at once?

No secret here: 
1. Use `ruff'`s `--fix --unsafe-fix` whenever it is possible.
2. Regex or simple find-and-replace for cases where it is possible.
3. Manual. Run -> get issues -> fix. Repeat till there are no issues :)

PR #29, opened 2024-05-22
Register strategies via Registry

amaslenn:
## Summary
1. Register strategies via Registry.
2. Move `TestTemplate` configuration to a higher level to avoid cycle dependencies.
3. Refactor imports, mostly to avoid cycle dependencies.

## Test Plan
CI

## Additional Notes
—


TaekyungHeo:
Let's discuss with @srinivas212. LGTM.

src/cloudai/schema/core/test.py
```diff
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 import sys
-from typing import Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
-from .test_template import TestTemplate
+if TYPE_CHECKING:
+    from .test_template import TestTemplate
```
TaekyungHeo, line 19: Can you help me understand this? What is the main purpose of using TYPE_CHECKING when we can just import the module directly? I did not know there is such a constant.

amaslenn: This is indeed not needed here. Left from my attempts to fix circular dependencies. Thanks for the catch!

TaekyungHeo (approved):

PR #31, opened 2024-05-22
fix updating nodes states

jeffnvidia:
## Summary
Based on main
This PR fixes a bug : cloudAI wasnt able to correctly assess the state of the nodes (IDLE, ALLOCATED ...)
fixed the parse_sinfo_output function

## Test Plan
1. Test by @amaslenn
CI
2. Test by @jeffnvidia
2.1 Slurm command generation for a given URL

$ cloudaix --mode run --system_config_path conf/v0.6/general/system/israel_1.toml --test_scenario_path conf/v0.6/general/test_scenario/llama/llama.toml

### DIFF
--- a/src/cloudai/schema/system/slurm/slurm_system.py
+++ b/src/cloudai/schema/system/slurm/slurm_system.py
@@ -589,20 +589,18 @@ class SlurmSystem(System):
             parts = line.split()
             partition, _, _, _, state, nodelist = parts[:6]
             partition = partition.rstrip("*")
+            node_names = self.parse_node_list([nodelist])
 
-            node_groups = nodelist.split(",")
-            for node_group in node_groups:
-                node_names = self.parse_node_list([node_group.strip()])
-                state_enum = self.convert_state_to_enum(state)
-
-                for node_name in node_names:
-                    for part_name, nodes in self.partitions.items():
-                        if part_name != partition:
-                            continue
-                        for node in nodes:
-                            if node.name == node_name:
-                                node.state = state_enum
-                                node.user = node_user_map.get(node_name, "N/A")
+            # Convert state to enum, handling states with suffixes
+            state_enum = self.convert_state_to_enum(state)
+            for node_name in node_names:
+                for part_name, nodes in self.partitions.items():
+                    if part_name != partition:
+                        continue
+                    for node in nodes:
+                        if node.name == node_name:
+                            node.state = state_enum
+                            node.user = node_user_map.get(node_name, "N/A")
 
     def convert_state_to_enum(self, state_str: str) -> SlurmNodeState:
         """


TaekyungHeo (changes requested):
Could you please update
https://github.com/NVIDIA/cloudai/blob/main/tests/test_slurm_system.py
as well, so that we can understand which cases were not covered and are now covered?

tests/test_slurm_system.py
```diff
@@ -48,6 +54,20 @@ def test_parse_sinfo_output(slurm_system):
...
+def test_parse_sinfo_output2(slurm_system):
+    sinfo_output = """
+    PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
+    main    up    3:00:00      1  inval node-081
+    main    up    3:00:00      5  drain node-[065-066,114,124-125]
+    main    up    3:00:00      2   resv node-[034-035]
+    main    up    3:00:00     88  alloc node-[033,036-064,067-080,082-113,115-123,126-128]
+    backup    up   12:00:00     16  alloc node-[01-16]
+    """
+    node_user_map = {'': 'user1', 'node-033': 'user2', 'node-[036-064': 'user3', '067-080': 'user3', '082-113': 'user3', '115-118]': 'user3', 'node-[119-123': 'user4', '126-128]': 'user4', 'node-01': 'user5', 'node-02': 'user5', 'node-03': 'user5', 'node-04': 'user5', 'node-05': 'user5', 'node-06': 'user5', 'node-07': 'user5', 'node-08': 'user5', 'node-09': 'user5', 'node-10': 'user5', 'node-11': 'user5', 'node-12': 'user5', 'node-13': 'user5', 'node-14': 'user5', 'node-15': 'user5', 'node-16': 'user5'}
```
TaekyungHeo, line 67: I introduced `node_user_map` to create a mapping between nodes and users. A node has an assigned username (UNIX name) if the node is currently being used by a user. My concern about this test case is that we have a malformed node name, such as `node-[036-064`, without a closing bracket. Is this the case during execution? If so, we need to fix it as well.

tests/test_slurm_system.py
```diff
@@ -8,8 +8,14 @@
 @pytest.fixture
 def slurm_system():
     nodes = [
-        SlurmNode(name="nodeA001", partition="main", state=SlurmNodeState.UNKNOWN_STATE),
-        SlurmNode(name="nodeB001", partition="main", state=SlurmNodeState.UNKNOWN_STATE),
+        SlurmNode(name="node-115", partition="main", state=SlurmNodeState.UNKNOWN_STATE),
```
TaekyungHeo, line 11: I see you made an assumption about node names. You assume that node names follow the format node-[num]. However, this is not always guaranteed. We need to cover cases with node[num] as well. I am curious if the updated code covers this case too.

tests/test_slurm_system.py
```diff
@@ -48,6 +54,20 @@ def test_parse_sinfo_output(slurm_system):
     slurm_system.parse_sinfo_output(sinfo_output, node_user_map)
     assert slurm_system.partitions["main"][0].state == SlurmNodeState.IDLE
     assert slurm_system.partitions["main"][1].state == SlurmNodeState.IDLE
+    
+def test_parse_sinfo_output2(slurm_system):
```
TaekyungHeo, line 58: Let's use more descriptive function names for test cases as well. Names like test1, test2, and test3 are not good choices.

TaekyungHeo (changes requested):

TaekyungHeo:
Please fix failing CI pipelines as well.

tests/test_slurm_system.py
```diff
@@ -37,26 +43,51 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
-    node_user_map = {"nodeA001": "root", "nodeB001": "user"}
+    sinfo_output = """
+    PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
+    main    up    3:00:00      1  inval node-081
+    main    up    3:00:00      5  drain node-[065-066,114,124-125]
+    main    up    3:00:00      2   resv node-[034-035]
+    main    up    3:00:00     88  alloc node-[033,036-064,067-080,082-113,115-123,126-128]
+    backup    up   12:00:00     16  alloc node-[01-16]
+    """
+    node_user_map = {'': 'user1',
```
amaslenn, line 55: Please run `ruff format .` locally or set up your editor to do this on save.

tests/test_slurm_system.py
```diff
@@ -37,26 +43,51 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
+    assert slurm_system.partitions["main"][1].state == SlurmNodeState.ALLOCATED
 
 
 @patch("cloudai.schema.system.SlurmSystem.get_squeue")
 @patch("cloudai.schema.system.SlurmSystem.get_sinfo")
 def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue, slurm_system):
-    mock_get_squeue.return_value = "nodeA001|root"
-    mock_get_sinfo.return_value = "PARTITION AVAIL TIMELIMIT NODES STATE NODELIST\n" "main up infinite 1 idle nodeA001"
+    mock_get_squeue.return_value = "node-115|user1"
+    mock_get_sinfo.return_value = "PARTITION AVAIL TIMELIMIT NODES STATE NODELIST\n" "main up infinite 1 idle node-115"
```
amaslenn, line 88: IMO we should keep both.

```py

@pytest.mark.parametrize("sinfo,squeue,expecter_user", [
    ("PARTITION AVAIL TIMELIMIT NODES STATE NODELIST\n" "main up infinite 1 idle nodeA001", "nodeA001|root", "root"),
    ("PARTITION AVAIL TIMELIMIT NODES STATE NODELIST\n" "main up infinite 1 idle node-115", "node-115|user1", "user1"),
])
def test_funtion(sinfo: str, squeue: str, expected_user: str) -> None:
    ...
```

amaslenn (changes requested):

tests/test_slurm_system.py
jeffnvidia: yes this is the output we get from IL1.
If you zant to change the tests you can do so but please approve my MR first

tests/test_slurm_system.py
jeffnvidia: the code I put is just the older version of the function. What's important is that it works for IL1 (while your code doesnt)
Once again, you can approve the MR and then change the function as you please but check that it still works for IL1

lappazos: @jeffnvidia can't you just add this other case here as well? does it add any more complexity?

tests/test_slurm_system.py
jeffnvidia: the old one was deleted now

tests/test_slurm_system.py
jeffnvidia: done

tests/test_slurm_system.py
jeffnvidia: I covered both cases now

tests/test_slurm_system.py
jeffnvidia: both are covered now

tests/test_slurm_system.py
```diff
@@ -37,26 +43,51 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
+                    'node-11': 'user5', 
+                    'node-12': 'user5', 
+                    'node-13': 'user5', 
+                    'node-14': 'user5', 
+                    'node-15': 'user5', 
+                    'node-16': 'user5'}
     slurm_system.parse_sinfo_output(sinfo_output, node_user_map)
-    assert slurm_system.partitions["main"][0].state == SlurmNodeState.IDLE
-    assert slurm_system.partitions["main"][1].state == SlurmNodeState.IDLE
+    assert slurm_system.partitions["main"][0].state == SlurmNodeState.ALLOCATED
```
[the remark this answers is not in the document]
jeffnvidia: done

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -49,7 +49,7 @@ class SlurmSystem(System):
     """
 
     @classmethod
-    def parse_node_list(cls, node_list: List[str]) -> List[str]:
+    def parse_node_list(cls, node_list: str) -> List[str]:
```
amaslenn, line 52: Please add unit tests just for this function:
```py
@pytest.mark.parametrize(...)
def test_parse_node_list(input: str, expected: List[str]):
    ...
```

So far I find its code very confusing and not following the documentation of the function.

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -589,20 +580,22 @@ def parse_sinfo_output(self, sinfo_output: str, node_user_map: Dict[str, str]) -
...
-                        for node in nodes:
-                            if node.name == node_name:
-                                node.state = state_enum
-                                node.user = node_user_map.get(node_name, "N/A")
+            node_names = self.parse_node_list(nodelist)
+
+            # Convert state to enum, handling states with suffixes
+            state_enum = self.convert_state_to_enum(state)
+
+            nodelist.split(",")
```
amaslenn, line 588: What is this line for?

src/cloudai/parser/system_parser/slurm_system_parser.py
```diff
@@ -89,7 +89,9 @@ def safe_int(value):
                 raise ValueError("Partition data does not include a 'name' field.")
 
             raw_nodes = partition_data.get("nodes", [])
-            node_names = set(SlurmSystem.parse_node_list(raw_nodes))
```
amaslenn, line 92: Why not to parse all groups inside `parse_node_list`? It would be easier to use this function. Internally, you can have `parse_one_group` if that is useful.

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -567,10 +560,8 @@ def parse_squeue_output(self, squeue_output: str) -> Dict[str, str]:
 
                 node_list_part, user = parts[0], "|".join(parts[1:])
                 # Handle cases where multiple node groups or ranges are specified
-                node_groups = node_list_part.split(",")
-                for node_group in node_groups:
-                    # Process each node or range using parse_node_list
-                    for node in self.parse_node_list([node_group.strip()]):
+                if node_list_part:
+                    for node in self.parse_node_list(node_list_part.strip()):
```
amaslenn, line 564: IMO `parse_node_list` should correctly handle empty output and return an empty list.

amaslenn (changes requested):

src/cloudai/schema/system/slurm/slurm_system.py
jeffnvidia: how is it confusing ? It's just parsing the outputs of sinfo.
Anyways, I just fixed the function, the description and its use are the same as before.

I'll add a unittest

amaslenn: Function arg is defined as a string: `node_list: str`.
But documentation says it should be a list of strings. So I wonder how is that even possible that input is different, but usage wasn't updated. I hope unit-tests will clarify it.

jeffnvidia: ah yes this part is my mistake it used to be list[str] now its str, I changed the arg but forgot to change the description
I'm updating now

src/cloudai/schema/system/slurm/slurm_system.py
jeffnvidia: right this was in the old version, forgot to delete it

src/cloudai/parser/system_parser/slurm_system_parser.py
jeffnvidia: The function parse_node_list is used in a lot of places and a lot of time, it's not used with a loop.

It is used with a loop at just 2 places so I think it's fine to keep this loop outside the function.

@TaekyungHeo what do you think ?

tests/test_slurm_system.py
```diff
@@ -39,24 +55,73 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
+    )
 
     slurm_system.update_node_states()
+    for node in slurm_system.partitions["backup"]:
+        if node.name == "node01":
+            assert node.state == SlurmNodeState.ALLOCATED
+            assert node.user == "root"
+
+
+def test_parse_node_list(slurm_system):
```
amaslenn, line 114: Please use `@pytest.mark.parametrize`:
```py
@pytest.mark.parametrize("input,expected", [
    ("node-[048-051]", ["node-048", "node-049", "node-050", "node-051"]),
    ...
])
def test_parse_node_list(input: str, expected: List[str]):
    parsed_nodes = slurm_system.parse_node_list(input)
    assert parsed_nodes == expected
```

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -49,41 +49,35 @@ class SlurmSystem(System):
...
-                    if "-" in r:
-                        expanded_nodes.extend([f"{prefix}{str(i).zfill(max_length)}" for i in range(start, end + 1)])
-                    else:
-                        # For single nodes, append directly with appropriate padding.
-                        expanded_nodes.append(f"{prefix}{start_str.zfill(max_length)}")
-            else:
-                expanded_nodes.append(node)
-        return expanded_nodes
+        nodes = []
+        if node_list:
```
amaslenn, line 64: To reduce nesting, you can handle this corner case explicitly:
```py
if not node_list:
    return []

...
```

tests/test_slurm_system.py
```diff
@@ -39,24 +55,73 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
+    node_lists = ["node-[048-051]", "node-[055,114]", "", "node-001", "node[1-4]"]
+
+    expected_outputs = [
```
amaslenn, line 117: Worth adding: `node-name[01-03,05-08,10]`.

tests/test_slurm_system.py
```diff
@@ -39,24 +55,73 @@ def test_parse_squeue_output_with_node_ranges_and_root_user(slurm_system):
...
+        "node05": "user5",
+        "node06": "user5",
+        "node07": "user5",
+        "node08": "user5",
+    }
     slurm_system.parse_sinfo_output(sinfo_output, node_user_map)
-    assert slurm_system.partitions["main"][0].state == SlurmNodeState.IDLE
-    assert slurm_system.partitions["main"][1].state == SlurmNodeState.IDLE
+    for i in range(len(slurm_system.partitions["main"])):
+        assert slurm_system.partitions["main"][i].state == SlurmNodeState.ALLOCATED
```
amaslenn, line 85: According to your `sinfo_output`, nodes have different states, but after parsing all of them are allocated. Is that expected?

src/cloudai/schema/system/slurm/slurm_system.py
```diff
@@ -567,11 +561,8 @@ def parse_squeue_output(self, squeue_output: str) -> Dict[str, str]:
 
                 node_list_part, user = parts[0], "|".join(parts[1:])
                 # Handle cases where multiple node groups or ranges are specified
-                node_groups = node_list_part.split(",")
-                for node_group in node_groups:
-                    # Process each node or range using parse_node_list
-                    for node in self.parse_node_list([node_group.strip()]):
-                        node_user_map[node] = user.strip()
+                for node in self.parse_node_list(node_list_part.strip()):
```
amaslenn, line 564: You can call `.strip()` during initialization to avoid repeating it in the loop.

amaslenn (changes requested):

src/cloudai/schema/system/slurm/slurm_system.py
jeffnvidia: sure I updated

tests/test_slurm_system.py
jeffnvidia: according to sinfo_output, all nodes in main are allocated and all nodes in backup are idle so the behaviour seems good for me

src/cloudai/schema/system/slurm/slurm_system.py
jeffnvidia: updated

tests/test_slurm_system.py
jeffnvidia: added

tests/test_slurm_system.py
jeffnvidia: I updated but then should all the tests look like this ?
Because the rest was done in another way

tests/test_slurm_system.py
amaslenn: Are we talking about the same code? :)

```py
    sinfo_output = """PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
    main    up    3:00:00      1  inval node-081                      # <- NOT ALLOCATED
    main    up    3:00:00      5  drain node-[065-066,114,124-125]    # <- NOT ALLOCATED
    main    up    3:00:00      2   resv node-[034-035]                # <- NOT ALLOCATED
    main    up    3:00:00     88  alloc node-[033,036-064,067-080,082-113,115-123,126-128]
    backup    up   12:00:00     16  idle node[01-16]
    """
```

jeffnvidia: ok I see the misunderstanding
Indeed not all nodes are allocated in the example but if you look at the slurm system, we're just checking nodes 115 to 122 (in main) which are indeed allocated

src/cloudai/parser/system_parser/slurm_system_parser.py
TaekyungHeo: @amaslenn , could you please review our comments and reply?

amaslenn: Current proposal is:
```py
node_names = set()
for group in raw_nodes:
    node_names.update(set(SlurmSystem.parse_node_list(group)))
```

I suggest changing it to:
```py
node_names = set(SlurmSystem.parse_node_list(raw_nodes))
```
And for this we need to move this loop into the `parse_node_list`.

I just don't understand why it is not suitable here. @jeffnvidia could you please explain?

tests/test_slurm_system.py
amaslenn: In other tests it is different, there is no loop by input-expected pairs.

src/cloudai/parser/system_parser/slurm_system_parser.py
TaekyungHeo: cc: @lappazos

jeffnvidia: > Current proposal is:
> 
> ```python
> node_names = set()
> for group in raw_nodes:
>     node_names.update(set(SlurmSystem.parse_node_list(group)))
> ```
> 
> I suggest changing it to:
> 
> ```python
> node_names = set(SlurmSystem.parse_node_list(raw_nodes))
> ```
> 
> And for this we need to move this loop into the `parse_node_list`.
> 
> I just don't understand why it is not suitable here. @jeffnvidia could you please explain?

Yes it is possible of courser.
Once again, I'm just saying that in 3 other places we use this function without a list so I would need to create an artificial list with one element, I think it wouldnt look good.
@amaslenn

amaslenn: Do you mean this:
```py
def parse_node_list(nodes: Union[List[str], str], ...):
    all_nodes = nodes
    if isinstance(nodes, str):
        all_nodes = [nodes]
    ...
```
?

Here is a more testable solution:
```py
def parse_one_node_list(nodes: str, ...) -> ...:
    ...

def parse_node_list(nodes: Union[List[str], str], ...) -> ...:
    all_nodes: List[str] = []
    if isinstance(nodes, list):
        all_nodes.extend(self.parse_one_node_list(...))
    else:
        all_nodes = self.parse_one_node_list(...)
    return all_nodes
```

jeffnvidia: are you sure this is really needed for just 2 loops ?
It looks overly complicated in my opinion

tests/test_slurm_system.py
jeffnvidia: updated

src/cloudai/parser/system_parser/slurm_system_parser.py
amaslenn: OK, let's keep external loops.

amaslenn (approved):

TaekyungHeo (approved):

PR #50, opened 2024-05-30
Lint imports and high level design doc

amaslenn:
## Summary
1. Add import linting to make sure `_core` doesn't depend on any higher-level modules.
2. Rework `TestTemplateStrategy` to avoid dependency on `SlurmSystem`.
3. Add initial design description.

## Test Plan
CI

## Additional Notes
—


TaekyungHeo:
> Add import linting to make sure _core doesn't depend on any higher-level modules.
I have not used lint-imports before, but it looks like a good idea.

doc/DEV.md
```diff
@@ -0,0 +1,30 @@
+# Development
```
TaekyungHeo, line 1: Let's make sure @srinivas212 is on the same page with this DEV.md.

src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
```diff
@@ -14,21 +14,12 @@
...
 from typing import Any, Dict, List
 
-from cloudai._core.system import System
 from cloudai.schema.system.slurm.strategy import SlurmCommandGenStrategy
 
 
 class ChakraReplaySlurmCommandGenStrategy(SlurmCommandGenStrategy):
     """Command generation strategy for ChakraReplay on Slurm systems."""
 
-    def __init__(
```
TaekyungHeo, line 24: Thanks for cleaning up. Just a quick question: Is it part of "Check that _core doesn't depend on other parts"? Did you add these changes to this commit for convenience?

amaslenn: Removal of the constructors wasn't necessary, I was checking the dependencies to ensure my change is safe. At some point I realised, that these constructors could be removed and decided to include it into the same PR. If you think it will be better to split, I can do that too.

TaekyungHeo: Let's have it in the same PR. Thanks.

doc/DEV.md
TaekyungHeo: @srinivas212  reviewed and approved.

TaekyungHeo (approved):

artemry-nv (approved):

PR #53, opened 2024-05-30
Implement NcclTestJobStatusRetrievalStrategy and add corresponding tests

TaekyungHeo:
## Summary
Implement NcclTestJobStatusRetrievalStrategy and add corresponding tests

## Test Plan
1. Added the following tests.
* TestNcclTestJobStatusRetrievalStrategy
  * test_no_stdout_file
  * test_successful_job
  * test_failed_job

2. Tested on a real system manually.
```
Test Scenario: nccl-test

Section Name: Tests.1
  Test Name: nccl_test_all_gather
  Description: all_gather
  No dependencies

Section Name: Tests.2
  Test Name: nccl_test_all_reduce
  Description: all_reduce
  Start Post Comp: Tests.1, Time: 0 seconds

Section Name: Tests.3
  Test Name: nccl_test_reduce_scatter
  Description: reduce_scatter
  Start Post Comp: Tests.2, Time: 0 seconds

Section Name: Tests.4
  Test Name: nccl_test_alltoall
  Description: alltoall
  Start Post Comp: Tests.3, Time: 0 seconds
[Kill Tests.1]
Job 260660 for test Tests.1 failed: Missing success indicators in stdout.txt: '# Out of bounds values', '# Avg bus bandwidth'. These keywords are expected to be present in stdout.txt, usually towards the end of the file. Please ensure the NCCL test ran to completion. You can run the generated sbatch script manually and check if stdout.txt is created and contains the expected keywords.
```

### DIFF
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -24,7 +24,8 @@ from types import FrameType
 from typing import Dict, List, Optional
 
 from .base_job import BaseJob
-from .exceptions import JobSubmissionError
+from .exceptions import JobFailureError, JobSubmissionError
+from .job_status_result import JobStatusResult
 from .system import System
 from .test import Test
 from .test_scenario import TestScenario
@@ -287,10 +288,35 @@ class BaseRunner(ABC):
         self.logger.debug("Monitoring jobs.")
         for job in list(self.jobs):
             if self.is_job_completed(job):
-                completed_jobs_count += 1
-                await self.handle_job_completion(job)
+                if self.mode == "dry-run":
+                    completed_jobs_count += 1
+                else:
+                    job_status_result = self.get_job_status(job)
+                    if job_status_result.is_successful:
+                        completed_jobs_count += 1
+                        await self.handle_job_completion(job)
+                    else:
+                        error_message = (
+                            f"Job {job.id} for test {job.test.section_name} failed: {job_status_result.error_message}"
+                        )
+                        self.logger.error(error_message)
+                        print(error_message, file=sys.stdout)
+                        await self.shutdown()
+                        raise JobFailureError(job.test.section_name, error_message, job_status_result.error_message)
 
         return completed_jobs_count
+
+    def get_job_status(self, job: BaseJob) -> JobStatusResult:
+        """
+        Retrieve the job status from a specified output directory.
+
+        Args:
+            job (BaseJob): The job to be checked.
+
+        Returns:
+            JobStatusResult: The result containing the job status and an optional error message.
+        """
+        return job.test.get_job_status(job.output_path)
 
     async def handle_job_completion(self, completed_job: BaseJob):
         """
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -283,17 +283,17 @@ class BaseRunner(ABC):
         Returns
             int: The number of completed jobs.
         """
-        completed_jobs_count = 0
+        successful_jobs_count = 0
 
         self.logger.debug("Monitoring jobs.")
         for job in list(self.jobs):
             if self.is_job_completed(job):
                 if self.mode == "dry-run":
-                    completed_jobs_count += 1
+                    successful_jobs_count += 1
                 else:
                     job_status_result = self.get_job_status(job)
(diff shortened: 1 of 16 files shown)


src/cloudai/_core/base_runner.py
```diff
@@ -287,11 +288,36 @@ async def monitor_jobs(self) -> int:
         self.logger.debug("Monitoring jobs.")
         for job in list(self.jobs):
             if self.is_job_completed(job):
-                completed_jobs_count += 1
-                await self.handle_job_completion(job)
+                if self.mode == "dry-run":
```
amaslenn, line 291: Would be nice to somehow use the default no-op behaviour here to avoid this conditions.

amaslenn (commented):
In general, everything is OK and compliant with our design.

But IMO we overuse Strategy pattern. For example, in this case we can put `get_job_status(path)` into a particular test implementation itself (`src/cloudai/schema/test_template/nccl_test/template.py`):
1. By default, `TestTemplate` has no-op returning OK.
2. Each test overrides it if needed. Because only test knows how to define it.
3. We won't need registrations and assertion if such implementation exists. Second part might be especially difficult because one might notice it only while running tests. So we might even consider making it an abstract method to highlight its importance.

I'm not saying we change it right now, but I want to discuss it and better understand why you made it like this.

src/cloudai/schema/test_template/nccl_test/job_status_retrieval_strategy.py
```diff
@@ -0,0 +1,60 @@
...
+                    f"Missing success indicators in stdout.txt: {', '.join(missing_indicators)}. "
+                    "These keywords are expected to be present in stdout.txt, usually towards the end of the file. "
+                    "Please ensure the NCCL test ran to completion. You can run the generated sbatch script manually "
+                    "and check if stdout.txt is created and contains the expected keywords."
+                )
+                return JobStatusResult(is_successful=False, error_message=error_message)
+        return JobStatusResult(
+            is_successful=False,
+            error_message=(
+                "stdout.txt file not found in the specified output directory. "
```
amaslenn, line 55: IMO worth adding the directory in which we are looking this file.

src/cloudai/_core/base_runner.py
```diff
@@ -287,11 +288,36 @@ async def monitor_jobs(self) -> int:
...
+                    completed_jobs_count += 1
+                else:
+                    job_status_result = self.get_job_status(job)
+                    if job_status_result.is_successful:
+                        completed_jobs_count += 1
```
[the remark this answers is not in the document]
TaekyungHeo: Updated.

src/cloudai/_core/base_runner.py
TaekyungHeo: I didn't quite understand. Could you clarify your suggestion?

src/cloudai/schema/test_template/nccl_test/job_status_retrieval_strategy.py
TaekyungHeo: It makes perfect sense.

TaekyungHeo:
Thanks, @amaslenn. This PR actually depends on https://github.com/NVIDIA/cloudai/pull/46. Therefore, I assume these comments are for that PR.

I agree that we may overuse the strategy pattern. However, I would like to keep the current design for future use cases. Currently, it is used in a single instance where the path is given directly. However, this may change, and we may need different implementations for each scheduler. For example, in Meta, we cannot access the file system directly because it has an internal remote file system called buckets. Therefore, simply providing an output path may not work. Moreover, the job status may need to be retrieved with a special library or interface as they have an internal scheduler and interfaces. For these reasons, I am still actively using the strategy pattern.

src/cloudai/schema/test_template/nccl_test/job_status_retrieval_strategy.py
TaekyungHeo: Updated.

src/cloudai/_core/base_runner.py
srinivas212: @amaslenn  No-op = run = else part, no?

amaslenn (approved):

PR #56, opened 2024-05-31
Improve installation

TaekyungHeo:
## Summary
Improve installation

## Test Plan
```

```


install_config.toml
srinivas212, line 1: We have some of this info like info path in the system schema file. We need to remove it. 

This file should be in conf/... and every user should be able to create their own for different systems. 

Separate to this PR, we should document all the different knobs we have in different schemas and add to user guide.

PR #21, opened 2024-05-19
use official online container

jeffnvidia:
## Summary
Based on PR 31
This MR is for using the official latest NeMo version

## Test Plan
Test by @amaslenn
CI
Test by @jeffnvidia
2.1 Slurm command generation
$ cloudaix --mode run --system_config_path conf/v0.6/general/system/israel_1.toml --test_scenario_path conf/v0.6/general/test_scenario/llama/llama.toml

PR #81, opened 2024-06-05
Add credentials for NGC to README

lappazos:
## Summary
Add additional information to user on needed step to add required credentials to NGC containers.

## Additional Notes
Following #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,15 @@
 CloudAI benchmark suite aims to develop an industry standard benchmark focused on grading Data Center (DC) scale AI systems in the Cloud.
 
 ## Quick Start
-First, clone the CloudAI repository to your local machine:
+First, ensure you have access to the docker repository. This is a manual step:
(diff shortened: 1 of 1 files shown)


srinivas212:
Should this go under https://github.com/NVIDIA/cloudai/blob/main/USER_GUIDE.md instead?

lappazos:
> Should this go under https://github.com/NVIDIA/cloudai/blob/main/USER_GUIDE.md instead?

I believe it belongs to README, as it is a required part of the needed step for installation. Also, README is the first document you encounter when opening the repo. Some users asked me how to do that. But also USER_GUIDE can fit, so whatever you decide. which one should it be?

lappazos:
@TaekyungHeo please review as well

TaekyungHeo (approved):
LGTM. Approved. Let's wait for Srinivas to discuss, update, and merge.

I believe having this in the README is better for users.

PR #83, opened 2024-06-06
New CLI args and filter tests templates for installation

amaslenn:
## Summary
1. Update CLI args to highlight where we need a directory or just a file. Replaced underscores with dashes as it seems to be nowadays' standard.
2. Use `TestScenario` for filtering tests and test templates used for installation.

## Test Plan
CI

## Additional Notes
—

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,31 +38,31 @@ Please make sure to use the correct system configuration file that corresponds t
 ```bash
 cloudai\
     --mode install\
-    --system_config_path conf/v0.6/general/system/example_slurm_cluster.toml
+    --system-config conf/v0.6/general/system/example_slurm_cluster.toml
 ```
 
 To simulate running experiments without execution, use the dry-run mode:
 ```bash
 cloudai\
     --mode dry-run\
-    --system_config_path conf/v0.6/general/system/example_slurm_cluster.toml\
-    --test_scenario_path conf/v0.6/general/test_scenario/sleep.toml
+    --system-config conf/v0.6/general/system/example_slurm_cluster.toml\
+    --test-scenario conf/v0.6/general/test_scenario/sleep.toml
 ```
 
 To run experiments, execute CloudAI CLI in run mode:
 ```bash
 cloudai\
     --mode run\
-    --system_config_path conf/v0.6/general/system/example_slurm_cluster.toml\
-    --test_scenario_path conf/v0.6/general/test_scenario/sleep.toml
+    --system-config conf/v0.6/general/system/example_slurm_cluster.toml\
+    --test-scenario conf/v0.6/general/test_scenario/sleep.toml
 ```
 
 To generate reports, execute CloudAI CLI in generate-report mode:
 ```bash
 cloudai\
     --mode generate-report\
-    --system_config_path conf/v0.6/general/system/example_slurm_cluster.toml\
-    --output_path /path/to/output_directory
+    --system-config conf/v0.6/general/system/example_slurm_cluster.toml\
+    --output-dir /path/to/output_directory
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -46,31 +46,31 @@ Please make sure to use the correct system configuration file that corresponds t
 ```bash
 cloudai\
     --mode install\
-    --system_config_path conf/v0.6/general/system/example_slurm_cluster.toml
+    --system-config conf/v0.6/general/system/example_slurm_cluster.toml
```
TaekyungHeo, line 49: - We need to update CloudAIX arguments and the README for consistency.
- We should inform users that the usage will be updated (#cloudai-users).

src/cloudai/__main__.py
```diff
@@ -96,29 +96,29 @@ def parse_arguments() -> argparse.Namespace:
         ),
     )
     parser.add_argument(
-        "--system_config_path",
+        "--system-config",
```
TaekyungHeo, line 99: I like these updates. "_path" is not essential. Having "-dir" at the end allows users to know whether they should provide a file or directory.

cc: @srinivas212

src/cloudai/__main__.py
```diff
@@ -127,92 +127,68 @@ def parse_arguments() -> argparse.Namespace:
...
-        system.output_path = str(output_path.absolute())
-
-    system.update()
-
     logging.info(f"System Name: {system.name}")
     logging.info(f"Scheduler: {system.scheduler}")
     logging.info(f"Test Scenario Name: {test_scenario.name}")
 
-    test_scenario.pretty_print()
+    if mode == "run":
```
TaekyungHeo, line 182: We need to update CloudAIX as well.

tests/test_acceptance.py
```diff
@@ -35,14 +36,13 @@ def test_slurm(tmp_path: Path, scenario: Dict):
...
-        "dry-run",
-        Path("conf/v0.6/general/system/example_slurm_cluster.toml"),
-        Path("conf/v0.6/general/test_template"),
-        Path("conf/v0.6/general/test"),
-        test_scenario_path,
-        tmp_path,
+    parser = Parser(
+        Path("conf/v0.6/general/system/example_slurm_cluster.toml"), Path("conf/v0.6/general/test_template")
     )
+    system, tests, test_scenario = parser.parse(Path("conf/v0.6/general/test"), test_scenario_path)
```
TaekyungHeo, line 42: **Explanation** - At this point, tests will have a list of tests that are used only in the test scenario. Therefore, we can safely check if the test templates were installed or not.

TaekyungHeo (approved):
LGTM. Let's wait for @srinivas212 and @artemry-nv. No need to change this PR, but changes are required in CloudAIX.

README.md
amaslenn: I'm not sure we should mirror this project and extensions. It might be even better to keep them as different as possible, because they are different :)

Anyway, I'm OK to update it there too, no problem. Once merged, we can create a new pre-release and update our users. Plus prepare update for Extensions.

tests/test_acceptance.py
amaslenn: Yes, I wanted to double check it with you, thank you for looking into it!

TaekyungHeo:
+ This requires changes to our internal Confluence page (Daria's test plan).

README.md
srinivas212: The intuition to remove _path makes sense. Given that's the case, we should do the same fix in CloudAIX.

srinivas212 (approved):

PR #88, opened 2024-06-07
Validate TestScenarios against schema

amaslenn:
## Summary
Introduce `jsonschema` to validate test scenarios.

The idea is to make sure our TOML files are compliant with some sort of schema. In future, we can deserialise object into dataclasses without real parsing.

This is likely breaking dependency references and we need to find a way to fix it.

## Test Plan
CI

## Additional Notes
—

### DIFF
--- a/conf/v0.6/general/test_scenario/chakra_replay.toml
+++ b/conf/v0.6/general/test_scenario/chakra_replay.toml
@@ -1,5 +1,6 @@
 name = "chakra_replay"
 
 [[Tests]]
-name = "chakra_replay"
+test = "chakra_replay"
+name = "Test.1"
 num_nodes = 2
--- a/conf/v0.6/general/test_scenario/chakra_replay.toml
+++ b/conf/v0.6/general/test_scenario/chakra_replay.toml
@@ -1,6 +1,5 @@
 name = "chakra_replay"
 
-[Tests]
-  [Tests.1]
-    name = "chakra_replay"
-    num_nodes = "2"
+[[Tests]]
+name = "chakra_replay"
+num_nodes = 2
--- a/conf/v0.6/general/test_scenario/nccl_test.toml
+++ b/conf/v0.6/general/test_scenario/nccl_test.toml
@@ -1,174 +1,174 @@
 name = "nccl-test"
 
-[Tests.1]
-  name = "nccl_test_all_reduce"
-  num_nodes = "2"
-  time_limit = "00:20:00"
+[[Tests]]
+name = "nccl_test_all_reduce"
+num_nodes = 2
+time_limit = "00:20:00"
 
-[Tests.2]
-  name = "nccl_test_all_gather"
-  num_nodes = "2"
-  time_limit = "00:20:00"
-  [Tests.2.dependencies]
-    start_post_comp = { name = "Tests.1", time = 0 }
+[[Tests]]
+name = "nccl_test_all_gather"
+num_nodes = 2
+time_limit = "00:20:00"
+[Tests.dependencies]
+start_post_comp = { name = "Tests.1", time = 0 }
 
-[Tests.3]
-  name = "nccl_test_reduce_scatter"
-  num_nodes = "2"
-  time_limit = "00:20:00"
-  [Tests.3.dependencies]
-    start_post_comp = { name = "Tests.2", time = 0 }
+[[Tests]]
+name = "nccl_test_reduce_scatter"
+num_nodes = 2
+time_limit = "00:20:00"
+[Tests.dependencies]
+start_post_comp = { name = "Tests.2", time = 0 }
 
-[Tests.4]
-  name = "nccl_test_alltoall"
-  num_nodes = "2"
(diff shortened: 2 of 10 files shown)


conf/v0.6/general/test_scenario/nccl_test.toml
```diff
@@ -1,174 +1,174 @@
...
-  [Tests.24.dependencies]
-    start_post_comp = { name = "Tests.23", time = 0 }
-
-[Tests.25]
-  name = "nccl_test_bisection"
-  num_nodes = "2"
-  time_limit = "00:20:00"
-  [Tests.25.dependencies]
-    start_post_comp = { name = "Tests.24", time = 0 }
+[[Tests]]
```
TaekyungHeo, line 3: We should find a way to uniquely identify each test. This change does not allow us to have multiple tests using the same test schema. Currently, section names like "Tests.1" work as unique IDs to identify test instances within a test scenario. There can be multiple `nccl_test_all_reduce` instances with different section names. However, this change does not permit that. We may need to introduce a new field for unique IDs.

tests/test_toml_files.py
```diff
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 from pathlib import Path
 
+import jsonschema
```
TaekyungHeo, line 18: I like the idea of using jsonschema to validate and enforce our schema.

conf/v0.6/general/test_scenario/nccl_test.toml
```diff
@@ -1,174 +1,174 @@
...
+name = "nccl_test_all_reduce"
+num_nodes = 2
+time_limit = "00:20:00"
+
+[[Tests]]
+name = "nccl_test_all_gather"
+num_nodes = 2
+time_limit = "00:20:00"
+[Tests.dependencies]
+start_post_comp = { name = "Tests.1", time = 0 }
```
TaekyungHeo, line 13: There is no name called "Tests.1". I believe this is not a valid dependency.

conf/v0.6/general/test_scenario/nccl_test.toml
amaslenn: I've introduced `test` field to keep Test name and `name` to have this case name and use it in dependencies spec.

conf/v0.6/general/test_scenario/nccl_test.toml
amaslenn: Thanks! Reworked it so it works again :)

conf/v0.6/general/test_scenario/nccl_test.toml
TaekyungHeo: It makes sense to me. Please review, @srinivas212.

conf/v0.6/general/test_scenario/chakra_replay.toml
```diff
@@ -1,6 +1,6 @@
 name = "chakra_replay"
 
-[Tests]
-  [Tests.1]
-    name = "chakra_replay"
-    num_nodes = "2"
+[[Tests]]
+test = "chakra_replay"
+name = "Test.1"
```
TaekyungHeo, line 5: How about switching the two lines so that the name field appears before the test field? Let's see what @srinivas212 thinks before we make changes.

TaekyungHeo, line 5: How about switching the two lines so that the name field appears before the test field? Let's see what @srinivas212 thinks before we make changes.

amaslenn: I don't think we can control ordering. I mean using some specific ordering in our own configs is fine, but we won't be able to automatically check and enforce it even in our repo.

TaekyungHeo: It makes sense.

conf/v0.6/general/test_scenario/nccl_test.toml
srinivas212: We need to align the terminology. 

In the test file, we have a test template name and test name. 

Previously, we maintain this distinction clearly. Test.1 was the unique id and name referred to the test name in test file. There was a 1:1 mapping. 

Now we are making that "name" -> "test" and changing the name to the unique name.

schemas/test_scenario.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
artemry-nv, line 88: Missing EOF - may lead to parsing issues.

conf/v0.6/general/test_scenario/nccl_test.toml
hocinemahni: That's an interesting point about the uniqueness of identifiers.

amaslenn:
To be re-implemented using Pydantic

PR #105, opened 2024-06-18
Refactor handle_install_and_uninstall to identify unique test templates

TaekyungHeo:
## Summary
Refactor installer for TestTemplate iterables in Slurm and base classes. Currently, all tests in a test scenario directory are passed to `is_installed`, `install`, and `uninstall`, which is inefficient. This approach triggers multiple processes for installation and uninstallation. This pull request now uniquely identifies test templates within the directory before passing them to the respective methods.

## Test Plan
**1. CI passes.
2. Ran on a system**
```
$ python ./cloudaix.py --mode install --system-config ...
[INFO] Scheduler: slurm                                                     
[INFO] Checking installation status of components.                          
[INFO] Verifying installation status of test templates.                     
[INFO] Test template NcclTest is not installed.
[INFO] Not all components are ready, preparing
...
```

### DIFF
--- a/src/cloudai/__main__.py
+++ b/src/cloudai/__main__.py
@@ -18,9 +18,9 @@ import logging
 import logging.config
 import sys
 from pathlib import Path
-from typing import List
+from typing import List, Set
 
-from cloudai import Installer, Parser, ReportGenerator, Runner, System, Test, TestScenario
+from cloudai import Installer, Parser, ReportGenerator, Runner, System, Test, TestScenario, TestTemplate
 
 
 def setup_logging(log_file: str, log_level: str) -> None:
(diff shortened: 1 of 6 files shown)


srinivas212 (approved):

amaslenn (commented):
This is a great improvement! I was going to talk about it with you @TaekyungHeo but now it is done :)

I disagree with the overall design though:
1. Uniq identification should be part of `BaseInstaller` class, not `__main__`.
2. We now have duplicating logic in two functions in `__main__` and all the dependencies will have to implement it as part of their CLIs.
3. Having it under `BaseInstaller` would isolate this functionality in a logical place of Cloud AI Core where it would be easier to test, current tests are not covering `identify_unique_test_templates`.

PR #111, opened 2024-06-19
Add pip install requirements.txt step to NeMoLauncherSlurmInstallStrategy

TaekyungHeo:
## Summary
Add pip install requirements.txt step to NeMoLauncherSlurmInstallStrategy
```
ModuleNotFoundError: No module named 'hydra''
ModuleNotFoundError: No module named 'tqdm''
ModuleNotFoundError: No module named 'zstandard''
```

## Test Plan
**1. CI passes**
**2. Ran on a real system**
```
$ pip list | grep hydra
hydra-core              1.2.0

$ pip uninstall hydra-core

$ python ./cloudaix.py --mode install --system-config ...
[INFO] Scheduler: slurm
[INFO] Not all components are ready
[INFO] Installing test templates.
...
[INFO] 5/5 Installation for NeMoLauncher finished with status: OK
[INFO] Installation successful.

$ pip list | grep hydra
hydra-core              1.2.0
```

### DIFF
--- a/src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
+++ b/src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
@@ -157,6 +157,7 @@ class NeMoLauncherSlurmInstallStrategy(SlurmInstallStrategy):
 
         try:
             self._clone_repository(subdir_path)
+            self._install_requirements(subdir_path)
             docker_image_result = self.docker_image_cache_manager.ensure_docker_image(
                 self.docker_image_url, self.SUBDIR_PATH, self.DOCKER_IMAGE_FILENAME
             )
@@ -293,3 +294,22 @@ class NeMoLauncherSlurmInstallStrategy(SlurmInstallStrategy):
         result = subprocess.run(checkout_cmd, cwd=repo_path, capture_output=True, text=True)
         if result.returncode != 0:
             raise RuntimeError(f"Failed to checkout commit: {result.stderr}")
+
+    def _install_requirements(self, subdir_path: str) -> None:
+        """
+        Installs the required Python packages from the requirements.txt file in the cloned repository.
+
+        Args:
+            subdir_path (str): Subdirectory path for installation.
+        """
+        repo_path = os.path.join(subdir_path, self.REPOSITORY_NAME)
+        requirements_file = os.path.join(repo_path, "requirements.txt")
+
+        if os.path.isfile(requirements_file):
+            logging.debug("Installing requirements from %s", requirements_file)
+            install_cmd = ["pip", "install", "-r", requirements_file]
+            result = subprocess.run(install_cmd, capture_output=True, text=True)
+            if result.returncode != 0:
+                raise RuntimeError(f"Failed to install requirements: {result.stderr}")
+        else:
+            logging.warning("requirements.txt not found in %s", repo_path)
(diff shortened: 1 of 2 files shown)


src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
```diff
@@ -293,3 +294,22 @@ def _clone_repository(self, subdir_path: str) -> None:
...
             raise RuntimeError(f"Failed to checkout commit: {result.stderr}")
+
+    def _install_requirements(self, subdir_path: str) -> None:
+        """
+        Installs the required Python packages from the requirements.txt file in the cloned repository.
+
+        Args:
+            subdir_path (str): Subdirectory path for installation.
+        """
+        repo_path = os.path.join(subdir_path, self.REPOSITORY_NAME)
```
amaslenn, line 305: _[minor, feel free to ignore]_ we can use `Path` here as well, not only in tests.

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
```diff
@@ -293,3 +294,22 @@ def _clone_repository(self, subdir_path: str) -> None:
...
+        requirements_file = os.path.join(repo_path, "requirements.txt")
+
+        if os.path.isfile(requirements_file):
+            logging.debug("Installing requirements from %s", requirements_file)
+            install_cmd = ["pip", "install", "-r", requirements_file]
```
amaslenn, line 310: Isn't it risky to call `pip` without knowing if virtual env is created? It might pollute main python env or fail if we don't have permission to change it.

TaekyungHeo: Yes, it makes sense. To update the behavior, we need to update how NeMo is installed. Currently, it is reusing the venv created for CloudAI. About 4-5 months ago, I tried to create a separate venv for NeMo and use it, but this approach did not work well, so I reverted to the current method. Let's stick with this approach for now to fix the bug, and then move on to a better solution.

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
TaekyungHeo: I see. Are you suggesting using Path instead of str for the type annotation of subdir_path?

amaslenn: Not just type annotation, but actually use `Path` objects instead of `str` + `os.path`.

amaslenn (approved):

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
TaekyungHeo: Do you mean something like this?
```
repo_path = Path(subdir_path) / self.REPOSITORY_NAME
```

amaslenn: Yes. Plus, `subdir_path` can also be a `Path` object.

TaekyungHeo: Okay, let's resolve this in a separate PR. This pattern is widely observed in different places, and we need to update it as a whole.

PR #119, opened 2024-07-02
Update BaseRunner to include scenario name in output directory

TaekyungHeo:
## Summary
Update BaseRunner to include scenario name in output directory

## Test Plan
Added unit tests and CI passes.

### DIFF
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -69,7 +69,7 @@ class BaseRunner(ABC):
         self.mode = mode
         self.system = system
         self.test_scenario = test_scenario
-        self.output_path = self.setup_output_directory(test_scenario.name, system.output_path)
+        self.output_path = self.setup_output_directory(system.output_path)
         self.monitor_interval = system.monitor_interval
         self.jobs: List[BaseJob] = []
         self.test_to_job_map: Dict[Test, BaseJob] = {}
@@ -77,13 +77,11 @@ class BaseRunner(ABC):
         self.shutting_down = False
         self.register_signal_handlers()
 
-    @staticmethod
-    def setup_output_directory(scenario_name: str, base_output_path: str) -> str:
+    def setup_output_directory(self, base_output_path: str) -> str:
         """
         Set up and return the output directory path for the runner instance.
 
         Args:
-            scenario_name (str): The name of the test scenario.
             base_output_path (str): The base output directory.
 
         Returns:
@@ -92,7 +90,7 @@ class BaseRunner(ABC):
         if not os.path.exists(base_output_path):
             os.makedirs(base_output_path)
         current_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
-        output_subpath = os.path.join(base_output_path, f"{scenario_name}_{current_time}")
+        output_subpath = os.path.join(base_output_path, f"{self.test_scenario.name}_{current_time}")
         os.makedirs(output_subpath)
         return output_subpath
 
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -69,7 +69,7 @@ class BaseRunner(ABC):
         self.mode = mode
         self.system = system
         self.test_scenario = test_scenario
-        self.output_path = self.setup_output_directory(system.output_path)
+        self.output_path = self.setup_output_directory(test_scenario.name, system.output_path)
         self.monitor_interval = system.monitor_interval
         self.jobs: List[BaseJob] = []
         self.test_to_job_map: Dict[Test, BaseJob] = {}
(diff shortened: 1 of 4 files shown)


src/cloudai/_core/base_runner.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         self.mode = mode
         self.system = system
         self.test_scenario = test_scenario
-        self.output_path = self.setup_output_directory(system.output_path)
+        self.output_path = self.setup_output_directory(test_scenario.name, system.output_path)
```
amaslenn, line 72: We already have `self.test_scenario` wouldn't it be better to access it from `setup_output_directory`? That would keep the API unchanged.

tests/test_base_runner.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pytest.fixture
+def mock_datetime_now():
+    with patch("cloudai._core.base_runner.datetime") as mock_datetime:
+        mock_datetime.now.return_value = datetime(2024, 1, 1, 12, 0, 0)
+        mock_datetime.strftime = datetime.strftime
+        yield mock_datetime
+
+
+def test_setup_output_directory(mock_datetime_now, tmp_path):  # noqa
```
amaslenn, line 32: Can we avoid using `# noqa`? It can fight back at some point but won't letting us know about real issues.

tests/test_base_runner.py
```diff
@@ -0,0 +1,74 @@
...
+    base_output_path.mkdir()
+    output_path = BaseRunner.setup_output_directory(scenario_name, str(base_output_path))
+
+    print(f"Expected path: {expected_path}, Output path: {output_path}")
+
+    assert os.path.exists(expected_path)
+    assert output_path == str(expected_path)
+
+
+def test_setup_output_directory_handles_oserror(mock_datetime_now, tmp_path):  # noqa
```
amaslenn, line 62: In this case we are testing standard lib behaviour, right? Do we need that?

tests/test_base_runner.py
```diff
@@ -0,0 +1,74 @@
...
+    scenario_name = "test_scenario"
+    base_output_path = tmp_path / "base_output_path"
+    expected_time_str = "2024-01-01_12-00-00"
+    expected_path = base_output_path / f"{scenario_name}_{expected_time_str}"
+
+    output_path = BaseRunner.setup_output_directory(scenario_name, str(base_output_path))
+
+    print(f"Expected path: {expected_path}, Output path: {output_path}")
+
+    assert os.path.exists(base_output_path)
```
amaslenn, line 42: [minor] could be `assert base_output_path.exists()`, same for `expected_path`.

tests/test_base_runner.py
TaekyungHeo: Linters are complaining that mock_datetime_now is not being used. How can I fix these errors?

tests/test_base_runner.py
TaekyungHeo: Deleted. Thanks!

tests/test_base_runner.py
TaekyungHeo: Updated. Thanks!

tests/test_base_runner.py
TaekyungHeo:
```
$ vulture . | grep 100       
tests/test_base_runner.py:32: unused variable 'mock_datetime_now' (100% confidence)
tests/test_base_runner.py:47: unused variable 'mock_datetime_now' (100% confidence)
```

src/cloudai/_core/base_runner.py
amaslenn: If we want to give users more control over output directory name, let's add a separate function for that and `setup_output_directory` will use that function (or property).

tests/test_base_runner.py
amaslenn: Ah, I see. Guess it should be a feature request for `vulture`, they don't handle `pytest`'s magic well.

Can we disable only `vulture` only for these lines? For `ruff` we can do: `... # noqa: <rule id>`.

TaekyungHeo: Updated.

src/cloudai/_core/base_runner.py
TaekyungHeo: Updated.

TaekyungHeo:
I appreciate your comments! All resolved. Please review, @amaslenn.

tests/test_base_runner.py
```diff
@@ -0,0 +1,86 @@
...
+    # Mock TestScenario and System
+    mock_test_scenario = MagicMock(spec=TestScenario)
+    mock_test_scenario.name = scenario_name
+    mock_system = MagicMock(spec=System)
+    mock_system.output_path = str(base_output_path)
+    mock_system.monitor_interval = 5
+
+    runner = MockRunner("run", mock_system, mock_test_scenario)
+
+    assert os.path.exists(expected_path)
```
amaslenn, line 85: [minor] for consistency, we can use `expected_path.exists()` here as well. And that should eliminate the need in `import os`.

amaslenn (approved):

PR #132, opened 2024-07-09
Update README: add mandatory args test-templates-dir and tests-dir

TaekyungHeo:
## Summary
Update README: add mandatory args test-templates-dir and tests-dir

## Test Plan
Not needed.


amaslenn (approved):
~~For Extensions we might need to communicate this change in advance.~~
UPD Sorry, it is just Doc update.

PR #141, opened 2024-07-09
Hierarchical Test template for support Grok/GPT via PAXML

srivatsankrishnan:
## Summary
Modification to Jaxtoolbox to support hierarchical test template to support both GPT and Grok models via PAXML. 

## Test Plan
Internal testing and pytests

## Additional Notes
Needs additional testing on same container. XLA Flags seems to have changed. It requires debugging to make Grok working on the same container.


src/cloudai/_core/test_template.py
```diff
@@ -137,6 +137,7 @@ def gen_exec_command(
         This method must be implemented by subclasses.
 
         Args:
+            test_name (str): The name of the test.
```
amaslenn, line 140: This is not needed anymore, right?

TaekyungHeo (dismissed):

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
amaslenn, line 1: I suggest that instead of changing this file, we create a new strategy. We won't have that many if-else blocks which will make it easier to read.

Do you think it is possible?

TaekyungHeo:
Approved. Let's see if @amaslenn has any comments and wait for @srinivas212 to merge this.

@amaslenn, this PR supports two different subtests in the JAX Toolbox test templates. This allows us to support Grok and GPT simultaneously in a hierarchical manner. The related PR is https://github.com/Mellanox/cloudaix/pull/2.

amaslenn (changes requested):
This PR comes with zero tests, yet changes a lot. Please add unit tests.

src/cloudai/_core/test_template.py
srivatsankrishnan: Fixed

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: It can be done. You mean defining two classes or multiple ones depending upon however many new tests will get added part of JaxToolBox. 

Something like his
 -  `class JaxToolboxGptCommandGenStrategy(SlurmCommandGenStrategy)` for GPT  
 - `class JaxToolboxGrokCommandGenStrategy(SlurmCommandGenStrategy)` for Grok

The if-else block will moved at a high-level instead of `slurm_command_gen_stratergy`. Maybe okay since it will be choosing the right class of having that for each method now. 

I see this as a refactor of this code. I will create a separate PR for refactoring that code for readability.

srivatsankrishnan: @amaslenn Added bunch of unit tests. Please take a look

amaslenn: If we agree on two classes approach, I'd say we do it right now, in this PR. I'm pretty sure you'll have higher priority tasks (compared to refactoring) and code will remain like this for longer. Plus it will be seen as a pattern to use for implementing more cases like this.

Anyway, let me know what you think because you know this feature better, let's also take into account the deadlines we have for it.

amaslenn (dismissed):

TaekyungHeo (dismissed):

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -173,8 +169,9 @@ def _write_sbatch_script(self, args: Dict[str, Any], env_vars_str: str, srun_com
             "#!/bin/bash",
             f"#SBATCH --job-name={args['job_name']}",
             f"#SBATCH -N {args['num_nodes']}",
+            f"#SBATCH --gres=gpu:8"
```
TaekyungHeo: I am not sure whether it is okay to hard-code the number of GPUs per node. What if another cluster has 4 GPUs per node instead of 8? Could we come up with a more general approach?

TaekyungHeo (approved):

PR #144, opened 2024-07-10
Refactor `gen_exec_command` flow

amaslenn:
## Summary
Refactor `gen_exec_command` flow

1. Removed layer from `Test` class as it just passed all the args further to `TestTemplate`.
2. Introduced a `TestRun` object to decouple `Test` for the concrete execution of a test.
3. Rely more on `slurm_system` instead of `slurm_args`.

## Test Plan
CI

## Additional Notes
—

### DIFF
--- a/src/cloudai/_core/test.py
+++ b/src/cloudai/_core/test.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import sys
-from typing import Dict, List, Optional, Union
+from typing import Dict, Optional, Union
 
 from .job_status_result import JobStatusResult
 from .test_template import TestTemplate
@@ -41,8 +41,6 @@ class Test:
         sol (Optional[float]): Speed-of-light performance for reference.
         weight (float): The weight of this test in a test scenario, indicating its relative importance or priority.
         ideal_perf (float): The ideal performance value for comparison.
-        time_limit (Optional[str]): Time limit for the test specified as a string in "hh:mm:ss" format, or None if no
-            limit.
     """
 
     __test__ = False
@@ -59,12 +57,9 @@ class Test:
         section_name: str = "",
         dependencies: Optional[Dict[str, "TestDependency"]] = None,
         iterations: Union[int, str] = 1,
-        num_nodes: int = 1,
-        nodes: Optional[List[str]] = None,
         sol: Optional[float] = None,
         weight: float = 0.0,
         ideal_perf: float = 1.0,
-        time_limit: Optional[str] = None,
     ) -> None:
         """
         Initialize a Test instance.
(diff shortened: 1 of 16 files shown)


src/cloudai/_core/test.py
```diff
@@ -41,8 +41,6 @@ class Test:
         sol (Optional[float]): Speed-of-light performance for reference.
```
TaekyungHeo, line 41: We should remove num_nodes and nodes from the docstring as well.

src/cloudai/_core/test.py
```diff
@@ -122,30 +111,6 @@ def __repr__(self) -> str:
...
-            self.env_vars,
-            self.cmd_args,
-            self.extra_env_vars,
-            self.extra_cmd_args,
-            output_path,
-            self.num_nodes,
-            self.nodes,
         )
 
     def get_job_id(self, stdout: str, stderr: str) -> Optional[int]:
```
TaekyungHeo, line 116: If the TestRun class represents a test that is running, then we need to move these methods (get_job_id and get_job_status) to the TestRun class as well.

TaekyungHeo:
Let @srinivas212 review these changes.

TaekyungHeo:
Did you get a chance to generate commands using the dry-run mode to see if this does not break any existing test template command generation code?

amaslenn:
> Did you get a chance to generate commands using the dry-run mode to see if this does not break any existing test template command generation code?

I did it for some examples, but I haven't compared all of them.

TaekyungHeo:
@amaslenn  I will run some manual tests once Srinivas reviews this PR.

src/cloudai/_core/test.py
amaslenn: Yes, that is the plan. And believe there are more fields like this.

src/cloudai/_core/test.py
amaslenn: Done.

src/cloudai/_core/test.py
amaslenn: Actually, these to methods could be part of Runner class, seems the fit there better. But looks like `iterations` could be moved to `TestRun` as part of this PR, what do you think?

TaekyungHeo: It makes sense.

TaekyungHeo (changes requested):
I agree with introducing a new class, `class TestRun`. Please resolve the conflicts and then update `doc/DEV.md` according to the current design. Alternatively, you can revert the changes, and I will approve this PR.

amaslenn:
I merged PR with the latest `main` and moved doc update into #184.

TaekyungHeo (approved):

srivatsankrishnan (approved):
LGTM

PR #145, opened 2024-07-11
Introduce Pydantic to verify Test schema

amaslenn:
## Summary
Use Pydantic model for Test Template and Test TOMLs.

## Test Plan
1. Extended CI.
2. Dry-run and compare generated scripts for the following runs. All found issues were covered with unit tests. Compared vs `main` state:
```bash
cloudai --mode dry-run --system-config slurm.toml --test-templates-dir conf/common/test_template/ --tests conf/common/test/ --test-scenario conf/common/test_scenario/chakra_replay.toml
cloudai --mode dry-run --system-config slurm.toml --test-templates-dir conf/common/test_template/ --tests conf/common/test/ --test-scenario conf/common/test_scenario/nccl_test.toml
cloudai --mode dry-run --system-config slurm.toml --test-templates-dir conf/common/test_template/ --tests conf/common/test/ --test-scenario conf/common/test_scenario/sleep.toml
cloudai --mode dry-run --system-config slurm.toml --test-templates-dir conf/common/test_template/ --tests conf/common/test/ --test-scenario conf/common/test_scenario/ucc_test.toml
```

## Additional Notes
—

# Highlights for Release notes
We are working on schema improvements to simplify configs management and make them verifiable. This will help ensure that configs are correct before expensive runs on real hardware. Today we are enabling it for Test configs. This is a continuation of #158.

1. Test Template TOML files were replaced with Pydantic models. That ensures mandatory arguments as well as its types and requires less code to maintain.
2. `--test-templates-dir` option was removed for all commands. All supported tests are registered in code using `Registry().add_test_definition(...)` and `Registry().add_test_template(...)`. Documentation was updated to reflect this change.
3. Test TOML files now take advantage of standard TOML format for all know arguments.
    Before:
    ```toml
    [cmd_args]
    "training" = "llama/llama2_70b"
    "training.trainer.max_steps" = "120"
    "training.model.global_batch_size" = "256"
    "training.model.pipeline_model_parallel_size" = "1"
    ```
    Now:
    ```toml
    [cmd_args]
      [cmd_args.training]
      values = "llama/llama2_70b"
        [cmd_args.training.trainer]
        max_steps = "120"
        [cmd_args.training.model]
        global_batch_size = "256"
        pipeline_model_parallel_size = "2"
    ```
4. `extra_cmd_args` converted from `str` to `dict[str, str]`:
    Before:
    ```toml
    extra_cmd_args = "--stepfactor 2"
    ```
    Now:
    ```toml
    [extra_cmd_args]
    "--stepfactor" = "2"
    ```
5. Add a new mode to verify if Tests TOMLs are valid: `cloudai --mode verify-tests --system-config conf/common/system/standalone_system.toml --tests-dir conf/common/test/chakra_replay.toml`


TaekyungHeo:
If we are going to accept this pydantic PR, then we should update USER_GUIDE.md as well to inform contributors how to add a new test template. Before updating USER_GUIDE.md, let's ensure @srinivas212 is on the same page.

TaekyungHeo:
It's good to know that you have tested this PR with CloudAI test templates!

amaslenn:
> LGTM. However, since it is a major change, let's wait for @srinivas212.

Do you think we should keep all the existing field validators as is? This might constraint some of our users, I think.

TaekyungHeo (changes requested):
@amaslenn, let's discuss this PR in the next meeting. This PR might break other test templates as it changes how we retrieve arguments.

TaekyungHeo:
@amaslenn 
* CloudAIX needs to be updated accordingly. Please create a PR.
* At least a few test templates, such as NCCL tests and sleep, should be tested on a Slurm system. If you cannot find an idle node, ensure that you run the sleep test on a standalone system (e.g., your laptop). Please add the results to the test plan section.

TaekyungHeo:
It seems that you tested this by comparing the generated sbatch scripts. That makes sense. Please resolve the conflicts, and I can approve it.

amaslenn:
I need more time for testing it, so far we can agree/disagree with the approach in general.

TaekyungHeo:
@amaslenn , Okay, then I will approve when you confirm this PR is ready. Overall, I agree with this approach.

amaslenn:
@TaekyungHeo I re-ran verification according to this PR description, made a small correction and expanded testing for this case. Now it is ready for review and approval.

TaekyungHeo (approved):

TaekyungHeo:
Could you please resolve the conflicts, @amaslenn?

TaekyungHeo (dismissed):

TaekyungHeo (approved):

TaekyungHeo:
@srinivas212 's comment - We should share this with our customers before we merge it.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -556,19 +502,17 @@ def _generate_python_command(
...
             f'--fdl_config="{fdl_config}"',
         ]
 
-        # Dynamically adding fdl. prefixed arguments
-        fdl_prefix = f"{self.test_name}.fdl."
-        fdl_args = {k[len(fdl_prefix) :]: v for k, v in cmd_args.items() if k.startswith(fdl_prefix)}
+        fdl_args: dict[str, str] = cmd_args[f"{self.test_name}.fdl"]
 
-        for key, value in fdl_args.items():
+        for key, value in sorted(fdl_args.items()):
```
srivatsankrishnan, line 515: There is no specific requirement for ordering of fiddle flags or it should be parsed in a specific order? Any reason why this change is needed here?

amaslenn: Main reason to add that was to simplify generated scripts comparison against current state. For that I also made similar changes in `main` during my tests.

As a side affect, it should be easier to track CLI options inside our scripts if that is required.

To finalise, the reason is to simplify development.

srivatsankrishnan: Can we have the generated test script in the test plan? There is lot of changes in this PR esp with the fdl flags and xla flags. These flags basically are needed for performance convergence. I trust you when you say it’s easier to parse. But want to see the generated scripts. Maybe your cloudaix PR should have it?

amaslenn: That is what I did. I run mentioned in description commands to generate run scripts. Did it for these changes and for main state. Then compared file to file. Same for Cloud AI X, just commands were different. If you think it is needed, I will of course upload these files.

You can also run this version for your configs and see if everything is in place. As I mentioned, I didn't run all the scenarios, but for every found issue I added a unit test.

amaslenn:
> ... to replace them appropriately.

Great catch, I've missed that, thank you! Yet for `TestTemplate` it is a bit more complicated as we still have such python object it is used.

I've updated docs, please let me know if you see more issues with it.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -256,18 +226,13 @@ def generate_full_srun_command(
         """
         self._create_run_script(slurm_args, env_vars, cmd_args, extra_cmd_args)
 
-        start_container_run = str(cmd_args.get("load_container", "False")).lower() in ("true", "1", "yes")
+        start_container_run = cmd_args.get(f"{self.test_name}.load_container", False)
```
srivatsankrishnan, line 229: Why is the self.test_name getting appended here? This is causing a bug in the generated sbatch script.

```
[cmd_args]
docker_image_url = "xxxxxx"
fdl_config = "paxml.contrib.gpu.scripts_gpu.configs.Synthetic5B"
load_container = true
```

ideally we should be able to parse this with `cmd_args.get("load_container", "False"). During parsing the cmd_args actually appends the test_name before it.

amaslenn: Legacy from the previous approach. I've updated logic so we can access this key directly.

TaekyungHeo (approved):

srivatsankrishnan (approved):
Met with Andrei and call to discuss this PR. LGTM.

PR #146, opened 2024-07-11
Add job_status_check flag to disable checks for specific tests

TaekyungHeo:
## Summary
Add the job_status_check flag to the test scenario schema to ignore checking the job status. The flag is applied globally. Currently, if a test is killed or fails in the middle and does not produce a proper result, it is considered a job failure by the status-checking logic. As a result, all tests are terminated, and reports are not generated. This PR introduces a new flag, job_status_check, to the test scenario schema so that users can optionally enable or disable the job status check. By default, it is set to True, and it can be set to False if users wish. The flag is applied to all tests in the test scenario; you cannot optionally turn the check on or off for each test individually.

## Test Plan
example_test.toml
```
name = "nccl-test"
job_status_check = false

[Tests.1]
  name = "nccl_test_all_reduce"
  num_nodes = "1"
  time_limit = "00:20:00"

[Tests.2]
  name = "nccl_test_all_gather"
  num_nodes = "1"
  time_limit = "00:20:00"
```
When job_status_check is set to true, if nccl_test_all_reduce is killed midway, nccl_test_all_gather is also terminated. When it is set to false, if nccl_test_all_reduce is killed midway, nccl_test_all_gather is not affected.

### DIFF
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -290,17 +290,29 @@ class BaseRunner(ABC):
                     successful_jobs_count += 1
                     await self.handle_job_completion(job)
                 else:
-                    job_status_result = self.get_job_status(job)
-                    if job_status_result.is_successful:
+                    if self.test_scenario.job_status_check:
+                        job_status_result = self.get_job_status(job)
+                        if job_status_result.is_successful:
+                            successful_jobs_count += 1
+                            await self.handle_job_completion(job)
+                        else:
+                            error_message = (
+                                f"Job {job.id} for test {job.test.section_name} failed: "
+                                f"{job_status_result.error_message}"
+                            )
+                            logging.error(error_message)
+                            await self.shutdown()
+                            raise JobFailureError(job.test.section_name, error_message, job_status_result.error_message)
+                    else:
+                        job_status_result = self.get_job_status(job)
+                        if not job_status_result.is_successful:
+                            error_message = (
+                                f"Job {job.id} for test {job.test.section_name} failed: "
+                                f"{job_status_result.error_message}"
+                            )
+                            logging.error(error_message)
                         successful_jobs_count += 1
                         await self.handle_job_completion(job)
-                    else:
-                        error_message = (
-                            f"Job {job.id} for test {job.test.section_name} failed: {job_status_result.error_message}"
-                        )
-                        logging.error(error_message)
-                        await self.shutdown()
-                        raise JobFailureError(job.test.section_name, error_message, job_status_result.error_message)
 
         return successful_jobs_count
 
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -26,20 +26,23 @@ class TestScenario:
     Attributes
         name (str): Unique name of the test scenario.
         tests (List[Test]): Tests in the scenario.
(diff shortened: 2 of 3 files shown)


src/cloudai/_core/base_runner.py
```diff
@@ -288,17 +288,22 @@ async def monitor_jobs(self) -> int:
                 if self.mode == "dry-run":
                     successful_jobs_count += 1
                 else:
-                    job_status_result = self.get_job_status(job)
-                    if job_status_result.is_successful:
+                    if job.test.job_status_check:
```
amaslenn, line 291: What do you think about creating a derivative of `BaseRunner` and overriding `monitor_jobs` method? Test will need a way to specify which Runner to use, currently it is assumed from the System.

TaekyungHeo: We can do that, but I don't think it is a solution to the problem that I want to solve. This is the issue that I want to solve: https://nvidia.slack.com/archives/C06LC3PRHDZ/p1720706394214339. The problem is that when we check the job status of all tests upon completion, if a job is killed due to dependency, it does not have valid output. As a result, we need a way to ignore the job status of some tests.

TaekyungHeo:
1. First, this PR draft should be reviewed by @srinivas212. Since this PR requires changes in the schema, it needs to be reviewed by PIC.
2. Second, I can merge this PR after https://github.com/NVIDIA/cloudai/pull/144 is merged.

src/cloudai/_core/base_runner.py
amaslenn: If we can delay it till Monday, I suggest we have a call to discuss it. If urgent — go with version.

TaekyungHeo: I can wait. Let's discuss.

src/cloudai/_core/base_runner.py
```diff
@@ -288,17 +288,22 @@ async def monitor_jobs(self) -> int:
...
+                            await self.handle_job_completion(job)
+                        else:
+                            error_message = (
+                                f"Job {job.id} for test {job.test.section_name} failed: "
+                                f"{job_status_result.error_message}"
+                            )
+                            logging.error(error_message)
+                            await self.shutdown()
+                            raise JobFailureError(job.test.section_name, error_message, job_status_result.error_message)
+                    else:
```
srivatsankrishnan, line 304: @TaekyungHeo Here, can we add the job status even if we choose to ignore the failure and proceed? When running it in NVIS flow, even though it prevents an auto-exit on failure, it says "Job completed". I think capturing the error and logging would avoid that confusion.

srivatsankrishnan, line 304: @TaekyungHeo  Here, can we add the job status even if we choose to ignore the failure and proceed? When running it in NVIS flow, even though it prevents an auto-exit on failure, it says "Job completed". I think capturing the error and logging would avoid that confusion.

```
else:
                        job_status_result = self.get_job_status(job)
                        if not job_status_result.is_successful:
                            error_message = (
                                    f"Job {job.id} for test {job.test.section_name} failed: {job_status_result.error_message}"
                                )
                            logging.error(error_message)
                        successful_jobs_count += 1
                        await self.handle_job_completion(job)
```

srinivas212 (approved):

PR #150, opened 2024-07-15
Handle disk quota exceeded error in cache_docker_image method

TaekyungHeo:
## Summary
Handle disk quota exceeded error in cache_docker_image method

## Test Plan
Added unit tests.

### DIFF
--- a/src/cloudai/util/docker_image_cache_manager.py
+++ b/src/cloudai/util/docker_image_cache_manager.py
@@ -251,9 +251,22 @@ class DockerImageCacheManager:
 
         try:
             p = subprocess.run(enroot_import_cmd, shell=True, check=True, capture_output=True)
+            stdout = p.stdout.decode() if p.stdout else ""
+            stderr = p.stderr.decode() if p.stderr else ""
+
+            if "Disk quota exceeded" in stderr or "Write error" in stderr:
+                error_message = (
+                    f"Failed to cache Docker image {docker_image_url}. Command: {enroot_import_cmd}. "
+                    f"Error: '{stderr}'\n\n"
+                    "This error indicates a disk-related issue. Please check if the disk is full or not usable. "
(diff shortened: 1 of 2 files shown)


src/cloudai/util/docker_image_cache_manager.py
```diff
@@ -251,9 +251,22 @@ def cache_docker_image(
 
         try:
             p = subprocess.run(enroot_import_cmd, shell=True, check=True, capture_output=True)
+            stdout = p.stdout.decode() if p.stdout else ""
+            stderr = p.stderr.decode() if p.stderr else ""
```
amaslenn, line 255: You can do `subprocess.run(..., text=True)` to avoid manually calling `decode()`.

src/cloudai/util/docker_image_cache_manager.py
```diff
@@ -251,9 +251,22 @@ def cache_docker_image(
...
+
+            if "Disk quota exceeded" in stderr or "Write error" in stderr:
+                error_message = (
+                    f"Failed to cache Docker image {docker_image_url}. Command: {enroot_import_cmd}. "
+                    f"Error: '{stderr}'\n\n"
+                    "This error indicates a disk-related issue. Please check if the disk is full or not usable. "
+                    "If the disk is full, consider using a different disk or removing unnecessary files."
```
amaslenn, line 262: [thought] Do we want to have suggestions right inside error messages? While it is useful, maybe we want to have a single place to map errors to suggestions, like a README file.

TaekyungHeo: Let's discuss this with @srinivas212. While it is a good idea, I would like to keep it in the error messages because I don't think users will read README.md. Users tend to report issues or ping us directly.

amaslenn (approved):

PR #152, opened 2024-07-16
Enhance error messages to provide guidance for missing schemas

TaekyungHeo:
## Summary
Enhance error messages to provide guidance for missing schemas

## Test Plan
CI passes


amaslenn (approved):
I think we should only load what is specified in Test Scenario. This would make this error very explicit. And users would be able to develop own configs without breaking the system. Let's keep this in mind, it should be possible when we migrate to schemas.

PR #161, opened 2024-07-24
add support for NeMo Launcher in the reservation

jeffnvidia:
## Summary
This PR is based on andrei's : https://github.com/NVIDIA/cloudai/pull/160
It enables reservation flag to be transmitted to NeMoLauncher

## Test Plan

system.toml

```
extra_srun_args = "--reservation my-reservation"
``` 

command 

```
$ python ./cloudaix.py --mode run --system-config ... --test-templates-dir conf/v0.6/general/test_template --tests-dir conf/v0.6/general/test --test-scenario conf/v0.6/general/test_scenario/llama/llama3_70B.toml
...

Section Name: Tests.1
Test Name: llama3_70B
Description: llama
No dependencies
[INFO] Initializing Runner
[INFO] Creating SlurmRunner
[INFO] Starting test scenario execution.
[INFO] Starting test: Tests.1
[INFO] Running test: Tests.1
...
```

### DIFF
--- a/src/cloudai/schema/test_template/nemo_launcher/slurm_command_gen_strategy.py
+++ b/src/cloudai/schema/test_template/nemo_launcher/slurm_command_gen_strategy.py
@@ -60,7 +60,7 @@ class NeMoLauncherSlurmCommandGenStrategy(SlurmCommandGenStrategy):
         self.final_cmd_args["cluster.partition"] = self.slurm_system.default_partition
         reservation_key = "--reservation "
         if self.slurm_system.extra_srun_args and reservation_key in self.slurm_system.extra_srun_args:
-            reservation = extra_cmd_args.split(reservation_key, 1)[1].split(" ", 1)[0]
+            reservation = self.slurm_system.extra_srun_args.split(reservation_key, 1)[1].split(" ", 1)[0]
             self.final_cmd_args["+cluster.reservation"] = reservation
         nodes = self.slurm_system.parse_nodes(nodes)
         if nodes:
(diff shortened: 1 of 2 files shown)


src/cloudai/schema/test_template/nemo_launcher/slurm_command_gen_strategy.py
```diff
@@ -58,6 +58,10 @@ def gen_exec_command(
         for key, value in final_env_vars.items():
             self.final_cmd_args[f"env_vars.{key}"] = value
         self.final_cmd_args["cluster.partition"] = self.slurm_system.default_partition
+        reservation_key = "--reservation "
+        if self.slurm_system.extra_srun_args and reservation_key in self.slurm_system.extra_srun_args:
+            reservation = extra_cmd_args.split(reservation_key, 1)[1].split(" ", 1)[0]
```
amaslenn, line 63: Should it be `self.slurm_system.extra_srun_args` not `extra_cmd_args`?

Anyway, it seems that we can add `extra_srun_args = "--reservation name"` into a System TOML and also add `extra_cmd_args = "+cluster.reservation=name"` into a Test TOML? This way we don't need to change python code.

jeffnvidia: yes thanks for noticing the error

well the problem with adding it in the test scenario is that it's not part of the test, it's dependant on a specific run.

What do you think @lappazos @srinivas212 ?

amaslenn: `extra_cmd_args` is part of Test TOML, not Scenario TOML. So that means you'll have to modify multiple files if follow my suggestion, which is not user-friendly.

Let's go with your solution, but please expand `TestNeMoLauncherSlurmCommandGenStrategy__GenExecCommand` tests to cover your case.

jeffnvidia: right I meant to say test not test scenario, my bad.

Alright sure I'll try to cover it today

jeffnvidia: > `extra_cmd_args` is part of Test TOML, not Scenario TOML. So that means you'll have to modify multiple files if follow my suggestion, which is not user-friendly.
> 
> Let's go with your solution, but please expand `TestNeMoLauncherSlurmCommandGenStrategy__GenExecCommand` tests to cover your case.

updated

amaslenn (approved):

amaslenn (approved):

PR #170, opened 2024-08-09
Add Pre-test for long LLMs run on Coreweave

srivatsankrishnan:
## Summary
Adding NCCL benchmarking test before launching large node runs on CW.  Long runs (2k-4k LLM jobs) needs to be resilient. The goal is to perform a NCCL test as a pre-test before the actual training jobs are run. This PR adds support for running NCCL jobs on the same nodes before the actual training starts using in CloudAI.  The running of long LLM training jobs is conditional upon the pre-test. If the pre-test fails due to some network issues, the LLMs job will not start. If the pre-test passes, the LLM job will gets scheduled.

The pre-test will be defined in the test template of the LLM job. 

## Test Plan

- Ensure the commands are generated correctly via cloudAI.
```
#Sbatch generated file
...
...
...
srun \
--mpi=pmix \
-o output_pretest-%j-%n-%t.txt \
-e error_pretest-%j-%n-%t.txt \
--container-image=nvcr.io/nvidia/pytorch:24.02-py3 \
/usr/local/bin/all_gather_perf_mpi \
--nthreads 1 \
--ngpus 256 \
--minbytes 8M \
--maxbytes 16G \
--stepbytes 1M \
--op sum \
--datatype float \
--root 0 \
--iters 20 \
--warmup_iters 5 \
--agg_iters 1 \
--average 1 \
--parallel_init 0 \
--check 1 \
--blocking 1 \
--cudagraph 0 \
--stepfactor 2

file_pattern="output_pretest-*.txt"
keyword="Avg bus bandwidth"

# Use grep to search for the keyword in the files
if grep -q "$keyword" $file_pattern; then
    keyword_found=true
fi

if [ "$keyword_found" = true ]; then
    srun \
    --mpi=pmix \
     \
    --export=ALL \
    -o output-%j-%n-%t.txt \
    -e error-%j-%n-%t.txt \
    --container-image=jax_toolbox.sqsh \
    /opt/paxml/workspace/run.sh
fi

```
- Test when the NCCL pre-test is successful. Expected behavior, both the NCCL and LLM jobs should behave as if they are run independently including its performance numbers.

**Successful Run of NCCL Pre-test**
```
# nThread 1 nGpus 1 minBytes 8388608 maxBytes 17179869184 step: 2(factor) warmup iters: 5 iters: 20 agg iters: 1 validation: 1 graph: 0
# Blocking Enabled: wait for completion and barrier after each collective 
#
# Using devices
#  Rank  0 Group  0 Pid 1581435 on hgx-isr1-048 device  0 [0x19] NVIDIA H100 80GB HBM3
#  Rank  1 Group  0 Pid 1581436 on hgx-isr1-048 device  1 [0x3b] NVIDIA H100 80GB HBM3
#  Rank  2 Group  0 Pid 1581437 on hgx-isr1-048 device  2 [0x4c] NVIDIA H100 80GB HBM3
#  Rank  3 Group  0 Pid 1581438 on hgx-isr1-048 device  3 [0x5d] NVIDIA H100 80GB HBM3
#  Rank  4 Group  0 Pid 1581439 on hgx-isr1-048 device  4 [0x9b] NVIDIA H100 80GB HBM3
#  Rank  5 Group  0 Pid 1581440 on hgx-isr1-048 device  5 [0xbb] NVIDIA H100 80GB HBM3
#  Rank  6 Group  0 Pid 1581441 on hgx-isr1-048 device  6 [0xcb] NVIDIA H100 80GB HBM3
#  Rank  7 Group  0 Pid 1581442 on hgx-isr1-048 device  7 [0xdb] NVIDIA H100 80GB HBM3
#
#                                                              out-of-place                       in-place          
#       size         count      type   redop    root     time   algbw   busbw #wrong     time   algbw   busbw #wrong
#        (B)    (elements)                               (us)  (GB/s)  (GB/s)            (us)  (GB/s)  (GB/s)       
     8388608        262144     float    none      -1    xxx  xxx   xxx      0    xxx  xxx  xxx     0
    16777216        524288     float    none      -1    xxx  xxx  1xxx      0    xxx  xxx  xxx      0
    33554432       1048576     float    none      -1    1xx  xxx  xxx      0    xxx  xxx  xxx      0
    67108864       2097152     float    none      -1    xxx  xxx  xxx      0    xxx  xxx  xxx      0
   134217728       4194304     float    none      -1    xxx  xxx xxx      0    xxx  xxx  xxx      0
   268435456       8388608     float    none      -1    xxx  xxx  xxx      0    xxx  xxx xxx      0
   536870912      16777216     float    none      -1   xxxxxxx7     xxx 0  xx x  xx  x      0
  1073741824      33554432     float    none      -1   xxx  xxx  xxx      0   xxx  xxx  xxx      0
  2147483648      67108864     float    none      -1   xxx  xxx  xxx      0   xxx  xxx  xxx      0
  4294967296     134217728     float    none      -1    xxx  xxx  xxx      0    xxx  xxx  xxx      0
  8589934592     268435456     float    none      -1    xxxx  xxx  xxx      0    xxx  xxx  xxx      0
 17179869184     536870912     float    none      -1    xxx  xxx  xxx      0    xxx  xxx  xxx      0
# Out of bounds values : 0 OK
# Avg bus bandwidth    : xxx 
```
** Successful Run of LLM training Job**

```
Min: 0.451397
Max: 0.477019
Average: 0.4640926125
Median: 0.4626145
Stdev: 0.0067373331133954205
```

- Test when NCCl-pre-test fails. This should log the errors related to pre-test but should not start the LLM training.

**Unsuccessful NCCL Pre-test**
```
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_reset_if_to: hgx-isr1-008 [2]: pmixp_coll_ring.c:741: 0x1458100093c0: collective timeout seq=0
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_log: xx-xxx-008xx [2]: pmixp_coll.c:281: Dumping collective state
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxx-xxx-008 [2]: pmixp_coll_ring.c:759: 0x1458100093c0: COLL_FENCE_RING state seq=0
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxx-xxx-0xx08 [2]: pmixp_coll_ring.c:762: my peerid: 2:xxxx8
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxxxx [2]: pmixp_coll_ring.c:769: neighbor id: next 3:xxx prev 1:xxxx
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxxx [2]: pmixp_coll_ring.c:778: Context ptr=0x145810009438, #0, in-use=0
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxxx[2]: pmixp_coll_ring.c:778: Context ptr=0x145810009470, #1, in-use=0
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log:xxxx [2]: pmixp_coll_ring.c:778: Context ptr=0x1458100094a8, #2, in-use=1
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxxxcoll_ring.c:787: 	 seq=0 contribs: loc=0/prev=10/fwd=10
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log:xxxx [2]: pmixp_coll_ring.c:791: 	 neighbor contribs [32]:
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log:xxxx [2]: pmixp_coll_ring.c:824: 		 done contrib: xxxxx]
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: hgx-isr1-008 [2]: pmixp_coll_ring.c:826: 		 wait contrib: xxx-xxxxx
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: xxxx [2]: pmixp_coll_ring.c:828: 	 status=PMIXP_COLL_RING_PROGRESS
slurmstepd: error:  mpi/pmix_v3: pmixp_coll_ring_log: hgx-isr1-008 [2]: pmixp_coll_ring.c:831: 	 buf (offset/size): 91878/388550
```

** LLM Job Killed**

```
srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
srun: error: xxx tasks 96-98,100-103: Killed
srun: error: xxx: tasks 80-81,83-87: Killed
srun: error: xxx: tasks 88,90-95: Killed
srun: error: xxxx: tasks 200-207: Killed
srun: error: xxx tasks 24-31: Killed
srun: error: xxx: tasks 104-111: Killed
srun: error: xxx: tasks 56-63: Killed
srun: error: xxxx-20,22: Killed
srun: error: xxx: task 89: Killed
srun: error: hgx-isr1-006: tasks 2,4-7: Killed
...
...
...
```
- CI/CD pipeline should pass.


src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -193,21 +194,129 @@ def _parse_slurm_args(
...
         srun_command_parts = [
             "srun",
-            f"--mpi={self.slurm_system.mpi}",
-            f"{self.slurm_system.extra_srun_args if self.slurm_system.extra_srun_args else ''}",
-            "--export=ALL",
-            f"-o {slurm_args['output']}",
-            f"-e {slurm_args['error']}",
-            f"--container-image={slurm_args['image_path']}",
-            f"--container-mounts={slurm_args['container_mounts']}",
+            f"--mpi={self.slurm_system.mpi} \\",
```
TaekyungHeo, line 228: Do we really need to add "\\"? Since we are using "join," can we include it as part of the arguments on line 239?

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -46,6 +46,7 @@ def gen_exec_command(
         final_cmd_args["output_path"] = output_path
 
         self.test_name = self._extract_test_name(cmd_args)
+        self.pre_test_value = None
```
TaekyungHeo, line 49: Two comments:
- Can we use a more descriptive name, such as run_pre_test?
- Let's ensure that we use a boolean type for this variable, confirming that it is either True or False.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -193,21 +194,129 @@ def _parse_slurm_args(
...
         self._create_run_script(slurm_args, env_vars, cmd_args, extra_cmd_args)
 
+        output_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "output_pretest-%j-%n-%t.txt")
+        error_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "error_pretest-%j-%n-%t.txt")
+
+        commands = []
+
+        pre_test_value = cmd_args.get("pre_test")
+
+        if pre_test_value == "True":
```
TaekyungHeo, line 218: I understand that it will work, but let's ensure that the value is either True or False, not any string values like "True". We have a helper function, as shown [here](https://github.com/NVIDIA/cloudai/blob/570181a5137c2defcddfa5904e64b41028bf77dd/src/cloudai/parser/system_parser/slurm_system_parser.py#L48). It might be a good idea to spin this off as a separate helper function.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -426,25 +426,53 @@ def setup_method(self):
...
     def test_only_mandatory_args(self, strategy_fixture: SlurmCommandGenStrategy, tmp_path: Path):
+        # Ensure MANDATORY_ARGS includes all the arguments you expect to be present
+        args = {
+            "job_name": "test_job",
+            "num_nodes": 2,
+            "partition": "default",
+            "node_list_str": "node1,node2",
+        }
+
+        print(f"Args passed to the method: {args}")  # Debugging line
```
TaekyungHeo, line 472: Let's remove any print statements in pytest.

TaekyungHeo (changes requested):

TaekyungHeo:
I understand that you can run pre-tests with this PR, but how do you determine if they were successful or not? How should CloudAI behave in the event of success or failure? Will it be recorded somewhere? Should the job be scheduled again on failure?

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: Fixed.
https://github.com/NVIDIA/cloudai/pull/170/commits/356e30464a35ebbc5df6d0ca3eb4b008b168cc06

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: Fixed

srivatsankrishnan:
@TaekyungHeo Regarding the behavior, take a look at test plan on what happens when the pre-test fails or what happens when it passes. @srinivas212 mentioned this is just to unblock internal benchmarking efforts. The plugin feature should ideally handle those cases you are mentioning. It needs a separate discussion.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: Sure, we can add it to join or here. Same thing.

But I wanted to enforce the "\" in the generated srun commands. So added to the string generation itself. But yeah I see your point.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: True. The parser has the type to check this. But in command gen, it doesn't have the type anymore. But I added a one line checker to convert this to bool to enforce it instead of string.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -193,8 +193,35 @@ def _parse_slurm_args(
...
+            str: The full srun command as a string.
+        """
         self._create_run_script(slurm_args, env_vars, cmd_args, extra_cmd_args)
 
+        output_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "output_pretest-%j-%n-%t.txt")
+        error_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "error_pretest-%j-%n-%t.txt")
+
+        commands = []
+
+        run_pre_test = cmd_args.get("pre_test", "False").lower() in ("true", "1", "yes")
```
TaekyungHeo, line 215: Two comments:
* Does it work even when the `pre_test` value is a boolean (true) and not a string ("True")?
* Would it be better to rename the schema field to `run_pre_test`? I'm not fully confident. Please provide your feedback, @srinivas212.

srivatsankrishnan: I think `cmd_args` has type of `Dict[str, str]`. At this stage, that value is basically a string and not a boolean. I think this method should work with a caveat being how strict the type checking tool is. We use Pyright and it didn't flag it.  But honestly this is a simple fix if we are not confident about how strongly type checking is enforced.

Fix would be something link this (should cover both cases)

```
pre_test_value = cmd_args.get("pre_test", "False")

if isinstance(pre_test_value, bool):
    run_pre_test = pre_test_value
else:
    run_pre_test = str(pre_test_value).lower() in ("true", "1", "yes")
```

Regarding the name in the schema, I don't have any strong preference of what name it should be. Either should work imo.

TaekyungHeo: > I think cmd_args has type of Dict[str, str]. At this stage, that value is basically a string and not a boolean. I think this method should work with a caveat being how strict the type checking tool is. We use Pyright and it didn't flag it. But honestly this is a simple fix if we are not confident about how strongly type checking is enforced.

Okay, I agree, and this makes sense if that is the case. However, I am not sure if treating true/false values as strings is the best approach. We can revisit this item later.

> Regarding the name in the schema, I don't have any strong preference of what name it should be. Either should work imo.
Same for me. Either is fine. Let Srinivas review this item, close the discussion, and merge this PR.

srivatsankrishnan: > Okay, I agree, and this makes sense if that is the case. However, I am not sure if treating true/false values as strings is the best approach. We can revisit this item later.

Great, thanks. Looking at the history of that file, seems like it was always `Dict[str, str]`. And we do strip the type field. Good to revisit this.

amaslenn:
> ... The plugin feature should ideally handle those cases you are mentioning. It needs a separate discussion.

Can we use our existing dependency feature for this pre-test?

srivatsankrishnan:
> > ... The plugin feature should ideally handle those cases you are mentioning. It needs a separate discussion.
> 
> Can we use our existing dependency feature for this pre-test?

Yeah that was the first thought actually. Dependencies means the test gets scheduled one after the other from orchestration point of view. That means there is a chance :

- you might lose your position in the queue and wait longer. High chance in cluster like CW.
- Node allocation: Since these LLM at this scale training are complex and will be run very large scale clusters, you request for the "x" (x>512 - 4096) Gpus and let the scheduler handle the allocation. Not possible to hardcode this in test scenario to get the same node allocation.

So this approach basically takes care of this by adding the tests part of same sbatch script thus using the same node allocation (which is kind of a requirement to test the stability of these nodes before commencing the LLM training)

The plugin feature we envision should take care of the above along with scenarios. It needs more discussion as I mentioned. This is to unblock E2E performance scaling studies in the meantime.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -426,25 +426,51 @@ def setup_method(self):
 
     def assert_positional_lines(self, lines: list[str]):
```
amaslenn, line 427: With this update this function name doesn't make much sense anymore. Can we stick to positional checks? We can also switch to use `_append_sbatch_directives` directly for `sbatch` args.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -426,25 +426,51 @@ def setup_method(self):
...
         args = self.MANDATORY_ARGS.copy()
         del args[missing_arg]
 
         with pytest.raises(KeyError) as exc_info:
             strategy_fixture._write_sbatch_script(args, self.env_vars_str, self.srun_command, str(tmp_path))
-        assert f"KeyError('{missing_arg}')" in str(exc_info)
+        assert missing_arg in str(exc_info.value)
 
     def test_only_mandatory_args(self, strategy_fixture: SlurmCommandGenStrategy, tmp_path: Path):
+        # Ensure MANDATORY_ARGS includes all the arguments you expect to be present
```
amaslenn, line 464: Could you please explain why is this change needed?

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -175,20 +176,48 @@ def _write_sbatch_script(self, args: Dict[str, Any], env_vars_str: str, srun_com
             f"#SBATCH -N {args['num_nodes']}",
         ]
 
+        self._append_sbatch_directives(batch_script_content, args, output_path)
+
+        batch_script_content.append(env_vars_str)
```
amaslenn, line 181: I liked empty lines as separators between these blocks, IMO it improves readability.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -175,20 +176,48 @@ def _write_sbatch_script(self, args: Dict[str, Any], env_vars_str: str, srun_com
...
+            args (Dict[str, Any]): Arguments including job settings.
+            output_path (str): Output directory for script and logs.
+        """
         if "output" not in args:
             batch_script_content.append(f"#SBATCH --output={os.path.join(output_path, 'stdout.txt')}")
         if "error" not in args:
             batch_script_content.append(f"#SBATCH --error={os.path.join(output_path, 'stderr.txt')}")
-        if args["partition"]:
+
+        # safely access parition
```
amaslenn, line 206: Not sure if it is needed, `_parse_slurm_args` raises if partition is not set, it doesn't make sense writing sbatch script without it. Not important, but the comment is not needed here.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -207,7 +234,88 @@ def generate_full_srun_command(
...
+        Returns:
+            str: The generated pre-test command.
+        """
+        nccl_test = {k.split(".")[-1]: v for k, v in cmd_args.items() if k.startswith("pre_test.nccl_test")}
+        pre_test_command_parts = [
+            "srun",
+            "--mpi=pmix",
+            f"-o {output_path}",
+            f"-e {error_path}",
+            f"--container-image={nccl_test.get('docker_image_url', 'nvcr.io/nvidia/pytorch:24.02-py3')}",
```
amaslenn, line 269: IMO we should use `self.docker_image_cache_manager.ensure_docker_image(...)` here as this image could be cached and we don't want to download it again.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -207,7 +234,88 @@ def generate_full_srun_command(
             "/opt/paxml/workspace/run.sh",
         ]
 
-        return " \\\n".join(srun_command_parts)
+        # Join the srun command parts with newlines and backslashes for readability
+        srun_command = " \\\n".join(srun_command_parts).strip()
+
+        # Add conditional check if pre_test_value is True
+        if run_pre_test:
```
amaslenn, line 241: We need a separate test for it. Probably exact commands and args are not very important, but we need to ensure this flag is handled.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -207,7 +234,88 @@ def generate_full_srun_command(
             "/opt/paxml/workspace/run.sh",
         ]
 
-        return " \\\n".join(srun_command_parts)
+        # Join the srun command parts with newlines and backslashes for readability
+        srun_command = " \\\n".join(srun_command_parts).strip()
+
+        # Add conditional check if pre_test_value is True
```
amaslenn, line 240: This commend doesn't add much value in this context, please remove.

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: The hardcoding of line number  or exact positioning doesn't make sense. It's not that it's a feature mandated by Slurm directive either.  I can change the function name for this test. Goal is to enable feature and have flexibility. Not to fine tune the logic because of a unit test.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: This container part of pre-test is technically not managed by CloudAI fully. Since this is going into the generated sbatch script, it is mostly managed at that level. When we get to the plugin feature, lets revisit this.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: If we don't specify the partition, it will go to default.  In CW, I let my account access to define that for me. So it works btw. That line was already there.  So I don't see why there was a check  exists if it didn't make sense to write one without partition in the first place. 

https://github.com/NVIDIA/cloudai/blob/main/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py#L182

This is just an additional check. I don't see it doing any harm.

tests/test_slurm_command_gen_strategy.py
amaslenn: Right, Slurm doesn't care about lines ordering, but this will pass the check, yet won't work: `#SBATCH ... #SBATCH ... #SBATCH ...`. So this line hardcoding is to make sure _our_ code is correct.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: I believe partition is required for Cloud AI internals, so this check was introduced. 

Anyway, what I mean is that using safe `agrs.get(...)` doesn't really add any safety compared to `args[...]` because it is anyway verified before.

Not super important, but double-check could be confusing for readers of the code.

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: Sure but I think this would still fail. It's actually checking for the whole line after #SBATCH. If you only have` #SBATCH ... #SBATCH ... #SBATCH`  this would fail too? Given we have other unit tests to check for mandatory args. It definitely has to enter one of those conditions and fail that assertion.

In general I don't think the test need to enforce in what order sbatch directives needs to be generated given Slurm doesn't enforce that. Except that `#!/bin/bash` line Slurm doesn't enforce anything. Hardcoding that line means in that order means one has to be aware of the unit test and modify their feature logic accordingly. That's like working backwards. We need to add feature and modify the test if needed. Not the other way around IMO.

Let's step back....the method name and logic are divergent. I agree with you on that. I can fix that. But I don't think hardcoding the line numbers is the right thing to do.

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: Right. Not needed. I just needed to test some arguments and passed my own instead of using instance variable defined at the top.

srivatsankrishnan: Fixed it.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: Didn't quite follow this comment.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: IMO getter method handles the key error and returns None instead if throwing KeyError in the latter. Anyway not inclined either way. Reverted back.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: done

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: done

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: Sorry, let me clarify. Before the change, it was like this: `batch_script_content.extend(["", env_vars_str, "", srun_command])`. This adds two empty lines between logical blocks (sbatch and env vars, env vars and srun args). IMO it contributed to readability of the final sbatch script.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -205,8 +229,11 @@ def _write_sbatch_script(self, args: Dict[str, Any], env_vars_str: str, srun_com
             batch_script_content.append(f"#SBATCH --account={args['account']}")
         if "distribution" in args:
             batch_script_content.append(f"#SBATCH --distribution={args['distribution']}")
-        if "gpus_per_node" in args:
+
+        gpus_per_node = args.get("gpus_per_node")
+        if gpus_per_node:
             batch_script_content.append(f"#SBATCH --gpus-per-node={args['gpus_per_node']}")
+            batch_script_content.append(f"#SBATCH --gres=gpu:{args['gpus_per_node']}")
```
amaslenn, line 236: _[minor]_ Using `gpus_per_node` instead of `args["gpus_per_node"]` will improve redability. (same for the line above)

tests/test_slurm_command_gen_strategy.py
```diff
@@ -185,6 +185,47 @@ def test_generate_full_srun_command(self, strategy_fixture: SlurmCommandGenStrat
         full_srun_command = strategy_fixture.generate_full_srun_command({}, {}, {}, "")
         assert full_srun_command == " \\\n".join(["srun", "--test", "test_arg", "test_command"])
 
+    def test_generate_full_srun_command_with_run_pre_test_flag(
+        self, jax_strategy_fixture: JaxToolboxSlurmCommandGenStrategy, slurm_system: SlurmSystem
+    ):
+        # Assign the slurm_system to the strategy fixture
+        jax_strategy_fixture.slurm_system = slurm_system
+
+        # Mock necessary methods and attributes
```
amaslenn, line 194: Would it make sense moving this code into the fixture itself? This will make the test easier to read and work with. Assert could be transformed into `assert jax_strategy_fixture._generate_pre_test_command.return_value in result`.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -185,6 +185,47 @@ def test_generate_full_srun_command(self, strategy_fixture: SlurmCommandGenStrat
         full_srun_command = strategy_fixture.generate_full_srun_command({}, {}, {}, "")
         assert full_srun_command == " \\\n".join(["srun", "--test", "test_arg", "test_command"])
 
+    def test_generate_full_srun_command_with_run_pre_test_flag(
+        self, jax_strategy_fixture: JaxToolboxSlurmCommandGenStrategy, slurm_system: SlurmSystem
+    ):
+        # Assign the slurm_system to the strategy fixture
```
amaslenn, line 191: This is my bias, but I really don't like comments that don't improve the understanding of the code. This is one of them. During development such comments make sense, I often write them myself, but for the maintenance such comments become a burden as it is easy to change the code and forget to change to comment.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -185,6 +185,47 @@ def test_generate_full_srun_command(self, strategy_fixture: SlurmCommandGenStrat
...
+        }
+        env_vars = {}
+        extra_cmd_args = ""
+
+        # run_pre_test is True
+        cmd_args = {"output_path": "/path/to/output", "pre_test": "true"}
+        result = jax_strategy_fixture.generate_full_srun_command(slurm_args, env_vars, cmd_args, extra_cmd_args)
+        assert "pre_test_command" in result
+        assert "pre_test_check_command" in result
+        assert "--mpi=fake-mpi" in result
```
amaslenn, line 213: IMO we can make this test more reliable with `assert f"--mpi={slurm_system.mpi}" in result` and similar for the rest two asserts.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -185,6 +185,47 @@ def test_generate_full_srun_command(self, strategy_fixture: SlurmCommandGenStrat
         full_srun_command = strategy_fixture.generate_full_srun_command({}, {}, {}, "")
         assert full_srun_command == " \\\n".join(["srun", "--test", "test_arg", "test_command"])
 
+    def test_generate_full_srun_command_with_run_pre_test_flag(
```
amaslenn, line 188: I'd recommend splitting this test into two: `test_..._with_pre_test` and `test_..._without_pre_test`.
1. You can define a special JaxStrategy fixture just for these two tests maximising code reuse.
2. It will be more unit-test like approach.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: Done

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: done

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: Can be done. Would move this discussion to seperate refactor PR as we discussed in the call. Unit tests unless broken or doesn't provide coverage shouldn't block the feature request. If there is anything functionally broken, let me know. Happy to add it.

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: done.

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: Subjective comments. But, point taken.

srivatsankrishnan: done

tests/test_slurm_command_gen_strategy.py
srivatsankrishnan: done

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -193,8 +193,40 @@ def _parse_slurm_args(
...
+            cmd_args (Dict[str, str]): A dictionary containing command arguments.
+            extra_cmd_args (str): Additional command arguments.
+
+        Returns:
+            str: The full srun command as a string.
+        """
         self._create_run_script(slurm_args, env_vars, cmd_args, extra_cmd_args)
 
+        output_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "output_pretest-%j-%n-%t.txt")
+        error_path = os.path.join(os.path.abspath(cmd_args["output_path"]), "error_pretest-%j-%n-%t.txt")
```
TaekyungHeo, line 211: Let's use pathlib.Path for path-related variables whenever possible. In PR183, @amaslenn and I attempted to replace all instances of os.path with pathlib.Path as shown ㅁㅅ https://github.com/NVIDIA/cloudai/pull/183/commits/1f5b36edb2761d09dc9b1a026d39ea442c3ee545.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -194,7 +195,29 @@ def _write_sbatch_script(
...
+    def _append_sbatch_directives(
+        self, batch_script_content: List[str], args: Dict[str, Any], output_path: Path
+    ) -> None:
+        """
+        Append SBATCH directives to the batch script content.
+
+        Args:
+            batch_script_content (List[str]): The list of script lines to append to.
+            args (Dict[str, Any]): Arguments including job settings.
+            output_path (str): Output directory for script and logs.
```
TaekyungHeo, line 217: The docstring is outdated (str -> Path).

TaekyungHeo (changes requested):

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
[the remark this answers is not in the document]
srivatsankrishnan: @TaekyungHeo  updated it. please take a look.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -207,7 +239,86 @@ def generate_full_srun_command(
...
+            srun_command = f'if [ "$keyword_found" = true ]; then\n{srun_command}\nfi'
+
+        commands.append(srun_command)
+
+        # Join all commands into the final full command string
+        full_command = "\n\n".join(commands)
+
+        return full_command
+
+    def _generate_pre_test_command(self, cmd_args: Dict[str, Any], output_path: str, error_path: str) -> str:
```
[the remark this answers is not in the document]
srivatsankrishnan: done. take a look @TaekyungHeo

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: updated it.

TaekyungHeo (approved):

amaslenn (approved):

PR #172, opened 2024-08-14
Fix generated html graph for constant message size

ybenvidia:
## Summary
We've noticed that when we run tests with a constant message size, the graph is not generated correctly because it expects a different x_min and x_max. We've modified the code to fix this issue.

## Test Plan
Test any test scenario with test having constant message (minbytes = maxbytes) size such as :

```
name = "nccl_test_alltoall"
description = "alltoall"
test_template_name = "NcclTest"

[cmd_args]
"subtest_name" = "alltoall_perf_mpi"
"ngpus" = "1"
"minbytes" = "8G"
"maxbytes" = "8G"
"stepbytes" = "0"
```

Before the fix:
image

After the fix :
image


TaekyungHeo:
Could you please attach the generated plot to this PR?

ybenvidia:
Before the fix :
Screenshot 2024-08-14 at 10 09 11

After the fix :
Screenshot 2024-08-14 at 14 22 16

TaekyungHeo:
Thank you for your contribution, @ybenvidia. I believe it's a valid point to include a plot showing bandwidth across iterations. However, I have a concern. The current behavior is inconsistent in generating different types of plots based on the number of data points. Instead, I think it's better to always generate two types of plots: one for bandwidth across message sizes and another for bandwidth across iterations. @srinivas212, what are your thoughts?

ybenvidia:
If the message size is constant, you can't display a single plot following iterations, as the size will never change. The x-axis will always have the same value if you try to display the band as a function of the message size (hence the current display error). That's why I've differentiated in the code the case where x_min=x_max and the case where the message size changes through iterations.

srinivas212:
> Before the fix : Screenshot 2024-08-14 at 10 09 11
> 
> After the fix : Screenshot 2024-08-14 at 14 22 16

Pl add to the test plan itself next time. Thanks!

src/cloudai/report_generator/tool/bokeh_report_tool.py
```diff
@@ -195,13 +201,26 @@ def add_log_x_linear_y_single_line_plot(
         x_min, x_max = self.find_min_max(df, x_column)
         y_min, y_max = self.find_min_max(df, y_column, sol)
 
+        # Check if x_min equals x_max - constant message size
+        if x_min == x_max:
```
amaslenn, line 205: We are adding identical changes into `add_log_x_linear_y_single_line_plot` and `add_log_x_linear_y_multi_line_plot`. Is there any reason we have them as two separate functions? Looks like _single_ could be removed and we can call _multi_ with a list of a single element.

src/cloudai/report_generator/tool/bokeh_report_tool.py
```diff
@@ -195,13 +201,26 @@ def add_log_x_linear_y_single_line_plot(
...
+            # Use iteration number as x-axis
+            df["iteration"] = range(1, len(df) + 1)
+            x_column = "iteration"
+            x_axis_label = "Iteration"
+            x_axis_type = "linear"
+            x_range = Range1d(start=1, end=len(df))
+        else:
```
amaslenn, line 212: _[minor]_ instead of if-else it could be default + if:
```py
x_axis_type = "log"
x_range = None
if x_min == x_max:  # constant message size
    ...
```

amaslenn (approved):

PR #173, opened 2024-08-14
Feature group all

ybenvidia:
## Summary
We're considering adding a new feature for the test scenario. Now, if we want to run our test using all the nodes in a certain group, we have to count the number of nodes available in that group and then write it to the toml file. 
This feature allows us to directly write the keyword 'max_avail': GROUP_NAME:max_avail, and cloudAI will directly select the maximum number of nodes available in the group in question.

## Test Plan
example_test.toml

```
name = "nccl-test"

[Tests.1]
    name = "nccl_test_all_reduce"
    nodes = [
        'defq:DEFQ_M25:max_avail',
        'defq:DEFQ_M27:2',
        'defq:DEFQ_M29:max_avail'
    ]
    time_limit = "00:03:00"
```

example_test_2.toml

```
name = "nccl-test"

[Tests.1]
    name = "nccl_test_all_reduce"
    nodes = [
        'defq:DEFQ_M25:1',
        'defq:DEFQ_M27:2',
        'defq:DEFQ_M29:3'
    ]
    time_limit = "00:03:00"
```


srinivas212:
Based on offline discussions - 
1. This is a good feature to add instead of counting number of nodes manually. 
2. Discussion is required on the right keyword. `all` is confusing because it would mean we need all the nodes in SU but there will be other jobs running, incl ones from the current test scenario.

ybenvidia:
There is a ruffer check error about the name of a function. However, this function already existed in the code. Should I ignore this error?

amaslenn:
> There is a ruffer check error about the name of a function. However, this function already existed in the code. Should I ignore this error?

It is not about the name, your modification increased the complexity of the function, see [here](https://docs.astral.sh/ruff/rules/complex-structure/) for more details. I suggest we extract some blocks to make it easier to read, it will help fixing this issue as well.

Please also add tests for this change.

ybenvidia:
> 2. Discussion is required on the right keyword. `all` is confusing because it would mean we need all the nodes in SU but there will be other jobs running, incl ones from the current test scenario.

As we discussed, I changed the keyword 'all' to 'max_avail'.

ybenvidia:
> > There is a ruffer check error about the name of a function. However, this function already existed in the code. Should I ignore this error?
> 
> It is not about the name, your modification increased the complexity of the function, see [here](https://docs.astral.sh/ruff/rules/complex-structure/) for more details. I suggest we extract some blocks to make it easier to read, it will help fixing this issue as well.
> 
> Please also add tests for this change.

The function has been split into small blocks.

amaslenn:
> The function has been split into small blocks.

Thank you! Please don't forget about tests.

lappazos:
@ybenvidia @amaslenn what's left to merge this one?

lappazos:
@ybenvidia please also add example of how to use this feature - ask @TaekyungHeo where to place it

amaslenn:
> @amaslenn what's left to merge this one?

We need unit tests to cover new behaviour.

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -348,7 +348,8 @@ def get_available_nodes_from_group(
         Args:
             partition_name (str): The name of the partition.
             group_name (str): The name of the group.
-            number_of_nodes (int): The number of nodes to retrieve.
+            number_of_nodes (int/str): The number of nodes to retrieve.
```
TaekyungHeo, line 351: Please replace (int/str) with Union[int, str].

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -713,7 +771,7 @@ def parse_nodes(self, nodes: List[str]) -> List[str]:
                 if len(parts) != 3:
                     raise ValueError("Format should be partition:group:num_nodes")
                 partition_name, group_name, num_nodes_str = parts
-                num_nodes = int(num_nodes_str)
+                num_nodes = int(num_nodes_str) if num_nodes_str != "max_avail" else num_nodes_str
```
TaekyungHeo, line 774: I don't think it's intuitive for num_nodes to have string values to support this feature. I believe it would be better design to expect an integer in num_nodes. Therefore, it might be preferable to add a new method to retrieve the number of idle nodes from a specified group and use it when num_nodes_str is set to `max_avail`. Please share your thoughts, @amaslenn & @srivatsankrishnan

TaekyungHeo:
Have you had a chance to test it on a Slurm system? If so, please add the test results to the test plan section of this PR.

src/cloudai/systems/slurm/slurm_system.py
ybenvidia: I understand, but if I do another method in the case num_nodes = 'max_avail', the method will be almost the same except for a few lines to differentiate the two cases.
What do you think @TaekyungHeo ?

TaekyungHeo: @ybenvidia , I have shared this with @srinivas212 last week.

ybenvidia:
> > @amaslenn what's left to merge this one?
> 
> We need unit tests to cover new behaviour.

@amaslenn Done

src/cloudai/systems/slurm/slurm_system.py
ybenvidia: @TaekyungHeo With the design you want me to do, I would call group_nodes_by_state() twice: once in parse_nodes() to get the count of available nodes and once in get_available_nodes_from_group to actually allocate the nodes.
Maybe I only have to change the word num_nodes to be more generic and not only integers ?

TaekyungHeo: @ybenvidia, I'm not asking you to change the design. I just want to discuss this item with the larger group, that's all. It seems others aren't responding, so you should bring this up in the group as Jeff did.

ybenvidia: @TaekyungHeo Sure

tests/test_slurm_system.py
```diff
@@ -155,3 +155,26 @@ def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue,
...
+            ]
+        }
+    }
+
+    available_nodes = slurm_system.get_available_nodes_from_group(partition_name, group_name, "max_avail")
+    expected_node_names = ["node01", "node02", "node03"]
+    returned_node_names = [node.name for node in available_nodes]
+
+    assert set(returned_node_names) == set(expected_node_names), "Should return all available nodes except DOWN nodes"
+    assert "node05" not in returned_node_names, "DOWN node should not be included"
```
amaslenn, line 180: This `assert` is always `True` as there is no `"node05"` in the list of nodes.

I suggest you use real names, like `slurm_system.groups["backup"][group_name][3]`. Same for asserting returned vs expected node names.

tests/test_slurm_system.py
```diff
@@ -155,3 +155,26 @@ def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue,
...
+            group_name: [
+                SlurmNode(name="node01", partition=partition_name, state=SlurmNodeState.IDLE),
+                SlurmNode(name="node02", partition=partition_name, state=SlurmNodeState.COMPLETING),
+                SlurmNode(name="node03", partition=partition_name, state=SlurmNodeState.IDLE),
+                SlurmNode(name="node04", partition=partition_name, state=SlurmNodeState.DOWN),
+            ]
+        }
+    }
+
+    available_nodes = slurm_system.get_available_nodes_from_group(partition_name, group_name, "max_avail")
```
amaslenn, line 175: We can focus on testing `allocate_nodes(...)` as it has less dependencies. It should be easier to test different options of `number_of_nodes`.

Existing test is also useful and make sense. Please extend it to test non-`"max_avail"` cases.

amaslenn (changes requested):

srivatsankrishnan (commented):
Looks good to me. Good feature to have. I think the PR has few unresolved comments wrt to tests and comments from @amaslenn. Can be merged once those are resolved.

tests/test_slurm_system.py
```diff
@@ -142,3 +143,88 @@ def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue,
...
+            slurm_system.groups[partition_name][group_name][1],
+        ],
+        SlurmNodeState.ALLOCATED: [],
+    }
+
+    return grouped_nodes
+
+
+def test_allocate_nodes_max_avail(slurm_system):
+    group_name = "backup"
```
amaslenn, line 189: Variable name is misleading, isn't it a partition name? (same for another test)

tests/test_slurm_system.py
```diff
@@ -142,3 +143,88 @@ def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue,
...
+
+    max_available_nodes = slurm_system.allocate_nodes(grouped_nodes, 2, group_name)
+    expected_node_names = [
+        slurm_system.groups["backup"][group_name][0].name,
+        slurm_system.groups["backup"][group_name][2].name,
+    ]
+    returned_node_names = [node.name for node in max_available_nodes]
+
+    assert set(returned_node_names) == set(expected_node_names), "Should return 2 available nodes"
+    assert slurm_system.groups["backup"][group_name][1] not in returned_node_names, "Only 2 nodes should be included"
```
amaslenn, line 218: Need to add `.name`, otherwise this assert is checking `object not in list[str]`.

tests/test_slurm_system.py
```diff
@@ -142,3 +143,88 @@ def test_update_node_states_with_mocked_outputs(mock_get_sinfo, mock_get_squeue,
 def test_parse_node_list(node_list: str, expected_parsed_node_list: List[str]):
     parsed_node_list = parse_node_list(node_list)
     assert parsed_node_list == expected_parsed_node_list
+
+
+def setup_mock_slurm_system(slurm_system, group_name):
```
amaslenn, line 148: `allocate_nodes` doesn't use any field from `SlurmSystem`, it can even be static. Anyway, everything you need for your test function is just `grouped_nodes` and it can be a simple fixture. Let's simplify the setup for these tests, we do not want making it comlicated.

amaslenn (changes requested):

tests/test_slurm_system.py
ybenvidia: From the function allocate_nodes : 
"""
        Allocate nodes based on the requested number or maximum availability.

        Args:
            grouped_nodes (Dict[SlurmNodeState, List[SlurmNode]]): Nodes grouped by their state.
            number_of_nodes (Union[int, str]): The number of nodes to allocate, or 'max_avail' to allocate
                all available nodes.
            **group_name (str): The name of the group.**

        Returns:
            List[SlurmNode]: A list of allocated nodes.

        Raises:
            ValueError: If the requested number of nodes exceeds the available nodes.
        """

amaslenn: Ah, I see. I was confused by `groups["backup"][group_name]`. Both partition and group names are the same.

BTW, `allocate_nodes` needs `group_name` only to raise an error, which feels a bit weird, WDYT?

ybenvidia: Exactly I can see that almost all the functions in the slurm_system.py file have this param. For `allocate_nodes`, we only need this param so that the user knows in which group we haven't managed to allocate the number of nodes he requested. I don't think we can remove that. What do you think ?

amaslenn: It is a signal that our class-internal logic is broken. Exception to be raised close to public API where we have all the inputs. If we constantly pass a variable as function arg, it should be a class member.

I don't have any good idea for reworking it right now.  And while this PR highlighted this issue, it is not a blocker here. So unless you have a good idea how to solve it, let's leave it as is.

P.S. other comments still make sense. And if we stay with `groups["backup"][group_name]` let's at lease give then better names, like `groups["partition_name"]["group_name"]` or something like that.

ybenvidia: I don't have a better idea right now to fix it...

src/cloudai/systems/slurm/slurm_system.py
ybenvidia: @amaslenn Maybe you have any idea about this comment ?

amaslenn: I agree with Taekyung on this, but the only improvement I can suggest is naming: use `num_nodes_spec: str` instead of `num_nodes_str: str`.

amaslenn (changes requested):

amaslenn (approved):
@ybenvidia thanks a lot for your contribution!

TaekyungHeo:
@ybenvidia , we need two approvals for major features. https://docs.google.com/document/d/1fqD_hBXmj0ikXX91iAVGvCLZk2tZdUKqK08TZdT7E0g/edit#heading=h.d8yv92wslwgf

ybenvidia:
@TaekyungHeo Damn ... I didn't know this PR was considered a major feature.
Sorry about that. I still don't have access to this document. I'll be careful next time.

PR #174, opened 2024-08-15
Reorg conf directory

srinivas212:
## Summary
Currently, the `conf` directory does not follow a consistent organization due to the addition on L40s configs in PR #138. This PR creates a better structure for `conf` directory.
 
## Test Plan
N/A


conf/common/system/example_slurm_cluster.toml
TaekyungHeo, line 1: We have inconsistencies with the naming. The ㄴlurm system schema is example_slurm_cluster.toml, and the standalone system schema is standalone_system.toml. Shouldn't it be slurm_system.toml instead?

TaekyungHeo (approved):

PR #176, opened 2024-08-19
Redefine Job, Runner, and System Responsibilities; Refactor Classes

TaekyungHeo:
## Summary
- Moved job lifecycle methods (is_running, is_completed, kill) from runner to job classes.
- Enhanced BaseJob with new methods for mode, system, and ID retrieval.
- Fixed a bug in parsing.

## Test Plan
**1. CI**
CI passes. New unit tests were added, but they do not cover all of the changes I made. I did not spend more time adding additional unit tests for the following reasons:
- The changes made in the Job classes are straightforward, and adding unit tests would add complexity without significantly improving coverage.
- The changes made in the System classes cannot be fully tested with unit tests.

**2. Ran on a server**
Sleep
```
$ python cloudaix.py --mode run --system-config conf/v0.6/general/system/system.toml --test-templates-dir conf/v0.6/general/test_template --tests-dir conf/v0.6/general/test --test-scenario conf/v0.6/general/test_scenario/sleep.toml 
[INFO] System configuration file: conf/v0.6/general/system/system.toml
[INFO] Test templates directory: conf/v0.6/general/test_template
[INFO] Tests directory: conf/v0.6/general/test
[INFO] Test scenario file: conf/v0.6/general/test_scenario/sleep.toml
[INFO] Output directory: None
[INFO] System Name: System
[INFO] Scheduler: slurm
[INFO] Test Scenario Name: test_scenario_example
[INFO] Test Scenario: test_scenario_example

Section Name: Tests.1
  Test Name: sleep_10
  Description: sleep_10
  No dependencies
Section Name: Tests.2
  Test Name: sleep_5
  Description: sleep_5
  Start Post Init: Tests.1, Time: 5 seconds
Section Name: Tests.3
  Test Name: sleep_5
  Description: sleep_5
  Start Post Comp: Tests.1, Time: 0 seconds
Section Name: Tests.4
  Test Name: sleep_20
  Description: sleep_20
  End Post Comp: Tests.1, Time: 5 seconds
[INFO] Initializing Runner
[INFO] Creating SlurmRunner
[INFO] Starting test scenario execution.
[INFO] Starting test: Tests.1
[INFO] Running test: Tests.1
[INFO] Executing command for test Tests.1: sbatch /obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.1/0/cloudai_sbatch_script.sh
[INFO] Starting test: Tests.4
[INFO] Running test: Tests.4
[INFO] Executing command for test Tests.4: sbatch /obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.4/0/cloudai_sbatch_script.sh
[INFO] Delayed start for test Tests.2 by 5 seconds.
[INFO] Starting test: Tests.2
[INFO] Running test: Tests.2
[INFO] Executing command for test Tests.2: sbatch /obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.2/0/cloudai_sbatch_script.sh
[INFO] Job completed: Tests.1
[INFO] Delayed start for test Tests.3 by 0 seconds.
[INFO] Starting test: Tests.3
[INFO] Running test: Tests.3
[INFO] Executing command for test Tests.3: sbatch /obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.3/0/cloudai_sbatch_script.sh
[INFO] Scheduling termination of job 286816 after 5 seconds.
[INFO] Job completed: Tests.2
[INFO] Job completed: Tests.4
[INFO] Job completed: Tests.3
[INFO] All test scenario results stored at: /obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25
[WARNING] Skipping directory '/obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.1/0' for test 'sleep_10'
[WARNING] Skipping directory '/obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.2/0' for test 'sleep_5'
[WARNING] Skipping directory '/obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.3/0' for test 'sleep_5'
[WARNING] Skipping directory '/obfuscated/path/results/test_scenario_example_2024-08-23_22-46-25/Tests.4/0' for test 'sleep_20'
```
NCCL test
```
$ python cloudaix.py --mode run --system-config conf/v0.6/general/system/system.toml --test-templates-dir conf/v0.6/general/test_template --tests-dir conf/v0.6/general/test --test-scenario conf/v0.6/general/test_scenario/nccl_test.toml
[INFO] System configuration file: conf/v0.6/general/system/system.toml
[INFO] Test templates directory: conf/v0.6/general/test_template
[INFO] Tests directory: conf/v0.6/general/test
[INFO] Test scenario file: conf/v0.6/general/test_scenario/nccl_test.toml
[INFO] Output directory: None
[INFO] System Name: System
[INFO] Scheduler: slurm
[INFO] Test Scenario Name: nccl-test
[INFO] Test Scenario: nccl-test

Section Name: Tests.1
  Test Name: nccl_test_all_reduce
  Description: all_reduce
  No dependencies
[INFO] Initializing Runner
[INFO] Creating SlurmRunner
[INFO] Starting test scenario execution.
[INFO] Starting test: Tests.1
[INFO] Running test: Tests.1
[INFO] Executing command for test Tests.1: sbatch /obfuscated/path/results/nccl-test_2024-08-23_22-50-48/Tests.1/0/cloudai_sbatch_script.sh
[INFO] Job completed: Tests.1
[INFO] All test scenario results stored at: /obfuscated/path/results/nccl-test_2024-08-23_22-50-48
```


src/cloudai/_core/base_job.py
```diff
@@ -14,40 +14,121 @@
...
         """
-        self.id = job_id
+        self.mode = mode
+        self.system = system
         self.test = test
         self.output_path = output_path
         self.terminated_by_dependency = False
 
-    def increment_iteration(self):
+    def get_mode(self) -> str:
```
amaslenn, line 53: In python we usually don't need getters and setters, because all properties are public. It looks too complicated, especially in this very simplistic form where you just return the value as it is.

In case we ever need to change it, python still has us covered, because we will just create a `@property` that will hide logic and won't require any change on a caller's side.

I suggest we remove all these `get_*` methods to keep code simply. And we can keep it also expressive, especially with converting `is_running` and `is_completed` into properties too. Caller's code will look more pythonic.

src/cloudai/_core/base_job.py
```diff
@@ -14,40 +14,121 @@
...
 
 
-class BaseJob:
+class BaseJob(ABC):
     """
     Base class for representing a job created by executing a test.
 
     Attributes
-        id (int): The unique identifier of the job.
+        mode (str): The mode of the job (e.g., 'run', 'dry-run').
```
amaslenn, line 30: Since we are refactoring it anyway, have you though about making it as two classes? DryRun will have simple implementation for `is_running` and other similar methods.

src/cloudai/_core/system.py
```diff
@@ -64,4 +63,75 @@ def __repr__(self) -> str:
...
+        """
+        error_message = (
+            "System update method is not implemented. All subclasses of the System class must implement the "
+            "'update' method to ensure the system's state can be refreshed as needed."
+        )
+        logging.error(error_message)
+        raise NotImplementedError(error_message)
+
+    @abstractmethod
+    def is_job_running(self, job: "BaseJob") -> bool:
```
amaslenn, line 80: Why do we need this function in `System` object? Can't users access `job` and ask its status?

src/cloudai/_core/base_job.py
```diff
@@ -14,40 +14,121 @@
...
+        system (System): The system in which the job is running.
```
amaslenn, line 31: I don't think `Job` should know about `System`. We already link them on `Runner` level, IMO it is a good idea to keep logic there.

Let's define which object does what and will try to stick to this idea. Something like this, but more detailed (note, that some facts below might be incorrect).
1. `System` — describes a system and knows how to interact with it.
2. `Job` — keeps track of a job and its dependencies. 
3. `Runner` — accepts a System and a Job and know how to run and keep track of it.

amaslenn (changes requested):

src/cloudai/_core/base_job.py
TaekyungHeo: > In python we usually don't need getters and setters, because all properties are public. It looks too complicated, especially in this very simplistic form where you just return the value as it is.

That makes sense, and this approach aligns with what I generally follow. I will make the necessary updates and avoid adding getters or setters.

> And we can keep it also expressive, especially with converting is_running and is_completed into properties too. 

Does this mean we should convert is_running into a property like job.status?

src/cloudai/_core/base_runner.py
```diff
@@ -136,7 +136,7 @@ async def shutdown(self):
             return
         logging.info("Terminating all jobs...")
         for job in self.jobs:
-            self.kill_job(job)
+            self.system.kill(job)
```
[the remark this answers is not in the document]
TaekyungHeo: The kill() method is implemented in the system classes because these classes understand how to terminate jobs. A Slurm system will use scancel, while a Kubernetes system will use the Kubernetes Python client. The runner classes can also terminate jobs, but it seems more intuitive to do so through the system classes, based on my understanding.

src/cloudai/_core/base_job.py
amaslenn: > Does this mean we should convert is_running into a property like job.status?

It make sense to me. So if you don't have any objections, let's go with this approach and use properties.

src/cloudai/_core/base_job.py
TaekyungHeo: Yes, you are right; the base job class does not need to know about its system class. Previously, I added this to refactor the code for unifying the job spec generation interface. It turns out that we no longer need this.

src/cloudai/_core/system.py
TaekyungHeo: Yes, you are right; the base job class does not need to know about its system class. Previously, I added this to refactor the code for unifying the job spec generation interface. It turns out that we no longer need this.

src/cloudai/_core/base_runner.py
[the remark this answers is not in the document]
amaslenn: I see, this sounds right and logical. Yet I still miss the whole picture. Runner object implies that it can start, stop, get_info about a job. Indeed, it does that through a System. As I said in another comment, let's try to describe our objects (we have DEV.md BTW) and agree on it.

src/cloudai/_core/base_job.py
TaekyungHeo: Sure, I will do it. BTW, does this apply to get_id() as well? Actually, I took this approach to maintain consistency after adding get_id(). get_id() returns a unique job ID for a given job, which is an abstract method later inherited by other job classes. Do you think we should take the property approach for this as well?

amaslenn: Make sense to me. But with `id` name it might be slightly different, because it is a keyword in Python. So maybe we rename it to `jobid` or something like that.

src/cloudai/_core/base_job.py
amaslenn: Yes, sure, let't split this functionality between the PRs. Just let align it, I think I don't have the full picture of your idea.

src/cloudai/_core/base_job.py
TaekyungHeo: Please find the updated code.

TaekyungHeo:
Rebased to the main branch.

amaslenn (approved):

srivatsankrishnan (approved):
No major issues with merging this. As discussed in the call, I understand this is blocking the K8s PR as well which needs rebasing. LGTM.

PR #185, opened 2024-08-30
Refactor JobSpecGenStrategy (CommandGenStrategy -> JobSpecGenStrategy)

TaekyungHeo:
#### 1. Background
- CloudAI generates job specifications from test templates.
- Currently, there are two types of job specifications:
  - **CommandGenStrategy**: Generates command-based job specs for Slurm systems.
  - **JsonGenStrategy**: Generates JSON-based job specs for K8s systems.

#### 2. Problem
- **Redundancy**: Both strategies serve the same purpose but for different systems (Slurm/K8s).
- **Complexity**: Maintaining separate strategies increases code duplication and complexity.

#### 3. Proposed Solution
- **Unify strategies**: Introduce `JobSpecGenStrategy` to replace both `CommandGenStrategy` and `JsonGenStrategy`. This class will generate job specifications in any required format.
  
#### 4. JobSpecGenStrategy Class
- Defines the interface for job specification generation.
- **Method**: `gen_job_spec` generates job specifications (command, JSON, etc.).
```
class JobSpecGenStrategy(TestTemplateStrategy):
    """
    Abstract base class for job specification generation.
    """
    @abstractmethod
    def gen_job_spec(
        self,
        env_vars: Dict[str, str],
        cmd_args: Dict[str, str],
        extra_env_vars: Dict[str, str],
        extra_cmd_args: str,
        output_path: Path,
        job_name: str,
        num_nodes: int,
        nodes: List[str],
    ) -> Any:
        pass

class ChakraReplaySlurmJobSpecGenStrategy(JobSpecGenStrategy):
    def gen_job_spec(
        self,
        env_vars: Dict[str, str],
        cmd_args: Dict[str, str],
        extra_env_vars: Dict[str, str],
        extra_cmd_args: str,
        output_path: Path,
        job_name: str,
        num_nodes: int,
        nodes: List[str],
    ) -> str:
        # Implementation for Slurm-based ChakraReplay job specification
        pass

```

#### 5. Benefits
- **Simplifies code**: Reduces duplication.
- **Easier maintenance**: Unified logic for different systems.

#### Note
* Related RM issue: https://redmine.mellanox.com/issues/4072473

## Test Plan
1. CI passes
2. Run on servers


TaekyungHeo:
Related RM Issue: https://redmine.mellanox.com/issues/4072473

TaekyungHeo:
@amaslenn 's comments.

It is a very good and detailed description, thank you for summarising it!
 
Yet I don't fully agree with implementation approach:
1. JobContext should not exist at all: part of it is TestRun, another part is related to Test args, output_dir is a property of a System. We already pass it here and there as multiple args, and you definitely make it better by wrapping it into an object. But let's make step further and get rid of it. In Pydantic for Test [PR](https://github.com/NVIDIA/cloudai/pull/145) I mostly stop using some of these parameters and further updates should get rid of passing them and will be accessing required properties from the relevant objects.
2. Going further, I'm not sure that cmd_gen strategies should exist the way they are now. I think it is too complicated creating a separate object for each test-system pair to generate a command line. Tests have own args in the for of str of dict[str, str] and accessible by system-specific launcher for building final executable unit. This way each Test will be any-system-runnable without introducing extra classes (cmd_gen strategies).

TaekyungHeo:
Created a new PR: https://github.com/NVIDIA/cloudai/pull/199

PR #186, opened 2024-09-03
Update NCCL Test Templates and Introduce Kubernetes-Specific Strategies

TaekyungHeo:
## Summary
* Add NCCL test templates.
* Store output log files upon job completion.

## Test Plan
1. CI passes.
2. https://github.com/Mellanox/cloudaix/pull/39#issue-2503854408

### DIFF
--- a/src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
+++ b/src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
@@ -80,8 +80,6 @@ class NcclTestKubernetesJsonGenStrategy(JsonGenStrategy):
         Returns:
             Dict[Any, Any]: A dictionary representing the Kubernetes MPIJob specification.
         """
-        docker_image_url = "ghcr.io/coreweave/nccl-tests:12.4.1-cudnn-devel-ubuntu20.04-nccl2.21.5-1-85f9143"
-
         return {
             "apiVersion": "kubeflow.org/v2beta1",
             "kind": "MPIJob",
@@ -99,7 +97,7 @@ class NcclTestKubernetesJsonGenStrategy(JsonGenStrategy):
                             "spec": {
                                 "containers": [
                                     {
-                                        "image": docker_image_url,
+                                        "image": cmd_args["docker_image_url"],
                                         "name": "nccl",
                                         "env": self._generate_env_list(env_vars),
                                         "command": self._generate_launcher_command(
@@ -119,7 +117,7 @@ class NcclTestKubernetesJsonGenStrategy(JsonGenStrategy):
                             "spec": {
                                 "containers": [
                                     {
-                                        "image": docker_image_url,
+                                        "image": cmd_args["docker_image_url"],
                                         "env": self._generate_env_list(env_vars),
                                         "name": "nccl",
                                         "resources": self._prepare_worker_resources(),
--- a/src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
+++ b/src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
@@ -1,0 +1,262 @@
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


src/cloudai/systems/kubernetes/kubernetes_system.py
```diff
@@ -591,6 +592,41 @@ def list_namespaces(self) -> List[str]:
...
+                    log_file.write(logs)
+                all_logs.append(logs)
+                logging.info(f"Logs for pod '{pod_name}' saved to '{log_file_path}'")
+            except client.ApiException as e:
+                logging.error(f"Error retrieving logs for pod '{pod_name}' in namespace '{namespace}': {e}")
+
+        if all_logs:
+            stdout_file_path = output_dir / "stdout.txt"
+            with stdout_file_path.open("w") as stdout_file:
+                stdout_file.write("\n".join(all_logs))
```
TaekyungHeo, line 627: @amaslenn 's comment: Write to stdout_file as soon as possible

amaslenn (dismissed):

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
```diff
@@ -0,0 +1,262 @@
...
+            final_num_nodes (int): The final number of nodes determined.
+            nodes (List[str]): A list of specific nodes to run the job on.
+            env_vars (Dict[str, str]): A dictionary of environment variables for the job.
+            cmd_args (Dict[str, str]): A dictionary of command-line arguments for the NCCL test.
+            extra_cmd_args (str): Additional command-line arguments for the NCCL test.
+
+        Returns:
+            Dict[Any, Any]: A dictionary representing the Kubernetes MPIJob specification.
+        """
+        docker_image_url = "ghcr.io/coreweave/nccl-tests:12.4.1-cudnn-devel-ubuntu20.04-nccl2.21.5-1-85f9143"
```
srivatsankrishnan, line 83: Why is this hardcoded here instead of reading from the cmd_args or test_template or its equivalent?

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
```diff
@@ -0,0 +1,262 @@
...
+            "warmup_iters",
+            "agg_iters",
+            "average",
+            "parallel_init",
+            "check",
+            "blocking",
+            "cudagraph",
+        ]
+
+        command_parts = [f"/opt/nccl_tests/build/{subtest_name}"]
```
srivatsankrishnan, line 208: This assumes that in the docker, this will be the build path. Is that a valid assumption? Not sure where the discussion was on owning the docker build process. But something we can configure in the test toml file?

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
```diff
@@ -0,0 +1,262 @@
...
+                command_parts.append(f"--{arg} {cmd_args[arg]}")
+
+        if extra_cmd_args:
+            command_parts.append(extra_cmd_args)
+
+        return [
+            "/bin/bash",
+            "-c",
+            (
+                f"mpirun -v --allow-run-as-root -np {final_num_nodes} "
```
srivatsankrishnan, line 220: `--allow-run-as-root` Does CloudAI or user need to be added in some specific group to have this privilege? In babyx we kind of own the cluster. But is this assumption always true?

TaekyungHeo: `allow-run-as-root` is required because you are running as root in the containers. It is not related to the host's permissions or privileges.

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
TaekyungHeo: It is okay to make assumptions about the binary paths. We already have assumptions about the path: https://github.com/NVIDIA/cloudai/blob/1e6d61845e263b9960f8c0f4da08398fa2dec1a7/src/cloudai/schema/test_template/nccl_test/slurm_command_gen_strategy.py#L88. The path differs from the Slurm schema because the CoreWeave Docker image places the binaries in `/opt/nccl_tests/build/`. This path will need to be updated if the binaries are located in a different directory.

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
TaekyungHeo: Good catch! Thank you for identifying this issue. I will update it accordingly.

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
srivatsankrishnan: Right, when we do make and make install, we usually set a flag to specify this installation path. If we own the process of building the NCCL docker ourselves, we can control this. It's a minor thing. Won't block the merge.

srivatsankrishnan (dismissed):

srivatsankrishnan (approved):
Thanks! LGTM

srinivas212 (approved):

PR #192, opened 2024-09-10
Log Runner Initialization with Run Mode Indication in BaseRunner

TaekyungHeo:
## Summary
Log Runner Initialization with Run Mode Indication in BaseRunner. Requested by Dmitry Kuzmin at https://redmine.mellanox.com/issues/3991308

## Test Plan
CI

### DIFF
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -63,6 +63,12 @@ class BaseRunner(ABC):
             test_scenario (TestScenario): The test scenario to run.
         """
         self.mode = mode
+
+        if self.mode == "dry-run":
+            logging.info("Initializing Runner [DRY-RUN mode]")
+        else:
+            logging.info("Initializing Runner [RUN mode]")
+
         self.system = system
         self.test_scenario = test_scenario
         self.output_path = self.setup_output_directory(system.output_path)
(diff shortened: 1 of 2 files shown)


src/cloudai/_core/base_runner.py
```diff
@@ -63,6 +63,12 @@ def __init__(self, mode: str, system: System, test_scenario: TestScenario):
             test_scenario (TestScenario): The test scenario to run.
         """
         self.mode = mode
+
+        if self.mode == "dry-run":
```
amaslenn, line 67: Can we do
```python
logging.info(f"Initializing Runner [{uc(mode)}] mode")
```
?

src/cloudai/_core/runner.py
```diff
@@ -43,7 +43,6 @@ class Runner:
     _runners = {}
 
     def __init__(self, mode: str, system: System, test_scenario: TestScenario):
-        logging.info("Initializing Runner")
```
amaslenn, line 46: Why moving this log line from here? IMO it make sense to keep it here as the most high-level part of runners logic.

TaekyungHeo: Okay, then I will update this line instead.

amaslenn (approved):

PR #196, opened 2024-09-17
Fix Bugs in K8s System Class Introduced During the Adoption of Pydantic

TaekyungHeo:
## Summary
- Pydantic does not support `__post_init__`. Member variables were not initialized. Replaced `__post_init__` with `__init__`.
- Added monitor_interval.
- Added docstrings for attributes to ensure completeness.
- Fixed a bug when loading Kubernetes configurations.
- Updated unit tests accordingly.

## Test Plan
1. CI passes
2. Tested on K8s
```
$ python cloudaix.py --mode run --system-config ../cloudai/conf/common/system/kubernetes_cluster.toml --test-templates-dir conf/staging/kubernetes/test_template/ --tests-dir conf/staging/kubernetes/test --test-scenario conf/staging/kubernetes/test_scenario/nccl_test.toml
...
[INFO] Logs for pod 'tests-1-launcher-f2wfl' saved to 'results/nccl-test_2024-09-17_14-13-20/Tests.1/0/tests-1-launcher-f2wfl.txt'
[INFO] Logs for pod 'tests-1-worker-0' saved to 'results/nccl-test_2024-09-17_14-13-20/Tests.1/0/tests-1-worker-0.txt'
[INFO] Logs for pod 'tests-1-worker-1' saved to 'results/nccl-test_2024-09-17_14-13-20/Tests.1/0/tests-1-worker-1.txt'
[INFO] All logs concatenated and saved to 'results/nccl-test_2024-09-17_14-13-20/Tests.1/0/stdout.txt'
[INFO] Job completed: Tests.1
[INFO] All test scenario results stored at: results/nccl-test_2024-09-17_14-13-20
```

### DIFF
--- a/tests/test_kubernetes_system.py
+++ b/tests/test_kubernetes_system.py
@@ -60,8 +60,8 @@ def k8s_system(kube_config_tempfile):
     """Fixture to create a KubernetesSystem instance with a valid kube config."""
     system_data = {
         "name": "test-system",
-        "install_path": Path("/fake/install/path"),
-        "output_path": Path("/fake/output/path"),
+        "install_path": "/fake/install/path",
+        "output_path": "/fake/output/path",
         "kube_config_path": kube_config_tempfile,
         "default_namespace": "default",
         "default_image": "test-image",


src/cloudai/systems/kubernetes/kubernetes_system.py
```diff
@@ -47,51 +56,61 @@ class KubernetesSystem(BaseModel, System):
     default_image: str
     scheduler: str = "kubernetes"
     global_env_vars: Dict[str, Any] = {}
+    monitor_interval: int = 1
     _core_v1: client.CoreV1Api
     _batch_v1: client.BatchV1Api
     _custom_objects_api: CustomObjectsApi
 
-    def __post_init__(self) -> None:
+    def __init__(self, **data):
```
amaslenn, line 64: Since we are using Pydantic v2, let's use `model_post_init` method: https://stackoverflow.com/questions/66571079/alter-field-after-instantiation-in-pydantic-basemodel-class

Let's also add K8s specific cases that would check that these properties are initialized and not re-initialized on property access.

TaekyungHeo: > Since we are using Pydantic v2, let's use model_post_init method: https://stackoverflow.com/questions/66571079/alter-field-after-instantiation-in-pydantic-basemodel-class
Updated.

> Let's also add K8s specific test cases that would check that these properties are initialized and not re-initialized on property access.
I will do.

srivatsankrishnan (approved):
LGTM

src/cloudai/systems/kubernetes/kubernetes_system.py
TaekyungHeo: > Let's also add K8s specific test cases that would check that these properties are initialized and not re-initialized on property access.

Updated, @amaslenn .

tests/test_kubernetes_system.py
```diff
@@ -0,0 +1,53 @@
...
+            default_namespace="default",
+            default_image="test-image",
+        )
+        k8s_system._core_v1 = MagicMock(client.CoreV1Api)
+        k8s_system._batch_v1 = MagicMock(client.BatchV1Api)
+        k8s_system._custom_objects_api = MagicMock(client.CustomObjectsApi)
+        yield k8s_system
+
+
+def test_initialization(k8s_system):
```
amaslenn, line 43: I'm not sure what this test actually verifies... It just mock everything, not that much code under test. I'd focus on:
1. `kube_config_path` existence.
2. `core_v1` and similar initialization.

For all that I suggest using raw dicts and `KubernetesSystem.model_validate(...)` method to have better coverage.

TaekyungHeo: I wanted to test the initialization of member variables because it failed in the main branch.

> kube_config_path existence.

We cannot test this because the configuration file does not exist in the pytest environment.

> core_v1 and similar initialization.

I intentionally did not test the initialization of these for two reasons. First, I am working on refactoring these variables, as shown in https://github.com/NVIDIA/cloudai/pull/197. Second, I am unsure whether the initialization will succeed because the configuration file is missing anyway.

TaekyungHeo: @amaslenn  Let's remove these unit tests if your initial intention was to test the API member variables. I assumed you wanted to verify whether the other member variables were initialized properly.

TaekyungHeo: Updated, @amaslenn

amaslenn: I see that use don't use `KubernetesSystem.model_validate(...)`, could you please explain why? IMO that would be a better reproducer.

TaekyungHeo: I missed it. Updated it.

tests/test_kubernetes_system.py
```diff
@@ -0,0 +1,90 @@
...
+        name="test-system",
+        install_path=Path("/fake/install/path"),
+        output_path=Path("/fake/output/path"),
+        kube_config_path=kube_config_tempfile,
+        default_namespace="default",
+        default_image="test-image",
+    )
+    k8s_system.model_post_init(None)
+
+    validated_system = KubernetesSystem.model_validate(k8s_system)
```
amaslenn, line 71: `k8s_system` should be defined as a dictionary, all types and defaults to be applied during object construction. This is what we are testing here.

amaslenn (changes requested):

tests/test_kubernetes_system.py
TaekyungHeo: I don't understand your point here. What do you mean by saying that k8s_system should be defined as a dictionary? Is it because KubernetesSystem uses Pydantic? Do you have any examples or pseudocode?

TaekyungHeo: Updated.
```
validated_system = KubernetesSystem.model_validate(k8s_system.model_dump())
```

tests/test_kubernetes_system.py
```diff
@@ -0,0 +1,90 @@
...
+        name="test-system",
+        install_path=Path("/fake/install/path"),
+        output_path=Path("/fake/output/path"),
+        kube_config_path=kube_config_tempfile,
+        default_namespace="default",
+        default_image="test-image",
+    )
+    k8s_system.model_post_init(None)
+
+    validated_system = KubernetesSystem.model_validate(k8s_system.model_dump())
```
amaslenn, line 71: This is a dump of already created object plus some hacks with `model_post_init` call. This is not what happens in regular flow. I think we should reproduce bugs as close as possible to real scenarios, that will be `KubernetesSystem.model_validate({"name": "...", ...})`.

amaslenn (changes requested):

tests/test_kubernetes_system.py
TaekyungHeo: I used model_dump because the example here suggested that it returns a dictionary: https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump. I have updated it as you suggested but used k8s_system.__dict__ instead. Please see if this works.

TaekyungHeo: I see. Please check if the updated code works. I followed your suggestions but couldn't pass the optional attributes (_core_v1, _batch_v1, and _custom_objects_api) because passing them to model_validate caused errors.

tests/test_kubernetes_system.py
```diff
@@ -0,0 +1,95 @@
...
+
+    yield kube_config_path
+
+
+@pytest.fixture
+def k8s_system(kube_config_tempfile):
+    """Fixture to create a KubernetesSystem instance with a valid kube config."""
+    system_data = {
+        "name": "test-system",
+        "install_path": Path("/fake/install/path"),
```
amaslenn, line 63: [minor] let's use `str`, not `Path`.

amaslenn (dismissed):

tests/test_kubernetes_system.py
amaslenn: That is expected, we don't set it in TOMLs either.

tests/test_kubernetes_system.py
TaekyungHeo: Updated.

amaslenn (approved):

PR #200, opened 2024-09-17
Support JaxToolbox in Kubernetes

TaekyungHeo:
## Summary
Support JaxToolbox in Kubernetes

## Test Plan
```

```


src/cloudai/schema/test_template/jax_toolbox/kubernetes_job_spec_gen_strategy.py
```diff
@@ -0,0 +1,224 @@
...
+
+            # Conditionally run main command
+            if [ "$keyword_found" = true ]; then
+                {main_command}
+            else
+                echo "Keyword not found. Skipping main command."
+            fi
+        """
+
+    def _generate_pretest_command(self, env_vars: Dict[str, str], cmd_args: Dict[str, str]) -> str:
```
srivatsankrishnan, line 170: Do we need the pre-test for initial k8 port? Assuming this will be officially part of the prologue and epilogue support? 

Also if this is going here, this needs to follow all the tricks and optimization you figured out during your NCCL tests on k8.

PR #208, opened 2024-09-26
Add Check for Installed Python Requirements in NeMoLauncherSlurmInstallStrategy

TaekyungHeo:
## Summary
Add Check for Installed Python Requirements in NeMoLauncherSlurmInstallStrategy

## Test Plan
CI passes


src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
```diff
@@ -107,6 +113,36 @@ def is_installed(self) -> InstallStatusResult:
...
+            logging.warning(f"requirements.txt not found in {repo_path}")
+            return False
+
+        with requirements_file.open() as f:
+            requirements = f.read().splitlines()
+
+        try:
+            pkg_resources.require(requirements)
+            return True
+        except pkg_resources.DistributionNotFound as e:
```
amaslenn, line 139: Could you please share examples for each issue? Please also add tests, you can verify by using Cloud AI's requirements.

amaslenn (changes requested):

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
TaekyungHeo: I realized that CloudAI falsely claims that the NeMo launcher has been installed even if the hydra-core package is missing while working on an NVIDIA-internal server. I wanted to fix the error, so I created this PR. My goal was not to check CloudAI's requirements.txt but to check NeMo's requirements.txt. How should I test this with unit tests? Is it still essential to include it in the unit tests?

amaslenn: > Is it still essential to include it in the unit tests?

Yes, any new code should be tested. It is especially true for user-facing changes like this one. We have to ensure our code is valid, otherwise user can face installation issue even though it is a problem in our code.

> How should I test this with unit tests?

I'd attempt something like this:
```py
def test_check_requirements_installed(tmp_path: Path) -> bool:
    (tmp_path / "requirements.txt").touch()
    # put some content into the file
    _check_requirements_installed(tmp_path)
```
I'd test the following conditions:
1. No `requirements.txt`.
2. `DistributionNotFound`
3. `VersionConflict`
4. Empty requirements list
5. All good.

(2) and (3) could be achieved by using data from Cloud AI's `requirements.txt`.

TaekyungHeo: Thanks, updated.

tests/test_slurm_install_strategy.py
```diff
@@ -198,6 +200,56 @@ def test_clone_repository_when_path_exists(self, strategy: NeMoLauncherSlurmInst
...
+        """Test when the requirements.txt file is missing."""
+        with patch("pathlib.Path.is_file", return_value=False):
+            assert not strategy._check_requirements_installed(Path())
+
+    def test_distribution_not_found(self, strategy: NeMoLauncherSlurmInstallStrategy):
+        """Test when a required Python package is missing."""
+        requirements_txt = "packageA==1.0\npackageB==2.0"
+        mock_dist = Mock(spec=Distribution)
+
+        with (
```
amaslenn, line 213: Mocks are hard to maintain and should be used as the last resort. They also expose too much information about code under test.

1. Why did you prefer this option over creating a `requirements.txt` file with controlled content?
2. With `pkg_resources.require` the content of the `requirements.txt` can be anything, let's keep it empty for simplicity.
3. Let's have at lease one test case with real `pkg_resources.require` call. We can take first requirement from our `requirements.txt` and check if it is satisfied. It should return `True` because we install it before running tests. This test can server as "all good" one. It will also ensure we use `pkg_resources.require` correctly.

amaslenn (changes requested):

tests/test_slurm_install_strategy.py
TaekyungHeo: > 1. Why did you prefer this option over creating a requirements.txt file with controlled content?

The requirements.txt for the NeMo launcher is generated when we run the install command. It is part of [the NeMo-Framework-Launcher repository](https://github.com/NVIDIA/NeMo-Framework-Launcher/blob/main/requirements.txt), and we do not have access to the file before the install command is executed. Therefore, creating a requirements.txt file in the unit tests is the only option to test the functionality.

amaslenn: > Therefore, creating a requirements.txt file in the unit tests is the only option to test the functionality.

I'm OK with creating `requirements.txt`, this is fine. I'm against `patch("builtins.open", mock_open(read_data=requirements_txt))`.

> There is no need to test the requirements.txt file in CloudAI

It is not about testing Cloud AI requirements, your unit tests never really call `pkg_resources.require`, it is always mocked. If we are using it with a wrong args list, tests won't catch it. I suggest we have at least one case that does such call for real and explained how to do that.

TaekyungHeo: Please check if the updated code makes sense.

tests/test_slurm_install_strategy.py
```diff
@@ -198,6 +200,53 @@ def test_clone_repository_when_path_exists(self, strategy: NeMoLauncherSlurmInst
                 text=True,
             )
 
+    def create_empty_requirements_file(self, tmp_path: Path, strategy) -> Path:
```
amaslenn, line 203: We can define it as a fixture:
```py
    @pytest.fixture
    def empty_requirements_txt(self, tmp_path: Path, strategy: NeMoLauncherSlurmInstallStrategy) -> Path:
        """Helper method to create an empty requirements.txt file."""
        requirements_dir = tmp_path / strategy.REPOSITORY_NAME
        requirements_dir.mkdir()
        requirements_file = requirements_dir / "requirements.txt"
        requirements_file.touch()
        return requirements_file
```

Then a test function will become:
```py
    def test_distribution_not_found(self, strategy: NeMoLauncherSlurmInstallStrategy, empty_requirements_txt: Path):
        mock_dist = Mock(spec=Distribution)
        with patch("pkg_resources.require", side_effect=pkg_resources.DistributionNotFound(mock_dist, None)):
            assert not strategy._check_requirements_installed(empty_requirements_txt.parent.parent)
```

To make test more useful and robust, I suggest checking the output as well, because otherwise it is hard to tell why return value is `False`:
```py
    def test_distribution_not_found(
        self,
        strategy: NeMoLauncherSlurmInstallStrategy,
        empty_requirements_txt: Path,
        capsys: pytest.CaptureFixture[str],
    ):
        """Test when a required Python package is missing."""

        mock_dist = Mock(spec=Distribution)
        with patch("pkg_resources.require", side_effect=pkg_resources.DistributionNotFound(mock_dist, None)):
            assert not strategy._check_requirements_installed(empty_requirements_txt.parent.parent)

        assert "Missing required package: " in capsys.readouterr().out
```

P.S. docstrings in the test funtions are hardly useful, they don't any value to its names.

tests/test_slurm_install_strategy.py
```diff
@@ -198,6 +200,53 @@ def test_clone_repository_when_path_exists(self, strategy: NeMoLauncherSlurmInst
...
+
+        with patch("pkg_resources.require", return_value=True):
+            assert strategy._check_requirements_installed(tmp_path)
+
+    def test_all_requirements_satisfied(self, strategy: NeMoLauncherSlurmInstallStrategy, tmp_path: Path):
+        """Test when all the required packages are installed without issues using real pkg_resources.require."""
+        requirements_dir = tmp_path / strategy.REPOSITORY_NAME
+        requirements_dir.mkdir()
+
+        requirements_txt = "setuptools>=40.0\n"
```
amaslenn, line 245: [minor] while works today and likely all our lives, someday it might break because of a completely unrelated change.

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
amaslenn, line 1: `pkg_resources` is deprecated: https://setuptools.pypa.io/en/latest/pkg_resources.html. I suggest we use a modern alternative today.

amaslenn (changes requested):

tests/test_slurm_install_strategy.py
TaekyungHeo: Please find the updated code.

src/cloudai/schema/test_template/nemo_launcher/slurm_install_strategy.py
TaekyungHeo: Please find the updated code.

tests/test_slurm_install_strategy.py
TaekyungHeo: I kept the return types as they are for two reasons. First, this PR is already printing logging messages. Second, we will have tighter integration as we move forward to NeMo 2.0. Additionally, we will implement a different installation scheme when NeMo 2.0 is released. Ideally, we should run NeMo directly in CloudAI without relying on the NeMo launcher.

tests/test_slurm_install_strategy.py
amaslenn: We can take any line from `Path.cwd() / "requirements.txt` which is Cloud AI's requirements file, it is available during test run. And since we install all the dependencies before running tests, this dependency is 100% satisfied.

tests/test_slurm_install_strategy.py
```diff
@@ -198,6 +199,48 @@ def test_clone_repository_when_path_exists(self, strategy: NeMoLauncherSlurmInst
                 text=True,
             )
 
+    def test_no_requirements_file(self, strategy: NeMoLauncherSlurmInstallStrategy):
+        with patch("pathlib.Path.is_file", return_value=False):
```
amaslenn, line 203: Let's remove this `patch`, not needed

tests/test_slurm_install_strategy.py
amaslenn: I do not follow, what types do you mean? And how NeMo 2.0 is related to this PR?

TaekyungHeo: My response was about 
> To make test more useful and robust, I suggest checking the output as well, because otherwise it is hard to tell why return value is False:

tests/test_slurm_install_strategy.py
TaekyungHeo: It makes sense.

tests/test_slurm_install_strategy.py
amaslenn: OK, I do not suggest changing this function return value, but extend tests to make sure if verifies what in suppose to verify.

Here is an example:
1. Test tries to verify that a particular exception is raised.
2. But requirements.txt doesn't exist and function returns `False`.
3. Test passes as it is the expected return value.
4. But test didn't test what it has to test.

tests/test_slurm_install_strategy.py
TaekyungHeo: Please find the updated code.

tests/test_slurm_install_strategy.py
TaekyungHeo: Please find the updated code.

tests/test_slurm_install_strategy.py
TaekyungHeo: Now I understand what you're proposing. I think the current unit tests are sufficient. I can work on these, but our goal should be to maximize the impact with the same amount of resources. I would prefer to invest our time in higher-priority items, like adding unit tests to https://github.com/NVIDIA/cloudai/pull/210/files or NeMo-K8s. What do you think? If you believe this is essential, I’ll do it.

TaekyungHeo: + It is likely that this code will become obsolete or be removed in a few months because our goal is to run NeMo directly without the NeMo launcher.

PR #210, opened 2024-09-26
Add Support for Cluster Account and gpus_per_node in Command Generation

TaekyungHeo:
## Summary
Add support for cluster account and gpus_per_node in Slurm command generation

## Test Plan
1. CI passes
2. Ran it on a real system and observed the expected error message. Did not get a chance to run it to completion because the data directory is not yet available.


srivatsankrishnan (approved):
LGTM

PR #212, opened 2024-09-27
Fix Bug in DockerImageCacheResult to Correctly Retrieve Absolute Paths

TaekyungHeo:
## Summary
This PR addresses a bug introduced by [PR #202](https://github.com/NVIDIA/cloudai/pull/202). In that PR, we aimed to retrieve the absolute path of the cached docker image because, on some servers, srun only worked with absolute paths. However, we found that remote URLs were also being incorrectly converted to absolute paths, and this PR resolves that issue. The bug was reported by Max, and I have fixed it. Additionally, I added a test case to cover docker images from remote URLs without local caching.

## Test Plan
1. CI passes
2. Tested on a server and ran an NCCL test with a remote URL.

### DIFF
--- a/src/cloudai/util/docker_image_cache_manager.py
+++ b/src/cloudai/util/docker_image_cache_manager.py
@@ -82,7 +82,7 @@ class DockerImageCacheResult:
             message (str): A message providing additional information about the result.
         """
         self.success = success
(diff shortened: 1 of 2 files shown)


srivatsankrishnan (approved):
LGTM.

PR #218, opened 2024-09-30
Remove unnecessary permission checks and exception handling in NeMoLauncherSlurmInstallStrategy

TaekyungHeo:
## Summary
Remove unnecessary permission checks and exception handling in NeMoLauncherSlurmInstallStrategy. When the path does not exist or permission is insufficient, other APIs will raise exceptions, and these exceptions will be propagated accordingly. There is no reason to check these edge cases manually if there are no additional benefits.

## Test Plan
CI passes


srivatsankrishnan (approved):
LGTM

PR #220, opened 2024-10-01
Update test definition and configurations for NeMo Launcher

TaekyungHeo:
## Summary
- **Add FSDP, optimizer, and num_nodes configuration to Trainer and Model classes**: Updated the test definition of NeMo launcher to support `gpt3/126m`.
- **Sort fields in `src/cloudai/test_definitions/nemo_launcher.py`**: Sorted the fields in the test definition for readability and maintainability

## Test Plan
1. CI passes
2. [Ran on a server](https://github.com/Mellanox/cloudaix/pull/54)

### DIFF
--- a/src/cloudai/test_definitions/nemo_launcher.py
+++ b/src/cloudai/test_definitions/nemo_launcher.py
@@ -49,6 +49,8 @@ class Trainer(BaseModel):
     max_steps: int = 400
     val_check_interval: int = 100
     log_every_n_steps: Literal["1", "2"] = "1"
+    limit_val_batches: int = 5
+    num_nodes: int = 1
     enable_checkpointing: bool = False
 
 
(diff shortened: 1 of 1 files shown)


src/cloudai/test_definitions/nemo_launcher.py
```diff
@@ -59,34 +61,52 @@ class TrainingModelData(BaseModel):
...
+    overlap_param_sync: str = "True"
+
+
 class TrainingModel(BaseModel):
     """Training model configuration."""
 
     model_config = ConfigDict(extra="forbid")
+    activations_checkpoint_num_layers: Literal["null"] = "null"
+    data: TrainingModelData = Field(default_factory=TrainingModelData)
+    fsdp: bool = True
```
srivatsankrishnan, line 81: How does this default values affect the IL-E2E teams existing Nemo Launcher runs?

srivatsankrishnan: FSDP and other parallelization optimization should ideally not be in the default. 

If we agree it to be default, then the test definitions for IL E2E test needs to be modified to explicitly set these flags as `false`. All the other Nemo test also needs to be defined that way.

Out of the box, without those changes, it will fail to run.  wdyt? 
cc: @srinivas212

TaekyungHeo: Correct, other configurations should be updated accordingly. We may need to explicitly unset these values by setting them to ~

TaekyungHeo: We need to make the following design choices:
* Are we going to adopt the idea presented in this PR?
  * If yes, should we update existing configurations to unset any arguments? Please note that we will need to repeat this process whenever we add a new argument to the test template. Additionally, the PR unsets arguments by setting their values to `~`. This requires the field to be a string. Should we adopt this approach, or do you have alternative suggestions?
  * If no, how will we provide new arguments?

PR #221, opened 2024-10-01
Use absolute paths in NeMo launcher

TaekyungHeo:
## Summary
CloudAI team found that absolute paths should be passed to NeMo launcher arguments. Otherwise, there can be an error because a Slurm system may not allow relative paths in container-mounts.

## Test Plan
1. CI passes
2. [Ran on a server](https://github.com/Mellanox/cloudaix/pull/54)


srivatsankrishnan (approved):
LGTM

PR #222, opened 2024-10-01
Rename NeMo-related TOML configurations to reflect the exact model

TaekyungHeo:
## Summary
Rename NeMo-related TOML configurations to reflect the exact model

## Test Plan
**[TODOs]**
* First, we need to make sure we are on the same page. Otherwise, we should discuss.
* Second, we need to reach out to anyone using GPT- or LLaMA-related configurations to confirm everything is functioning properly.


TaekyungHeo:
I believe we need @Bohatchuk's approval.

Bohatchuk (approved):

PR #223, opened 2024-10-01
Unset argument in NeMo launcher when its value is ~

TaekyungHeo:
## Summary
It is essential to support removing keys. Hydra allows users to remove keys with `~`, and it should be supported. Once `~` is found in the value, the key is removed. This is necessary for some test scenarios.

## Test Plan
* CI passes. Added a unit test.
* It works on a server with additional commits as shown at [here](https://github.com/Mellanox/cloudaix/pull/54)


srivatsankrishnan (approved):
LGTM

PR #228, opened 2024-10-04
Pass TestRun to gen_exec_command and gen_json

TaekyungHeo:
## Summary
Currently, CloudAI passes a very long list of arguments to `gen_exec_command` and `gen_json`. This PR proposes passing only the `TestRun` object to these methods. `TestRun` is a class that represents an instance of a test with run-specific arguments and configurations, such as the number of nodes and their list. Moreover, since `TestRun` has access to the original `Test` class, CloudAI no longer needs to pass arguments individually. Passing `TestRun` should suffice.

## Test Plan
1. CI passes
2. Ran on a server
```
$ python ./cloudaix.py --mode run --system-config ... --tests-dir conf/common/test --test-scenario conf/common/test_scenario/gpt/gpt_175B.toml 
[INFO] Scheduler: slurm
[INFO] Test Scenario Name: gpt_175B
[INFO] Test Scenario: gpt_175B

Section Name: Tests.1
  Test Name: gpt3_126m_pile
  Description: gpt3_126m_pile
  No dependencies
...
[INFO] Job completed: Tests.1
[INFO] All test scenario results stored at: results/gpt_175B_2024-10-07_07-47-02

```

### DIFF
--- a/src/cloudai/_core/command_gen_strategy.py
+++ b/src/cloudai/_core/command_gen_strategy.py
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 from abc import abstractmethod
-from pathlib import Path
-from typing import Dict, List
+
+from cloudai._core.test_scenario import TestRun
 
 from .test_template_strategy import TestTemplateStrategy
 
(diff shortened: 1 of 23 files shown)


src/cloudai/_core/command_gen_strategy.py
```diff
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 from abc import abstractmethod
-from pathlib import Path
-from typing import Dict, List
+
+from cloudai._core.test_scenario import TestRun
```
amaslenn, line 19: Should be `from .test_scenario import TestRun`.

src/cloudai/runner/kubernetes/kubernetes_runner.py
```diff
@@ -48,7 +48,7 @@ def _submit_test(self, tr: TestRun) -> KubernetesJob:
             k8s_system: KubernetesSystem = cast(KubernetesSystem, self.system)
             job_name, job_namespace = k8s_system.create_job(job_spec)
 
-        return KubernetesJob(self.mode, self.system, tr, job_namespace, job_name, job_kind, job_output_path)
```
amaslenn, line 51: Since we are passing `tr` as an argument, can we avoid passing `tr.output_path`? Would that make sense? Same for other runners.

tests/test_slurm_command_gen_strategy.py
```diff
@@ -56,6 +57,29 @@ def strategy_fixture(slurm_system: SlurmSystem) -> SlurmCommandGenStrategy:
...
+        "data_dir": "/fake/data_dir",
+    }
+    test.extra_env_vars = {"TEST_VAR_1": "value1"}
+    test.extra_cmd_args = "extra_args"
+
+    tr = TestRun(
+        test=test,
+        num_nodes=2,
+        nodes=[],
+        output_path=Path("/tmp/output"),
```
amaslenn, line 77: Let's use pytest's fixture (`tmp_path`) here to cleanup after test run automatically. Same for `"data_dir"`.

TaekyungHeo:

amaslenn (approved):

PR #244, opened 2024-10-08
Fix bug in violating Kubernetes naming rules

TaekyungHeo:
## Summary
Using the section name as the job name in Kubernetes can lead to a bug due to noncompliance with Kubernetes naming rules. Resource names must be lowercase, can only include alphanumeric characters, hyphens, and periods, and must start and end with an alphanumeric character. This pull request fixes the bug.

## Test Plan
1. CI passes
2. Ran on a server


amaslenn (dismissed):

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
```diff
@@ -27,7 +27,7 @@ def gen_json(self, tr: TestRun) -> Dict[Any, Any]:
         final_cmd_args = self._override_cmd_args(self.default_cmd_args, tr.test.cmd_args)
         final_num_nodes = self._determine_num_nodes(tr.num_nodes, tr.nodes)
         job_spec = self._create_job_spec(
-            tr.name, final_num_nodes, tr.nodes, final_env_vars, final_cmd_args, tr.test.extra_cmd_args
```
amaslenn, line 30: Let's introduce a `sanitize_name` as part of this class function and add unit tests for it.

TaekyungHeo: Please resolve this comment if you agree, @amaslenn

srivatsankrishnan (dismissed):
LGTM

src/cloudai/schema/test_template/nccl_test/kubernetes_json_gen_strategy.py
amaslenn: I disagree here. Imaging me or other developer taking part in K8s development, I won't remember this issue and will likely face it again.

While we might re-write the whole K8s System multiple times, this fundamental limitation will remain as it comes from the K8s itself. I suggest we have some unit tests for and sanitizing function to avoid this issue in future. If possible to move such guard into a generic location — it would be even better.

TaekyungHeo: Please find the updated code.

amaslenn (approved):

PR #247, opened 2024-10-08
Add unit tests for SlurmCommandGenStrategy

TaekyungHeo:
## Summary
Add unit tests for SlurmCommandGenStrategy

## Test Plan
CI passes

### DIFF
--- a/tests/slurm_command_gen_strategy/test_chakra_replay.py
+++ b/tests/slurm_command_gen_strategy/test_chakra_replay.py
@@ -22,21 +22,9 @@ from cloudai.systems import SlurmSystem
 
 
 class TestChakraReplaySlurmCommandGenStrategy:
-    def get_slurm_args(
-        self,
-        slurm_system: SlurmSystem,
-        job_name_prefix: str,
-        env_vars: Dict[str, str],
-        cmd_args: Dict[str, str],
-        num_nodes: int,
-        nodes: List[str],
-    ) -> Dict[str, Any]:
-        return ChakraReplaySlurmCommandGenStrategy(slurm_system, {})._parse_slurm_args(
-            job_name_prefix, env_vars, cmd_args, num_nodes, nodes
-        )
-
-    def get_test_command(self, slurm_system: SlurmSystem, cmd_args: Dict[str, str], extra_cmd_args: str) -> List[str]:
-        return ChakraReplaySlurmCommandGenStrategy(slurm_system, {}).generate_test_command({}, cmd_args, extra_cmd_args)
+    @pytest.fixture
+    def cmd_gen_strategy(self, slurm_system: SlurmSystem) -> ChakraReplaySlurmCommandGenStrategy:
+        return ChakraReplaySlurmCommandGenStrategy(slurm_system, {})
 
     @pytest.mark.parametrize(
         "job_name_prefix, env_vars, cmd_args, num_nodes, nodes, expected_result",
@@ -67,6 +55,7 @@ class TestChakraReplaySlurmCommandGenStrategy:
     )
     def test_parse_slurm_args(
         self,
+        cmd_gen_strategy: ChakraReplaySlurmCommandGenStrategy,
         job_name_prefix: str,
         env_vars: Dict[str, str],
         cmd_args: Dict[str, str],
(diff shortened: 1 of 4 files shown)


tests/slurm_command_gen_strategy/test_sleep.py
```diff
@@ -0,0 +1,56 @@
...
+
+from typing import Dict, List
+
+import pytest
+from cloudai.schema.test_template.sleep.slurm_command_gen_strategy import SleepSlurmCommandGenStrategy
+from cloudai.systems import SlurmSystem
+
+
+class TestSleepSlurmCommandGenStrategy:
+    def get_test_command(
```
amaslenn, line 25: I would suggest you reworking the pipeline:
1. `generate_test_command` is the function under test, it would be better to call it explicitly from the test function.
2. Command gen strategy can become a fixture.

```py
class TestSleepSlurmCommandGenStrategy:
    @pytest.fixture
    def cmd_gen_strategy(self, slurm_system: SlurmSystem) -> SleepSlurmCommandGenStrategy:
        return SleepSlurmCommandGenStrategy(slurm_system, {})

    def test_func(self, cmd_gen_strategy: SleepSlurmCommandGenStrategy):
        cmd_gen_strategy.generate_test_command(...)
```

_(this is the same for all new tests)_

tests/slurm_command_gen_strategy/test_sleep.py
```diff
@@ -0,0 +1,56 @@
...
+        self, slurm_system: SlurmSystem, env_vars: Dict[str, str], cmd_args: Dict[str, str], extra_cmd_args: str
+    ) -> List[str]:
+        return SleepSlurmCommandGenStrategy(slurm_system, {}).generate_test_command(env_vars, cmd_args, extra_cmd_args)
+
+    @pytest.mark.parametrize(
+        "env_vars, cmd_args, extra_cmd_args, expected_command",
```
amaslenn, line 31: As you are not permutating `env_vars` and `extra_cmd_args`, IMO it would be better to remove the to simplify the test.

tests/slurm_command_gen_strategy/test_chakra_replay.py
```diff
@@ -0,0 +1,132 @@
...
+        self, slurm_system: SlurmSystem, cmd_args: Dict[str, str], extra_cmd_args: str, expected_result: List[str]
+    ) -> None:
+        cmd = self.get_test_command(slurm_system, cmd_args, extra_cmd_args)
+        assert cmd == expected_result
+
+    def test_generate_test_command_invalid_args(self, slurm_system: SlurmSystem) -> None:
+        cmd_args: Dict[str, str] = {"trace_type": "comms_trace", "backend": "nccl", "device": "gpu"}
+        extra_cmd_args: str = "--max-steps 100"
+
+        with pytest.raises(KeyError):
```
amaslenn, line 131: Let's also check the exception details here. _(same for above)_

TaekyungHeo:

tests/slurm_command_gen_strategy/test_chakra_replay.py
```diff
@@ -75,19 +64,23 @@ def test_parse_slurm_args(
...
         cmd_args = {"trace_path": "/workspace/traces/"}  # Missing "docker_image_url"
         num_nodes = 2
         nodes = ["node1", "node2"]
 
-        with pytest.raises(KeyError):
-            self.get_slurm_args(slurm_system, job_name_prefix, env_vars, cmd_args, num_nodes, nodes)
+        with pytest.raises(KeyError) as exc_info:
+            cmd_gen_strategy._parse_slurm_args(job_name_prefix, env_vars, cmd_args, num_nodes, nodes)
+
+        assert str(exc_info.value) == "'docker_image_url'", "Expected missing docker_image_url key"
```
amaslenn, line 83: Such issues should be handled on TestDefinition level. Let's leave it as is for now and will remove this code later.

amaslenn (approved):

tests/slurm_command_gen_strategy/test_chakra_replay.py
TaekyungHeo: Would you like me to remove any SlurmCommandGenStrategy tests that raise exceptions, which should be handled by TestDefinition? I will create a follow-up PR if that’s the case.

PR #250, opened 2024-10-09
Remove assigning null when the value is null (NeMo launcher)

TaekyungHeo:
## Summary
This is a bug fix for https://redmine.mellanox.com/issues/4111103. Previously, we introduced https://github.com/NVIDIA/cloudai/pull/223 to allow users to remove keys from the argument list when not needed. However, it turns out that we cannot apply this approach to the data_dir argument. Therefore, this PR removes the feature, and any arguments that should be excluded by default should not be added to the test definition. Instead, they should be included in the extra_cmd_args section when necessary.

## Test Plan
1. CI passes.
2. Ran on a server

### DIFF
--- a/tests/slurm_command_gen_strategy/test_nemo_launcher.py
+++ b/tests/slurm_command_gen_strategy/test_nemo_launcher.py
@@ -16,7 +16,7 @@
 
 from pathlib import Path
 from typing import cast
-from unittest.mock import Mock, mock_open, patch
+from unittest.mock import Mock, PropertyMock, mock_open, patch
 
 import pytest
 from cloudai._core.test import Test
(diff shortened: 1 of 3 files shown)


tests/slurm_command_gen_strategy/test_nemo_launcher.py
```diff
@@ -135,21 +135,13 @@ def test_gpus_per_node_value(self, nemo_cmd_gen: NeMoLauncherSlurmCommandGenStra
...
-
     def test_data_impl_mock_skips_checks(
         self, nemo_cmd_gen: NeMoLauncherSlurmCommandGenStrategy, nemo_test_run: TestRun
     ):
-        tdef: NeMoLauncherTestDefinition = cast(NeMoLauncherTestDefinition, nemo_test_run.test.test_definition)
-        tdef.cmd_args.data_dir = "DATA_DIR"
-
-        cmd = nemo_cmd_gen.gen_exec_command(nemo_test_run)
-        assert "data_dir" in cmd
+        with patch.object(Test, "extra_cmd_args", new_callable=PropertyMock) as mock_extra_cmd_args:
```
amaslenn, line 141: Can we assign a value to `nemo_test_run.test....` directly instead of patching it? Mocks are hiding real behaviour.

TaekyungHeo: I tried to assign a value to test_definition.extra_cmd_args_str, but it failed, indicating that I was not allowed to do so. I assumed it was not allowed because of the pydantic implementation.

amaslenn: Right, but you should be able to assign values to `tdef.extra_cmd_args`, it is a simple dictionary.

TaekyungHeo: Updated, thanks.

amaslenn (approved):

PR #252, opened 2024-10-09
Refactor NeMoLauncherSlurmCommandGenStrategy unit tests

TaekyungHeo:
## Summary
Refactor NeMoLauncherSlurmCommandGenStrategy unit tests

## Test Plan
CI passes

### DIFF
--- a/tests/slurm_command_gen_strategy/test_nemo_launcher.py
+++ b/tests/slurm_command_gen_strategy/test_nemo_launcher.py
@@ -56,45 +56,45 @@ def nemo_test_run(tmp_path: Path) -> TestRun:
 
 class TestNeMoLauncherSlurmCommandGenStrategy__GenExecCommand:
     @pytest.fixture
-    def nemo_cmd_gen(self, slurm_system: SlurmSystem) -> NeMoLauncherSlurmCommandGenStrategy:
+    def cmd_gen_strategy(self, slurm_system: SlurmSystem) -> NeMoLauncherSlurmCommandGenStrategy:
         cmd_args = {"test_arg": "test_value"}
         strategy = NeMoLauncherSlurmCommandGenStrategy(slurm_system, cmd_args)
         return strategy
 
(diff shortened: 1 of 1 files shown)


tests/slurm_command_gen_strategy/test_nemo_launcher.py
```diff
@@ -15,168 +15,205 @@
...
-
-    def test_generate_exec_command(self, nemo_cmd_gen: NeMoLauncherSlurmCommandGenStrategy, nemo_test_run: TestRun):
-        cmd = nemo_cmd_gen.gen_exec_command(nemo_test_run)
-        assert "TEST_VAR_1" in cmd
-        assert "test_value" in cmd
+    def cmd_gen_strategy(self, slurm_system: SlurmSystem) -> NeMoLauncherSlurmCommandGenStrategy:
+        return NeMoLauncherSlurmCommandGenStrategy(slurm_system, {})
+
+    @pytest.mark.parametrize(
+        "expected_content",
```
amaslenn, line 58: I'm a bit concerned about this hardcoded list. If I get you idea currently, we need to ensure:
1. Test object correctly returns cmd args via TestDefinition's `cmd_args_dict`. Guess this can be taken as "it always works".
2. `expected_content` is in fact: TestDefinition's cmd_args, env_vars, extra_env_vars and extra_cmd_args. But your list is likely limited to a subset of them.
3. If we update defaults, remove a field or add a new one, this test will require an update. Or worse, won't be covering what we need.

Let's think about this step as the first step of refactoring and will continue improvements in future.

amaslenn (approved):

tests/slurm_command_gen_strategy/test_nemo_launcher.py
TaekyungHeo: Yes, I agree, and it works as a placeholder for future improvements in unit testing. I wanted to check that at least any code modifications will not break existing code functions.

PR #253, opened 2024-10-10
Refactor SlurmCommandGenStrategy (_write_sbatch_script)

TaekyungHeo:
## Summary
Remove the duplicated code in the *SlurmCommandGenStrategy classes.

## Test Plan
1. CI passes
2. Manual runs (NCCL / Grok)

### DIFF
--- a/src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
+++ b/src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
@@ -24,11 +24,7 @@ class ChakraReplaySlurmCommandGenStrategy(SlurmCommandGenStrategy):
     """Command generation strategy for ChakraReplay on Slurm systems."""
 
     def gen_exec_command(self, tr: TestRun) -> str:
-        final_env_vars = self._override_env_vars(self.system.global_env_vars, tr.test.extra_env_vars)
-        final_cmd_args = self._override_cmd_args(self.default_cmd_args, tr.test.cmd_args)
-        slurm_args = self._parse_slurm_args("chakra_replay", final_env_vars, final_cmd_args, tr.num_nodes, tr.nodes)
-        srun_command = self.generate_srun_command(slurm_args, final_env_vars, final_cmd_args, tr.test.extra_cmd_args)
-        return self._write_sbatch_script(slurm_args, final_env_vars, srun_command, tr.output_path)
+        return self._write_sbatch_script("chakra_replay", tr)
 
     def _parse_slurm_args(
         self,
--- a/src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
+++ b/src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
@@ -16,15 +16,11 @@
 
 from typing import Any, Dict, List
 
-from cloudai import TestRun
 from cloudai.systems.slurm.strategy import SlurmCommandGenStrategy
 
 
 class ChakraReplaySlurmCommandGenStrategy(SlurmCommandGenStrategy):
     """Command generation strategy for ChakraReplay on Slurm systems."""
-
-    def gen_exec_command(self, tr: TestRun) -> str:
-        return self._write_sbatch_script("chakra_replay", tr)
 
     def _parse_slurm_args(
         self,
--- a/src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
+++ b/src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
@@ -34,13 +34,8 @@ class JaxToolboxSlurmCommandGenStrategy(SlurmCommandGenStrategy):
     def gen_exec_command(self, tr: TestRun) -> str:
         self.test_name = self._extract_test_name(tr.test.cmd_args)
         self._update_env_vars(tr)
-
-        final_env_vars = self._override_env_vars(self.system.global_env_vars, tr.test.test_definition.extra_env_vars)
-        cmd_args = tr.test.test_definition.cmd_args_dict
-        cmd_args["output_path"] = str(tr.output_path)
-        slurm_args = self._parse_slurm_args("JaxToolbox", final_env_vars, cmd_args, tr.num_nodes, tr.nodes)
-        srun_command = self.generate_srun_command(slurm_args, final_env_vars, cmd_args, tr.test.extra_cmd_args)
-        return self._write_sbatch_script(slurm_args, final_env_vars, srun_command, tr.output_path)
+        tr.test.test_definition.cmd_args.output_path = str(tr.output_path)
+        return self._write_sbatch_script("jax_toolbox", tr)
 
     def _extract_test_name(self, cmd_args: Dict[str, Any]) -> str:
         """
@@ -134,7 +129,7 @@ class JaxToolboxSlurmCommandGenStrategy(SlurmCommandGenStrategy):
         ).docker_image_path
 
         local_workspace_dir = Path(cmd_args["output_path"]).resolve()
-        docker_workspace_dir = cmd_args[f"{key_prefix}.setup_flags"]["docker_workspace_dir"]
+        docker_workspace_dir = cmd_args[f"{key_prefix}.setup_flags.docker_workspace_dir"]
         container_mounts = f"{local_workspace_dir}:{docker_workspace_dir}"
 
         if "pgo_nsys_converter.profile_path" in cmd_args:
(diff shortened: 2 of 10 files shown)


src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -275,15 +270,19 @@ def _generate_python_command(
             "--num_hosts=$SLURM_NTASKS",
             "--server_addr=$SLURM_JOB_MASTER_NODE:12345",
             "--host_idx=$SLURM_PROCID",
-            f"--job_log_dir={cmd_args[f'{self.test_name}.setup_flags']['docker_workspace_dir']}",
-            f"--tfds_data_dir={cmd_args[f'{self.test_name}.setup_flags']['tfds_data_dir']}",
-            f"--enable_checkpoint_saving={cmd_args[f'{self.test_name}.setup_flags']['enable_checkpoint_saving']}",
+            f"--job_log_dir={cmd_args[f'{self.test_name}.setup_flags.docker_workspace_dir']}",
+            f"--tfds_data_dir={cmd_args[f'{self.test_name}.setup_flags.tfds_data_dir']}",
+            f"--enable_checkpoint_saving={cmd_args[f'{self.test_name}.setup_flags.enable_checkpoint_saving']}",
```
amaslenn, line 275: Why are you making this change? Is it fixing a bug?

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -275,15 +270,19 @@ def _generate_python_command(
...
             "--multiprocess_gpu",
             "--alsologtostderr",
             f'--fdl_config="{fdl_config}"',
         ]
 
-        fdl_args: Dict[str, str] = cmd_args[f"{self.test_name}.fdl"]
```
amaslenn, line 286: This change doesn't look as a refactoring either, is it fixing an issue?

tests/slurm_command_gen_strategy/test_common.py
```diff
@@ -28,24 +30,38 @@ def strategy_fixture(slurm_system: SlurmSystem) -> SlurmCommandGenStrategy:
...
 
 
-def test_filename_generation(strategy_fixture: SlurmCommandGenStrategy, tmp_path: Path):
-    args = {"job_name": "test_job", "num_nodes": 2, "partition": "test_partition", "node_list_str": "node1,node2"}
-    env_vars = {"TEST_VAR": "VALUE"}
-    srun_command = "srun --test test_arg"
-    output_path = tmp_path
+@pytest.fixture
+def testrun_fixture(tmp_path: Path) -> TestRun:
+    mock_test_definition = MagicMock(spec=TestDefinition)
```
amaslenn, line 35: Why `MagicMock` and not regular `Mock`? Are we using magic methods somehow?

src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
```diff
@@ -24,11 +24,7 @@ class ChakraReplaySlurmCommandGenStrategy(SlurmCommandGenStrategy):
     """Command generation strategy for ChakraReplay on Slurm systems."""
 
     def gen_exec_command(self, tr: TestRun) -> str:
```
amaslenn, line 26: This change effectively makes `gen_exec_command` == `_write_sbatch_script`. I questionable:
1. `gen_exec_command` is a wrapper to run a "pipeline". Subclasses can override details of this pipeline, but it still consists of several part.
2. `_write_sbatch_script` is complicated enough already to include even more logic. And it is harder to test this function one, need to mock or rely on other functions' logic.

I do like that `_write_sbatch_script` now accepts `TestRun` though, it seems logic. And maybe we could apply it all other "building blocks" of `gen_exec_command` keeping it as a wrapper. What do you think?

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
TaekyungHeo: Yes, the changes that I introduced in this PR flatten the dictionary as a side effect. Therefore, these changes are required to make it work.

src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
TaekyungHeo: > This change effectively makes gen_exec_command == _write_sbatch_script.

It is true for most of the test templates, but not all of them. NeMo launcher does not use _write_sbatch_script for gen_exec_command.

Are you proposing to merge _write_sbatch_script into gen_exec_command? This may simplify _write_sbatch_script, but it would make it harder for us to differentiate between NeMo launcher's gen_exec_command and others' gen_exec_command, which seems essential to me. Moreover, I believe the current complexity of _write_sbatch_script is acceptable."

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
TaekyungHeo: Yes, the changes that I introduced in this PR flatten the dictionary as a side effect. Therefore, these changes are required to make it work.

tests/slurm_command_gen_strategy/test_common.py
TaekyungHeo: Let me see if I can use regular Mock.

TaekyungHeo: Updated.

src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
amaslenn: No, I'm not suggesting merging these two functions. But I'm not sold on the idea of putting more logic into _write_sbatch_script either. So honestly not sure if this PR in current state makes sense.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
TaekyungHeo: I agree, and I will do that. I prefer to take smaller baby steps instead of making large changes. I just wanted to make sure we have a set of unit tests working before introducing such changes. After merging this PR, https://github.com/NVIDIA/cloudai/pull/259 will clean up the JaxToolboxSlurmCommandGenStrategy unit tests. After that, we can overhaul the SlurmCommandGenStrategy again to access tr.test_definition.

amaslenn: Unit tests is a different topic and a separate PR. Refactoring doesn't change the behaviour, it is about "form".

TaekyungHeo: Yes, it is a separate topic. My point is that I want to make the large change only after we have working unit tests to ensure that the changes do not break the existing logic.

src/cloudai/schema/test_template/chakra_replay/slurm_command_gen_strategy.py
amaslenn: I'm trying to understand why
```py
def gen_exec_cmd(...):
    return self._write_sbatch_script(...)

def _write_sbatch_script(...):
    env_vars = self._override_env_vars(...)
    cmd_args = self._override_cmd_args(...)
    slurm_args = self._parse_slurm_args(...)
    srun_command = self.generate_srun_command(...)

    # file writing logic
```
is better compared to
```py
def gen_exec_cmd(...):
    env_vars = self._override_env_vars(...)
    cmd_args = self._override_cmd_args(...)
    slurm_args = self._parse_slurm_args(...)
    srun_command = self.generate_srun_command(...)
    return self._write_sbatch_script(...)

def _write_sbatch_script(...):
    # file writing logic
```

Second option looks better from logic isolation point of view.

TaekyungHeo: > Second option looks better from logic isolation point of view.

I agree that this makes sense.

Let me clarify my intentions. First, if we take the second approach, the function calls will be duplicated across test templates (e.g., _override_env_vars, _override_cmd_args, etc.). Second, I am considering this from the perspective of the plugin implementation. Following the current implementation would require calling plugin-related functions in gen_exec_cmd, exacerbating code duplication. My goal is to allow users to override functions like _parse_slurm_args and generate_srun_command as needed in the templates. I also plan to remove or simplify _override_env_vars and _override_cmd_args.

amaslenn (approved):

PR #264, opened 2024-10-15
Handle node allocation errors gracefully, log details, and exit on failure

TaekyungHeo:
## Summary
This is a bug fix for https://redmine.mellanox.com/issues/4100971.
* Handle node allocation errors gracefully, log details, and exit on failure
* Enhance exception message in allocate_nodes

## Test Plan
CI passes.

### DIFF
--- a/src/cloudai/systems/slurm/slurm_system.py
+++ b/src/cloudai/systems/slurm/slurm_system.py
@@ -542,13 +542,20 @@ class SlurmSystem(BaseModel, System):
         Raises:
             ValueError: If the requested number of nodes exceeds the available nodes.
         """
-        # Allocate nodes based on priority: idle, then completing, then allocated
         allocated_nodes = []
+
         if isinstance(number_of_nodes, str) and number_of_nodes == "max_avail":
             allocated_nodes.extend(grouped_nodes[SlurmNodeState.IDLE])
             allocated_nodes.extend(grouped_nodes[SlurmNodeState.COMPLETING])
+
             if len(allocated_nodes) == 0:
-                raise ValueError(f"No available nodes in group '{group_name}'.")
+                raise ValueError(
+                    f"CloudAI is requesting the maximum available nodes from the group '{group_name}', "
+                    f"but no nodes are available. Please review the available nodes in the system and ensure "
+                    f"there are sufficient resources to meet the requirements of the test scenario. Additionally, "
+                    f"verify that the system is capable of hosting the maximum number of nodes specified in the test "
+                    "scenario."
+                )
 
         elif isinstance(number_of_nodes, int):
             for state in grouped_nodes:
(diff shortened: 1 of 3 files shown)


src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -466,22 +467,27 @@ def get_available_nodes_from_group(
...
 
-        return allocated_nodes
+            return allocated_nodes
+
+        except ValueError as e:
+            logging.error(
+                f"Error occurred while allocating nodes from group '{group_name}' in partition '{partition_name}': {e}",
+                exc_info=True,
+            )
+            print(f"Error: {e}")
```
amaslenn, line 489: Debug print?

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -466,22 +467,27 @@ def get_available_nodes_from_group(
             ValueError: If the partition or group is not found, or if the requested number of nodes exceeds the
                 available nodes.
         """
-        self.validate_partition_and_group(partition_name, group_name)
-        self.update_node_states()
+        try:
```
amaslenn, line 470: Can we reduce try-except block to only cover `self.allocate_nodes` call?

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -466,22 +467,27 @@ def get_available_nodes_from_group(
...
-        return allocated_nodes
+            return allocated_nodes
+
+        except ValueError as e:
+            logging.error(
+                f"Error occurred while allocating nodes from group '{group_name}' in partition '{partition_name}': {e}",
+                exc_info=True,
+            )
+            print(f"Error: {e}")
+            sys.exit(1)
```
amaslenn, line 490: I don't like we do exit in the middle of runtime. Let's consider two options:
1. Allow unhandled exception propagation. We've already logged the error, so bug is fixed.
2. Extend System with `.check()` (or something alike) to verify if system is in valid state. Or do that per-test, but explicitly.

TaekyungHeo:

amaslenn (commented):
Could you please explain how this scenario will work:
1. test1 has enough nodes
2. test2 doesn't have enough nodes
3. test3 has enough nodes
?

What will user see as the result? Which test will run?

TaekyungHeo:
@amaslenn CloudAI can run test 1, but when it tries to run test 2, it will fail and raise an exception. Test 3 will not run as CloudAI has failed.

TaekyungHeo:
@amaslenn I recall that we had a similar discussion with Jeff and Srivatsan earlier. This is an important topic, and we should make a decision, but it is outside the scope of this PR.

amaslenn (approved):

PR #265, opened 2024-10-15
Update NeMo launcher commit hash and image tag

TaekyungHeo:
## Summary
Bug fix for [this issue](https://redmine.mellanox.com/issues/4107571). The problem is due to missing LLAMA3 configuration files, as reported by Arik Gelman. The bug was introduced by [this PR](https://github.com/NVIDIA/cloudai/pull/145). We should have used the commit hash 599ecfcbbd64fd2de02f2cc093b1610d73854022, but we are currently using cf411a9ede3b466677df8ee672bcc6c396e71e1a. You can see [here](https://github.com/NVIDIA/cloudai/blob/854e06f6487af364f93e29e606c9960a38899829/conf/common/test_template/nemo_launcher.toml#L26) that we used 599ecfcbbd64fd2de02f2cc093b1610d73854022 before this PR. The [commit hash](https://github.com/NVIDIA/NeMo-Framework-Launcher/tree/599ecfcbbd64fd2de02f2cc093b1610d73854022/launcher_scripts/conf/training/llama) has the required configuration files, such as [llama3_70b.yaml](https://github.com/NVIDIA/NeMo-Framework-Launcher/blob/599ecfcbbd64fd2de02f2cc093b1610d73854022/launcher_scripts/conf/training/llama/llama3_70b.yaml).

## Test Plan
Ran on an NVIDIA server manually. Could submit a job.

### DIFF
--- a/src/cloudai/test_definitions/nemo_launcher.py
+++ b/src/cloudai/test_definitions/nemo_launcher.py
@@ -94,8 +94,8 @@ class NeMoLauncherCmdArgs(CmdArgs):
     """NeMoLauncher test command arguments."""
 
     repository_url: str = "https://github.com/NVIDIA/NeMo-Framework-Launcher.git"
-    repository_commit_hash: str = "cf411a9ede3b466677df8ee672bcc6c396e71e1a"
-    docker_image_url: str = "nvcr.io/nvidia/nemo:24.01.01"
+    repository_commit_hash: str = "599ecfcbbd64fd2de02f2cc093b1610d73854022"
+    docker_image_url: str = "nvcr.io/nvidia/nemo:24.05.01"
     stages: str = '["training"]'
     numa_mapping: NumaMapping = Field(default_factory=NumaMapping)
     cluster: Cluster = Field(default_factory=Cluster)


TaekyungHeo:
@amaslenn Considering the pros and cons of both options, I would prefer to make them mandatory arguments with default values. Consistent behavior across configurations is better for debugging. If anything fails, we cannot be certain whether it is due to a CloudAI bug or a misconfiguration. Furthermore, there is little to no advantage in removing default values when they are mandatory.

amaslenn:
Hm, this is not about "anything fails", but explicit or implicit configuration.

When user creates a configuration, he/she doesn't know which values are default and to look it up he/she needs to go into code. For many args it is OK, but docker and sha1 are different.

We already require docker url for Chakra IIRC, so in this context it is already inconsistent.

Usually explicit values are easier to debug, yet in our case debugging is likely unaffected by this, we can trace values from logs and running scripts.

amaslenn:
Let's look at is this way: when we updated default and then upgrade Cloud AI X, _all_ configs get this new value. This seems as a risky change which is hard to verify. If each config has own value set, this is not an issue.

TaekyungHeo:
> Let's look at is this way: when we updated default and then upgrade Cloud AI X, all configs get this new value. This seems as a risky change which is hard to verify. 

@amaslenn, this point makes sense. However, for this bug fix, I believe it’s better to go with the default values, as these are what users expect. Given the tight deadline, I don’t think we can introduce a different behavior (mandatory arguments without default values) for the October release. @srinivas212 and @srivatsankrishnan, please review the comments.

srivatsankrishnan:
I would assume that there has to be a default value for the commit hash. Lot of users who just want to run out of the box benchmark might not even know which ones to use. There are other users like E2E/GCA_E2E who exactly know what commits hash to use. Since we support both cases. So better to define it with a default hash.

amaslenn:
> I would assume that there has to be a default value for the commit hash. Lot of users who just want to run out of the box benchmark might not even know which ones to use. There are other users like E2E/GCA_E2E who exactly know what commits hash to use. Since we support both cases. So better to define it with a default hash.

The problem right now is that no one sets commit explicitly in the configs. And we are doomed to repeat this issue, I'm afraid.

Running "out of the box" means "clone && run", right? In this case explicit values in configs will work as well.

srinivas212 (approved):

PR #276, opened 2024-10-21
Reorganize ReportGenerationStrategy unit tests

TaekyungHeo:
## Summary
Reorganize unit tests for ReportGenerationStrategy

## Test Plan
CI passes


amaslenn (commented):
What is the benefit of longer names (for cmd gen files)?

TaekyungHeo:
@amaslenn  Initially, I didn't want to use longer names, preferring shorter ones like nccl_test.py and jax_toolbox in each directory. However, pytest complained that the file names needed to be unique, so I ended up adding the directory name to the end of each file.

amaslenn (approved):

PR #283, opened 2024-10-23
Fix bug where tests were not correctly filtered in test_no_scenario

TaekyungHeo:
## Summary
Fix bug where tests were not correctly filtered in test_no_scenario. Please confirm if it would work, @amaslenn.

## Test Plan
CI passes

### DIFF
--- a/src/cloudai/parser.py
+++ b/src/cloudai/parser.py
@@ -83,6 +83,8 @@ class Parser:
                 exit(1)  # exit right away to keep error message readable for users
             scenario_tests = set(tr.test.name for tr in test_scenario.test_runs)
             filtered_tests = [t for t in tests if t.name in scenario_tests]
+        else:
+            filtered_tests = []
 
         return system, filtered_tests, test_scenario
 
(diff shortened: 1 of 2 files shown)


amaslenn (commented):
You are changing the behaviour, if no `--test-scenario` specified, we should return _all_ available tests

`cloudai install --system-config ... --tests-dir ...` should install everything for all tests, with this change it will install nothing.

`cloudai install --system-config ... --tests-dir ... --test-scenario ...` should run installation only for tests used in the scenario, this is an optimisation to save time and space.

TaekyungHeo:
I see, thanks @amaslenn . I will close this PR.

TaekyungHeo:
@amaslenn Do we still need the following lines?
```
        fake_scenario = Mock()
        fake_scenario.tests = [Mock()]
        fake_scenario.tests[0].name = "test-1"
```

amaslenn:
> @amaslenn Do we still need the following lines?

I don't know, you can try removing it: if everything still works fine, feel free to remove.

PR #293, opened 2024-10-29
NeMo 2.0

TaekyungHeo:
## Summary
Provide a concise summary of the changes introduced by this pull request. Detail the purpose and scope of the changes, referencing any relevant issues or discussions. Explain how these changes address the problem or improve the project.

## Test Plan
In this section, describe the testing you have performed to verify the changes. Include:
- A clear description of the testing environment.
- The steps you followed to test the new features or bug fixes.
- Any specific commands used during testing, along with their outputs.
- A description of the results and observations from your testing.
This information is crucial for reviewers to understand how the changes have been validated.

## Additional Notes


src/cloudai/schema/test_template/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,57 @@
...
+
+        tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)
+        base_args.update({"image_path": tdef.docker_image.installed_path})
+
+        return base_args
+
+    def generate_test_command(self, env_vars: Dict[str, str], cmd_args: Dict[str, str], tr: TestRun) -> List[str]:
+        command = ["nemo", "llm"]
+
+        task = cmd_args.get("task", "pretrain")
```
amaslenn, line 41: As we are writing it from scratch, let's drop `cmd_args` and use `TestDefinition` object.

src/cloudai/schema/test_template/nemo_run/grading_strategy.py
amaslenn, line 1: Do we need this class in current implementation?

src/cloudai/schema/test_template/nemo_run/report_generation_strategy.py
amaslenn, line 1: Same for this one, do we need this class?

amaslenn (commented):
Please add a new case into `test_acceptance`.

src/cloudai/schema/test_template/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,63 @@
...
+
+        tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)
+        base_args.update({"image_path": tdef.docker_image.installed_path})
+
+        return base_args
+
+    def generate_test_command(self, env_vars: Dict[str, str], cmd_args: Dict[str, str], tr: TestRun) -> List[str]:
+        command = ["nemo", "llm"]
+
+        task = tr.test.cmd_args.get("task")
```
amaslenn, line 41: Sorry, I wasn't clear. What I meant is to use `tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)` and `tdef.cmd_args` object (not a dictionary).

TaekyungHeo:
**Design Discussion (Nov 18th, 2024)**
* Srivatsan - Let's see if you can support more models with this PR. Additional complexities may appear

src/cloudai/test_definitions/nemo_run.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+
+from cloudai import CmdArgs, TestDefinition
+from cloudai.installer.installables import DockerImage, Installable
+
+
+class NeMoRunCmdArgs(CmdArgs):
+    """NeMoRun test command arguments."""
+
+    docker_image_url: str = "nvcr.io/nvidia/nemo:24.09"
```
amaslenn, line 26: I suggest we change it to non-optional fields without default values. This will raise a flag on invalid configurations much earlier (in CI in our case) other then during execution. Guard checks in cmd_gen won't be needed.

No default value for docker URL would make it more explicit and reduce risk on upgrading this image: it will be done per config and only affected configs will be a subject for verification.

TaekyungHeo: Please find the updated code.

src/cloudai/schema/test_template/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,54 @@
...
+    ) -> Dict[str, Any]:
+        base_args = super()._parse_slurm_args(job_name_prefix, env_vars, cmd_args, tr)
+
+        tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)
+        base_args.update({"image_path": tdef.docker_image.installed_path})
+
+        return base_args
+
+    def generate_test_command(self, env_vars: Dict[str, str], cmd_args: Dict[str, str], tr: TestRun) -> List[str]:
+        command = ["nemo", "llm"]
```
amaslenn, line 39: [minor] could be simplified:
```py
tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)
command = ["nemo", "llm", tdef.cmd_args.task, "--factory", tdef.cmd_args.recipe_name, "-y"]
```

amaslenn (approved):

src/cloudai/schema/test_template/nemo_run/slurm_command_gen_strategy.py
TaekyungHeo: Thanks, updated.

srivatsankrishnan (approved):
Based on the conversation in the last call, the direct executor method of directly calling the srun without `sbatch` is what we are calling as Nemo2.0 support. 

https://github.com/NVIDIA/cloudai/pull/293#issuecomment-2483587148


1.  Can we ensure this works with test hooks?  (@TaekyungHeo If I recall, you mentioned this should be simpler in Nemo 2.0 integration with CloudIAI), if yes, as part of calling Nemo2.0 support complete, can we have an example configurations that is also tested with test hooks? Could be a different PR but I feel it should be there.

2.  If direct executor is going to be generic feature in Nemo 2.0, can we test it with other models to ensure this simpler interface holds true across different models. Zsolt seems to be running more complex models via Nemo 2.0. Can we keep these models in the radar and ensure this approach works for those models as well?

((If I recall both @TaekyungHeo and @amaslenn mentioned they are okay with this PR as is and any future PR should address it.)


So I will approve this PR but those ^ should be added to call Nemo 2.0 support complete IMO.

cc: @srinivas212

src/cloudai/__init__.py
```diff
@@ -75,6 +75,8 @@
     NeMoLauncherSlurmJobIdRetrievalStrategy,
 )
 from .schema.test_template.nemo_launcher.template import NeMoLauncher
+from .schema.test_template.nemo_run.slurm_command_gen_strategy import NeMoRunSlurmCommandGenStrategy
+from .schema.test_template.nemo_run.template import NeMoRun
```
amaslenn, line 79: BTW, we don't have template anymore.

TaekyungHeo:
Thanks, @srivatsankrishnan.

1. The PR #305 shows how pre-test works with NeMo-run. Please check.
3. This PR shows that the direct executor works for NeMo 2.0. However, it's hard to say that we support all models in NeMo 2.0 with this PR. Some models may need additional arguments or mount points. Still, this is a valid starting point, and we can claim that the NeMo 2.0 POC is ready. When we supported NeMo 1.0, we did not support all models in the first PR. The first PR introduced the idea while supporting a single model, and we gradually improved and refactored the code when needed. We can take the same approach here, understanding that refactoring or additional changes may be required.

PR #298, opened 2024-11-11
Generic Slurm Container job

amaslenn:
## Summary
Introduce a new workload type: GenericSlurmContainer.

1. Works with Slurm only and have limited set of arguments to set.
2. Expanded support for Installables to GitRero, previously it was supported only via PythonExecutable.

## Test Plan
CI

## Additional Notes
—


TaekyungHeo (approved):

src/cloudai/schema/test_template/slurm_container/report_generation_strategy.py
srivatsankrishnan, line 1: How is this generic  slurm container report generation? Is this very specific to the framework used for VFM models? Correct?

amaslenn: Right now it is a CSV report that simply converts this (`stdout.txt`):
```log
...
Training epoch 0, iteration 7/9 | lr: 0.0001 | global_batch_size: 256 | global_step: 7 | reduced_train_loss: 1.266 | train_step_timing in s: 112.1 | consumed_samples: 2048
Training epoch 0, iteration 8/9 | lr: 0.0001 | global_batch_size: 256 | global_step: 8 | reduced_train_loss: 0.3826 | train_step_timing in s: 112.0 | consumed_samples: 2304
Training epoch 0, iteration 9/9 | lr: 0.0001 | global_batch_size: 256 | global_step: 9 | reduced_train_loss: 0.7765 | train_step_timing in s: 112.2 | consumed_samples: 2560
```
to this (`report.csv`):
```csv
epoch,iteration,lr,global_batch_size,global_step,reduced_train_loss,train_step_timing,consumed_samples
...
0,7,0.0001,256,7,1.266,112.1,2048
0,8,0.0001,256,8,0.3826,112.0,2304
0,9,0.0001,256,9,0.7765,112.2,2560
```

It is not _very_ specific to the data, but not too generic either.

srivatsankrishnan: Right, it looks like the trainer script used by Ethan prints this. But if we use some other trainer script (even with the same framework), the stdout might look much different? For example parameters like `lr`, `reduced_train_loss` etc might not be present in the stdout.txt. Just FYI.

I think we can approve it given this is needed urgently. But do you think it's better to name it as such to reflect the framework+workload like how we define it for Nemo and Jaxtoolbox frameworks? WDYT?

amaslenn: Yes, you are right. It is tricky with this workload though: as it is generic, I have no clue what model/framework/workload will be executed and therefore I don't know how to name it properly.

Usually we have 1:1 relation between Test and Report, but in this scenario it is one Report to many Tests.

I rely on `can_handle_directory` method which will decline handling results that are not supported. I wouldn't call it ideal, but would like narrowing the scope by renaming this reporting strategy either.

TaekyungHeo (approved):

PR #321, opened 2025-01-06
Parser for DSE

srivatsankrishnan:
## Summary
As [discussed and converged](https://docs.google.com/document/d/1oNDa4G1Xz0fusxprzGoQcfi7x-RbPlaN4hocB7oHNKQ/edit?tab=t.0) before the break, here is the first PR that implements the parser for DSE. The ranges will be in the Test TOML instead of test scenario as discussed. The definitions of all the variables will be under the cmd_args.

[Update]: Based on the discussions with @amaslenn, The CmdArgs type is used throughput instead of having a seperate dict of str for DSE. The downside  is since `cmd_args` is used pretty much in all the workloads, the type definitions/function headers/signatures has to be defined correctly. Bulk of the changes you see in the files are fixing the types for `cmd_args` to make pyright/pytest happy.

The definition of the test toml when we need to specify the ranges.
```
name = "example DSE"
description = "Example DSE"
test_template_name = "ExampleEnv"

[cmd_args]
"a" = [1, 16]
"b" = [1, 2, 4, 8]
"c" = [1, 2, 4, 8, 16]
"d" = [1, 2, 4, 8]
"e" = [10, 100, 500]
"f" = [10, 100, 500]
num_layers = "4"
use_fp8 = "1"

[extra_env_vars]
ENV1 = "0"
ENV2 = "1"
ENV3 = "3221225472"
```

The definition of test TOML when we need to specify the static ranges (as how it is defined currently).

```
name = "CloudAI Test"
description = "Existing CloudAI test"
test_template_name = "ExampleTemplate"

[cmd_args]
"a" = "1"
"b" = "4"
"c" = "16"
"d" = "8"
"e" =  "100"
"f" = "500"
num_layers = "4"
use_fp8 = "1"

[extra_env_vars]
ENV1 = "0"
ENV2 = "1"
ENV3 = "3221225472"
```

The parser will now support both specification of static values as well as list. The downstream logic will determine how it wants to use the cmd_args. For instance, if we need to enumerate all the possible values in a list (if specified), the downstream logic will implement an iterator to manipulate the cmd_args accordingly. This specification is typically handled by the agent or the environment itself. But the parser responsibility is to ensure that it parses and retains the value ranges. 

## Test Plan

CI/CD.
[Update]: Removing these tests based on discussion today in design meeting. For history with previous commit hash, retaining these info. The rationale was unit test that doesn't improve coverage is not useful. I would expect the code coverage for the unit test for future PRs as a requirement to determine usefulness of any unit tests. 

Specifically we add the following tests for the parser logic modifications.

| Test Purpose | Description |
|---------------|-------------|
| cmd_args with ranges | Tests that `cmd_args` with ranges are correctly parsed and retain lists. Asserts that the `cmd_args` in the `ConcreteTestDefinition` instance match the expected values. Creates a `Test` instance and asserts that `cmd_args` are the same as `raw_cmd_args`. |
| cmd_args with static values | Tests that `cmd_args` with static values are correctly parsed. Asserts that the `cmd_args` in the `ConcreteTestDefinition` instance match the expected values. Creates a `Test` instance and asserts that `cmd_args` are the same as `raw_cmd_args`. |
| Generate commands with ranges | Tests command generation when `cmd_args` contain ranges. Asserts that the generated commands match the expected commands for all combinations. |
| Generate commands with static values | Tests command generation when `cmd_args` contain only static values. Asserts that the generated commands match the expected single command. |
| Generate commands with empty cmd_args | Tests command generation with empty `cmd_args`. Asserts that no commands are generated. |
| Invalid TOML parsing: missing fields | Tests that `TestParser` raises an error for TOML data missing required fields. Asserts that a `ValidationError` is raised when required fields are missing. |
| Invalid TOML parsing: unexpected field | Tests that `TestDefinition` raises an error for unexpected fields in TOML. Asserts that a `ValidationError` is raised for extra fields. |

### Using Existing Grok TestDefinition in CloudAI

| Test Function Name                          | Description                                                                                           |
|---------------------------------------------|-------------------------------------------------------------------------------------------------------|
| test_grok_cmd_args_with_static_values     | Tests `GrokTestDefinitionWrapper` with all static values for `cmd_args`.                               |
| test_grok_cmd_args_with_mixed_values     | Tests `GrokTestDefinitionWrapper` with a mix of static and list values for `cmd_args`.                 |
| test_grok_cmd_args_with_list_values      | Tests `GrokTestDefinitionWrapper` with all list values for `cmd_args`.                                 |
| test_grok_cmd_args_with_xla_flags_as_lists| Tests `GrokTestDefinitionWrapper` with static FDL values and list XLA flags.                           |
| test_grok_cmd_args_with_various_types     | Parametrized test for `GrokTestDefinitionWrapper` with various types (single values and lists).        |
| test_grok_cmd_args_with_incorrect_types   | Parametrized test for `GrokTestDefinitionWrapper` with incorrect types to ensure validation errors.    |

### Real System Testing

Tested on `sanity-grok-proxy-1` model on IL1.

```
...
...
...
I0108 23:23:39.034415 140503331768640 programs.py:380] [PAX STATUS]: train_step() took 0.341928 seconds.
I0108 23:23:39.034525 140503331768640 programs.py:515] steps/sec: 2.720266
I0108 23:23:39.034636 140503331768640 py_utils.py:1040] [PAX STATUS]: Elapsed time for <run>: 0.37 seconds  (@ <.../paxml/executors.py:441>)
I0108 23:23:39.034677 140503331768640 executors.py:458] [PAX STATUS]:  Starting eval_step().
I0108 23:23:39.034754 140503331768640 executors.py:421] Training loop completed (step (`20`) greater than or equal to num_train_step (`20`).
I0108 23:23:39.034785 140503331768640 executors.py:559] [PAX STATUS]: Saving checkpoint for final step.
I0108 23:23:39.034812 140503331768640 checkpoint_creators.py:229] Saving a ckpt at final step: 20
I0108 23:23:39.034897 140503331768640 checkpoint.py:407] Closing _NonBlockingCheckpointMetadataStore(enable_write=True, _write_lock=<locked _thread.RLock object owner=140503331768640 count=1 at 0x7fc56bba8600>, _store_impl=<orbax.checkpoint.metadata.checkpoint._CheckpointMetadataStoreImpl object at 0x7fc56bb7c790>, _single_thread_executor=<concurrent.futures.thread.ThreadPoolExecutor object at 0x7fc56bb7cbe0>, _write_futures=[])
I0108 23:23:39.035296 140381926053440 deleter.py:176] Delete thread exited.
I0108 23:23:39.035467 140503331768640 executors.py:569] [PAX STATUS]: Final checkpoint saved.
I0108 23:23:39.035516 140503331768640 executors.py:309] [PAX STATUS]: Shutting down executor.
I0108 23:23:39.428077 140503331768640 summary_utils.py:473] Closed SummaryWriter `/opt/paxml/workspace/summaries/eval_train`.
I0108 23:23:39.428449 140503331768640 summary_utils.py:473] Closed SummaryWriter `/opt/paxml/workspace/summaries/train`.
I0108 23:23:39.428491 140503331768640 executors.py:315] [PAX STATUS]: Executor shutdown complete.
I0108 23:23:39.428538 140503331768640 py_utils.py:1040] [PAX STATUS]: Elapsed time for <train_and_evaluate>: 70.26 seconds  (@ <.../paxml/main.py:325>)
I0108 23:23:39.428574 140503331768640 py_utils.py:340] Starting sync_global_devices All tasks finish. across 8 devices globally
I0108 23:23:39.431515 140503331768640 py_utils.py:343] Finished sync_global_devices All tasks finish. across 8 devices globally
I0108 23:23:39.431588 140503331768640 py_utils.py:1040] [PAX STATUS]: Elapsed time for <run_experiment>: 70.60 seconds  (@ <.../paxml/main.py:470>)
I0108 23:23:39.431623 140503331768640 py_utils.py:1040] [PAX STATUS]: Elapsed time for <run>: 70.77 seconds  (@ <.../paxml/main.py:560>)
I0108 23:23:39.431658 140503331768640 py_utils.py:1040] [PAX STATUS]: E2E time: Elapsed time for <_main>: 74.58 seconds  (@ <.../paxml/main.py:495>)
Generated:
    /opt/paxml/workspace/nsys_profile_profile.nsys-rep
```
## Additional Notes
Local linting is passing but failing with upstream CI/CD. The pytest failing for the golden checks for acceptance test in sbatch generation. [Update]: This is fixed.


tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    # Simulate TestDefinition validation
+    with pytest.raises(ValidationError, match="name\n  Field required"):
+        ConcreteTestDefinition(**invalid_toml_data)
+
+
+def test_invalid_toml_parsing_unexpected_field():
+    """Test that TestDefinition raises an error for unexpected fields in TOML."""
+    # Invalid data: includes an unexpected field
```
TaekyungHeo, line 252: Please remove this comment, as the function name is already descriptive.

src/cloudai/_core/test_template_strategy.py
```diff
@@ -28,50 +28,53 @@ class TestTemplateStrategy:
...
+                        args[full_key] = value.get("default", "")
 
                     # Recursively process nested dictionaries
                     nested_args = construct_args(
                         {k: v for k, v in value.items() if k not in ["type", "default", "values"]},
                         full_key,
                     )
                     args.update(nested_args)
+                elif isinstance(value, list):
+                    # Flatten ranges into comma-separated strings
```
TaekyungHeo, line 74: Let's remove this comment.

src/cloudai/_core/test_template_strategy.py
```diff
@@ -119,22 +122,24 @@ def _override_env_vars(
...
         Returns:
-            Dict[str, str]: A dictionary of command-line arguments with overrides applied.
+            Dict[str, str]: A dictionary of command-line arguments with overrides applied and ranges flattened.
         """
         final_cmd_args = default_cmd_args.copy()
         flattened_args = self._flatten_dict(provided_cmd_args)
 
         for key, value in flattened_args.items():
-            final_cmd_args[key] = value
+            # Flatten ranges into comma-separated strings if value is a list
```
TaekyungHeo, line 142: Let's remove this comment.

tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+from cloudai._core.system import System
+from cloudai._core.test import Test, TestDefinition, TestTemplate
+from pydantic import ValidationError
+
+
+class MockSystem(System):
+    """
+    Mock subclass of System for testing purposes.
+    """
+
```
TaekyungHeo, line 32: Let's remove these lines.

tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+    def kill(self, job: "BaseJob") -> None:
+        pass
+
+
+class MockTestTemplate(TestTemplate):
+    """
+    Mock subclass of TestTemplate for testing purposes.
+    """
+
```
TaekyungHeo, line 50: Let's remove these lines.

tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+    pass
+
+
+class ConcreteTestDefinition(TestDefinition):
+    """
+    Concrete subclass of TestDefinition for testing purposes.
+    """
+
```
TaekyungHeo, line 58: Let's remove these lines.

tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+    cmd_args = {}
+
+    commands = generate_commands(cmd_args)
+
+    # No commands should be generated
+    assert commands == []
+
+
+def test_invalid_toml_parsing_missing_fields():
+    """Test that TestParser raises an error for TOML data missing required fields."""
```
TaekyungHeo, line 235: I recall that we agreed not to have single-line comments in unit tests when the purpose is already clarified by the function name.

tests/test_dse_parser.py
```diff
@@ -0,0 +1,265 @@
...
+    assert commands == expected_commands
+
+
+def test_generate_commands_with_empty_cmd_args():
+    """Test command generation with empty cmd_args."""
+    cmd_args = {}
+
+    commands = generate_commands(cmd_args)
+
+    # No commands should be generated
```
TaekyungHeo, line 230: This comment can be removed

TaekyungHeo (changes requested):
Please review whether all comments are necessary. If not, remove them, leaving only the essential ones.

TaekyungHeo:
Unit test suggestion: I would like to see whether the list representation works with non-numeric values, such as ["string1", "string2"] or [true, false].

src/cloudai/_core/test_parser.py
```diff
@@ -203,14 +203,20 @@ def _parse_data(self, data: Dict[str, Any]) -> Test:
 
         return Test(test_definition=test_def, test_template=test_template)
 
-    def _parse_cmd_args(self, cmd_args_str: str) -> List[str]:
+    def _parse_cmd_args(self, cmd_args: Dict[str, Any]) -> Dict[str, Union[str, List[str]]]:
```
amaslenn, line 206: Looks like this function is not used at all, let's remote it instead of changing.

src/cloudai/_core/test_template_strategy.py
```diff
@@ -28,50 +28,53 @@ class TestTemplateStrategy:
...
-            args = {}
+        def construct_args(cmd_args: Dict[str, Union[str, List[str]]], parent_key: str = "") -> Dict[str, str]:
+            args: Dict[str, str] = {}
             for key, value in cmd_args.items():
                 full_key = f"{parent_key}.{key}" if parent_key else key
 
                 if isinstance(value, dict):
                     # If 'default' is present, add it to the arguments
                     if "default" in value:
-                        args[full_key] = value["default"]
```
amaslenn, line 65: We should never get under `if "default" in value`, have you actually see different behavior?

srivatsankrishnan:
> * Unit test suggestion: I would like to see whether the list representation works with non-numeric values, such as ["string1", "string2"] or [true, false].
> * It would be a good idea to update USER_GUIDE.md since Daria is likely to request it.

Added more unit tests for 1st point. Regarding the USER_GUIDE, yes, will be added after the End-to-End integration. Before that and doing it in this PR is premature.

src/cloudai/_core/test_template_strategy.py
srivatsankrishnan: Right, this is for the Pyright. This condition is for the Pyright without which it was failing. Will dig deeper to see if why Pyright is failing without this in the first place.

tests/test_dse_parser.py
amaslenn, line 1: I'm not sure how these tests correlate with the idea we are introducing. We agreed, that type checking will be handled by pydantic right inside existing models. That means, that instead of this:
```py
class GPTXLAFlags(XLAFlags):
    xla_gpu_all_reduce_combine_threshold_bytes: int = 447741952
    xla_gpu_enable_while_loop_double_buffering: bool = False
```
we should have this:
```py
class GPTXLAFlags(XLAFlags):
    xla_gpu_all_reduce_combine_threshold_bytes: Union[int, list[int]] = 447741952
    xla_gpu_enable_while_loop_double_buffering: Union[bool, list[bool] = False
```

Having this in mind:
1. I expect that this PR introduces DSE support for some existing models, Grok, GPT or any others. I don't see much value in creating `ConcreteTestDefinition`, especially without its main typed part: cmd_args.
2. Some cases, like checking for mandatory/unknown args, are pure pydantic validations, I'm not sure we need them, especially here.
3. I would expect some cases like this:
    ```py
    @pytest.mark.parametrize(
        "v1,v2",
        [
            (1, False),
            ([1, 2], [False, True]),
        ],
    )
    def test_(v1, v2):
        data = {
            "fdl_config": "/p",
            "xla_flags": {
                "xla_gpu_all_reduce_combine_threshold_bytes": v1,
                "xla_gpu_enable_while_loop_double_buffering": v2,
            },
        }
        GPTCmdArgs(**data)
    ```
    It ensures that both lists and singular values are supported. Plus add some negative tests with string where int or bool is expected (for both list and singular). It would anyway test pydantic functionality, but as we have a limited set of args for DSE, it make sense.

src/cloudai/_core/test.py
```diff
@@ -105,8 +105,12 @@ class TestDefinition(BaseModel, ABC):
     extra_cmd_args: dict[str, str] = {}
 
     @property
-    def cmd_args_dict(self) -> Dict[str, str]:
-        return self.cmd_args.model_dump()
+    def cmd_args_dict(self) -> Dict[str, Union[str, List[str]]]:
+        if isinstance(self.cmd_args, CmdArgs):
```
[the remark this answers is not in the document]
srivatsankrishnan: I primarily wanted the downstream integration  handle these checks  for end-to-end integration without breaking the existing flow and support. So currently the strict pydantic validation isn't happening for the list of ranges for DSE. This is what I was mentioning in the call on Monday. 


But curious now, so shouldn't this TestDefinition.cmd_args should be explicitly defined as `CmdArgs` instead of `Any`? 

```
class TestDefinition(BaseModel, ABC):
    """Base Test object."""

    __test__ = False

    name: str
    description: str
    test_template_name: str
    cmd_args: CmdArgs
    extra_cmd_args: dict[str, str] = {}
"""

File: tests/test_dse_parser.py
Comment (srivatsankrishnan) [reply tests/test_dse_parser.py:1]: Good suggestions. The PR description is agnostic to any GPT/Grok models. This just a parser that introduces how we can support a range of values in the test toml.  So the `ConcreteTestDefinition` is just for a hypothetical test toml. I think it's better to create a create a seperate test for Grok and GPT and use the existing test definitions for DSE. That way would be easier to contain changes rather than having a kitchen sink of all possible tests.

Comment (srivatsankrishnan) [reply tests/test_dse_parser.py:1]: Made these changes for existing Grok/Jaxtoolbox model definitions to support DSE (i.e., lists of ranges). Also added some more tests with above scenarios. Pretty much using existing/all variables in Grok.

File: src/cloudai/_core/test.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (srivatsankrishnan) [reply src/cloudai/_core/test.py:110]: updated it btw. Pydantic checking for cmd_args is enabled.

File: tests/dse/test_dse_parser_grok.py [hunk 12]
```diff
@@ -0,0 +1,503 @@
...
+    def cmd_args_dict(self):
+        return self._cmd_args_dict
+
+    @property
+    def installables(self):
+        return []
+
+
+def test_grok_cmd_args_with_static_values():
+    data = {
```
amaslenn, line 65: Here you are mixing JSON and pythonic definitions. Pydantic can handle that, but actual use cases will be different: either pure JSON (loaded from TOML) or pure pythonic (I don't see use cases for that right now).

Anyway, I don't see much value is such tests. One case with int and one with list value would provide the same coverage, but there will be less code. We don't need TestDefinition, we don't need Test objects, let's focus on CmdArgs or underlying objects.

The question here is if we should test _every_ fields in both DSE and non-DSE mode. I'm not sure, IMO we can skip it for now as it anyway handled by pydantic and not our code. If we start seeing issues, we can reconsider that.

srivatsankrishnan: True if I could just load the simply toml files. Maybe I will explore that and pass an actual test toml as we mature towards end-to-end integration. But it's coming from the fact that we have separate classes defined for XLAFLAGS, FDL flags. The PoC we did at least varies these both and its easier to just use the TestDefinitions and Test objects for this purpose.

The requirement you had was to enforce pydantic checks for these arguments at this level. My approach was not to have these checks at this point in the integration. We cannot keep moving the goal post. This does the job of having pydantic type checks for both static and list of ranges for each of the variables. 

Does pydantic support mixing JSON it native and pythonic definitions? Pythonic definition is artifact of our choices in the first place. It does support it.  It's a unit test and we should see the data as a place holder and there is a particular variable that I want to express as a static and a list. I think we are nit picking on this. 

Also we are changing the `FDL` and XLA fields type to support list and static value and we should expect this user for Grok models at least. So if we are changing the types in the test definition, I don't see an issue with having a test to validate that. That is the reason why the tests are broken down into its seperate XLA FLGAS, FDL flags with and without static values. If you want just two variables with int and list, the generic parser test has that for you and it does not mix styles.

src/cloudai/_core/test.py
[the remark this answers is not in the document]
srivatsankrishnan: Unless we are very very sure that it will always be a `CmdArgs` instance.  The if condition will protect any errors from `model_dump`.  Will remove it since this was fine before.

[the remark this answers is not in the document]
srivatsankrishnan: Fixed.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -193,7 +193,11 @@ def gen_post_test(self, post_test: TestScenario, base_output_path: Path) -> str:
         return "\n".join(post_test_commands)
 
     def _gen_srun_command(
-        self, slurm_args: Dict[str, Any], env_vars: Dict[str, str], cmd_args: Dict[str, str], tr: TestRun
+        self,
+        slurm_args: Dict[str, Union[str, List[str]]],
```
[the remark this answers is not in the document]
srivatsankrishnan: Thanks. It was a mistake. `Any` probably allowed It to go through.

src/cloudai/_core/test_parser.py
srivatsankrishnan: Old code. Removed.

TaekyungHeo (approved):

src/cloudai/_core/test.py
```diff
@@ -88,6 +88,7 @@ class CmdArgs(BaseModel):
     """Test command arguments."""
 
     model_config = ConfigDict(extra="forbid")
+    cmd_args: ClassVar[Dict[str, Union[str, List[str]]]] = {}
```
amaslenn, line 91: Wait, missed that before. Why is this needed?

srivatsankrishnan: Related to this comment about having `CmdArgs` class and how `Grok` and others use this type.  https://github.com/NVIDIA/cloudai/pull/321#discussion_r1905108732

Is your question about using `ClassVar` or defining the type for `cmd_args`. Since we are so particular about typing I wonder why type was not defined at all. And using `Any` as the type for `cmd_args` in `TestDefinition`.

amaslenn: No, wait, this is a new field, it didn't exist before. And therefore my question: why is it needed now?

TestDefinition.cmd_args has type Any. But each specific TestDefinition (GPT, Grok, etc.) defines exact type, which in turn is inherited from CmdArgs.

srivatsankrishnan: Yes it was based on the [comments](https://github.com/NVIDIA/cloudai/pull/321#discussion_r1905108732) you made originally about `CmdArgs` class. 

`cmd_args` type not set in its class `CmdArgs` but has way to restrict new fields. But doesn't define the type  there. 

`TestDefinition.cmd_args` has type as `Any` but shouldn't this actually be `CmdArgs`? And the typing for this wasn't originally defined. Hence this new field was added.

amaslenn: My comment was about `TestDefinition.cmd_args`, let's live it aside for now, we don't need changing it in this PR.

My current question is about `CmdArgs.cmd_args`, why is this field needed?

srivatsankrishnan: By setting TestDefinition.cmd_args to `Any` we were loosening the typing requirement. Ideally it should have been the `CmdArgs` instead of `Any`? Do you agree on this?

If you agree on this, this field fixes the typing for the `cmd_args`.

amaslenn: If we apply `CmdArgs` type to `TestDefinition.cmd_args`, it will cause errors like this: `src/cloudai/test_definitions/chakra_replay.py:36:5 - error: "cmd_args" overrides symbol of same name in class "TestDefinition"`. That is why it is set to `Any` right now, let's leave it as it is.

I still do not understand how adding a new field into a different objects fixes typing. New type is also not `CmdArgs`. I've tried to find usages of this field in new code it looks like it is not even used.

srivatsankrishnan: Something is off about this implementation the way it is now. Somewhere in the previous commits, I didn't an error with Pyright when I had defined the TestDefinition type as `CmdArgs`.  We know the type of `cmd_args` is but we are not able to define it exactly without specifying `Any`.

Anyway, removed this (without adding this my Pyright was failing before).

srivatsankrishnan: The reason of the type error you are seeing is because the `cmd_args` is changed to `<Test>CmdArgs`.

```
class ChakraReplayCmdArgs(CmdArgs):
    """ChakraReplay test command arguments."""

    docker_image_url: str
    mpi: str = "pmix"
    trace_type: str = "et"
    trace_path: Optional[str] = None
    num_replays: int = 1
    ```
   
In ChakraReplayTestDefinition(TestDefinition), the `cmd_args` type is set as `ChakraReplayCmdArgs` and the base class doesn't even define the `cmd_args`. 

if we are consistent we shouldn't see the above error.
class ChakraReplayTestDefinition(TestDefinition):
    """Test object for ChakraReplay."""

    cmd_args: CmdArgs

amaslenn (approved):

TaekyungHeo (approved):

PR #324, opened 2025-01-07
makes sleep work

vallard:
## Summary

With the original Test 2 starting after Test.1 did `start_post_init` it never ran. This change made it work, though I don't know if its right. 

Keeping the original the scenario never finished cause test 2 never seemed to start.


amaslenn:
@vallard thank you for feedback and contribution!

Could you please share which version of CloudAI you are using and exact command line you are running?

amaslenn:
Closing as there is not activity for long time. Feel free to re-open if needed.

PR #335, opened 2025-01-13
retry docker image enroot when cluster requires specifying GPU resource

lilyw97:
## Summary
Currently the docker image enroot command only contains "account", "partition", "docker image url"
But some clusters may require explicit specify the GPU resource by using `--gpu=N`, like `CW-DFW`

## Test Plan
Use Cloudaix to do the testing
- Login into CW-DFW data copy node or login node
- Setup Cloudaix env
- Clone this branch, and install cloudai manually
- Run: `python cloudaix.py install --tests-dir ./conf/staging/dgx_cloud_acceptance_test/test/ --system-config ./conf/common/system/coreweave.toml`
- Confirm if the docker image has been downloaded to ./install successfully

## Additional Notes
`--gpu=1` will only be added into enroot command when the original command got `Cannot find GPU specification` error

### DIFF
--- a/src/cloudai/util/docker_image_cache_manager.py
+++ b/src/cloudai/util/docker_image_cache_manager.py
@@ -183,7 +183,7 @@ class DockerImageCacheManager:
         return DockerImageCacheResult(False, Path(), message)
 
     def _import_docker_image(
-        self, srun_prefix: str, docker_image_url: str, docker_image_path: Path, retry: bool = False
+        self, srun_prefix: str, docker_image_url: str, docker_image_path: Path
     ) -> DockerImageCacheResult:
         enroot_import_cmd = f"{srun_prefix} enroot import -o {docker_image_path} docker://{docker_image_url}"
         logging.debug(f"Importing Docker image: {enroot_import_cmd}")
@@ -205,10 +205,6 @@ class DockerImageCacheManager:
             logging.debug(f"Command used: {enroot_import_cmd}, stdout: {p.stdout}, stderr: {p.stderr}")
             return DockerImageCacheResult(True, docker_image_path.absolute(), success_message)
         except subprocess.CalledProcessError as e:
-            # Retry enroot by adding `--gpus=1`` if cluster requires specifing GPU resource
-            if not retry and e.stderr and "Cannot find GPU specification" in e.stderr:
-                srun_prefix += " --gpus=1"
-                return self._import_docker_image(srun_prefix, docker_image_url, docker_image_path, True)
             error_message = (
                 f"Failed to import Docker image {docker_image_url}. Command: {enroot_import_cmd}. Error: {e.stderr}"
             )
@@ -251,6 +247,8 @@ class DockerImageCacheManager:
         srun_prefix = f"srun --export=ALL --partition={self.system.default_partition}"
         if self.system.account:
             srun_prefix += f" --account={self.system.account}"
+        if self.system.gpus_per_node:
+            srun_prefix += " --gres=gpu:1"
 
         return self._import_docker_image(srun_prefix, docker_image_url, docker_image_path)
 
--- a/src/cloudai/util/docker_image_cache_manager.py
+++ b/src/cloudai/util/docker_image_cache_manager.py
@@ -183,7 +183,7 @@ class DockerImageCacheManager:
         return DockerImageCacheResult(False, Path(), message)
 
     def _import_docker_image(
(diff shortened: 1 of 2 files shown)


src/cloudai/util/docker_image_cache_manager.py
```diff
@@ -182,6 +182,39 @@ def check_docker_image_exists(self, docker_image_url: str, docker_image_filename
...
+                return DockerImageCacheResult(False, Path(), error_message)
+
+            success_message = f"Docker image cached successfully at {docker_image_path}."
+            logging.debug(success_message)
+            logging.debug(f"Command used: {enroot_import_cmd}, stdout: {p.stdout}, stderr: {p.stderr}")
+            return DockerImageCacheResult(True, docker_image_path.absolute(), success_message)
+        except subprocess.CalledProcessError as e:
+            # Retry enroot by adding `--gpus=1`` if cluster requires specifing GPU resource
+            if not retry and e.stderr and "Cannot find GPU specification" in e.stderr:
+                srun_prefix += " --gpus=1"
```
amaslenn, line 210: Does it hurt if `--gpus` is always set? Will that work for IL1 or EOS? Alternatively, can we use `SlurmSystem.gpus_per_node` value to check if `--gpus` should be specified?

Main idea is to construct CMD correctly from the start and do not retry based on an error.

P.S. thanks for moving code into own function, it totally make sense.
P.P.S. please extend unit tests (`tests/test_docker_image_cache_manager.py`), you'll find many examples there.

lilyw97: I checked the slurm doc, haven't find any clues to check if it's required to add --gpus in command line, cause it's kind of being set per node, and before we actual run `srun`, we can't know which node will be assigned, please correct me if I was wrong
so I may more tend to use try catch cause it's simplest

amaslenn: CloudAI is using `SlurmSystem.gpus_per_node` for sbatch scripts generation:
```py
...
if self.system.gpus_per_node:
    batch_script_content.append(f"#SBATCH --gpus-per-node={self.system.gpus_per_node}")
    batch_script_content.append(f"#SBATCH --gres=gpu:{self.system.gpus_per_node}")
...
```

So if we are relying on it for sbatch, it make sense to use it for srun as well. Please double check on IL1.

lilyw97: I see, so `system.gpus_per_node` relies on the same parameter set in input `system.toml` file
And I think those parameters in `system.toml` file are being set manually? in cloudaix, `IL1` and `EoS` doesn't have `gpus_per_node`, and `Corewave` have it
Wondering is there a formal way to determine the parameters for different clusters when generating `system.toml`?
cc += @srivatsankrishnan

amaslenn: From my understanding, such requirements are per-cluster and there is no way to know that upfront.

TaekyungHeo: @lilyw97 ,
- Please use the gpus_per_node field in the system schema, and add the GPU option when it is presented.
- Please ensure that your PR functions correctly on IL-1, EOS, and Coreweave.

tests/test_docker_image_cache_manager.py
```diff
@@ -198,8 +200,12 @@ def test_ensure_docker_image_no_local_cache(slurm_system: SlurmSystem):
     assert result.message == ""
 
 
-def test_system_with_account(slurm_system: SlurmSystem):
-    slurm_system.account = "test_account"
+@pytest.mark.parametrize(
+    "account, gpus_per_node", [(None, None), ("test_account", None), (None, 8), ("test_account", 8)]
+)
+def test_system_with_account(slurm_system: SlurmSystem, account, gpus_per_node):
```
amaslenn, line 206: Let's rename this function to something more generic, like `test_docker_import_with_extra_system_config`

amaslenn (changes requested):
Let's double check that new code is verified on: IL1, EOS and CW clusters, can you confirm that?

lilyw97:
Hi @amaslenn , I also changed to use `--gres=gpu:1` to compatible with possible older slurm version, please help to review and trigger CI ,  thanks!

src/cloudai/util/docker_image_cache_manager.py
amaslenn: I'm good to resolve this thread. @TaekyungHeo please let us know if you have more questions.

amaslenn (approved):

PR #341, opened 2025-01-16
Manage fields' serialization of SlurmSystem

amaslenn:
## Summary
Manage fields' serialization of `SlurmSystem`. This is needed to make sure model can be dumped and then loaded back. Will be useful for external scripts that generate systems.

## Test Plan
CI.

## Additional Notes
—


TaekyungHeo (approved):

srivatsankrishnan (commented):
Where exactly and what feature request is this change needed for? Is this a bug fix or a feature request from a customer? What is the use case?

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -150,6 +150,10 @@ def groups(self) -> Dict[str, Dict[str, List[SlurmNode]]]:
 
         return groups
 
+    @field_serializer("install_path", "output_path")
+    def _path_serializer(cls, v: Path) -> str:
```
srivatsankrishnan, line 154: Where is this method currently used? Who is calling this method?

amaslenn: It is being called when `SlurmSystem.model_dump()` is called, it is part of pydantic API.

srivatsankrishnan:
> > Where exactly and what feature request is this change needed for? Is this a bug fix or a feature request from a customer? What is the use case?
> 
> I'm looking at run.py and dgx.py scripts and they are creating system TOMLs either via regular dicts or as strings. This process is error prone and can be replaced by using real python objects. This is also a useful exercise for using our python API without TOMLs and see if something is missing: and indeed, serialization didn't work for store --> retrieve flow.

Error prone for whom? The primary user of `run.py` and `dgx.py` were NVIS and others. We haven't seen any issues with this yet. More importantly let's not modify this until we get a clear user requirement from the acceptance tests folks. We are in process of getting clear requirements from them currently.


> serialization didn't work for store --> retrieve flow

What is the customer use case for this flow? Is this something we are being proactive and finding a new use case?

src/cloudai/systems/slurm/slurm_system.py
srivatsankrishnan: In the test plan, can you add the one line reproducer/command line you used? That will be helpful for me to reproduce this test on my end.

amaslenn:
@srivatsankrishnan do you think we can proceed with this?

srivatsankrishnan:
> @srivatsankrishnan do you think we can proceed with this?

Sure. regarding the use case of dumping and loading back the model, there isn't an external use case yet. Have pinged Alisson. Qq: The test case seems to have model dump. But do we need for test counterpart for model load as well?

srivatsankrishnan (dismissed):
Please answer the comments when you get a chance. But can be approved since I don't see this breaking anything.

amaslenn:
> The test case seems to have model dump. But do we need for test counterpart for model load as well?

Good point, I've modified test case to compare that recreated object produces the same dump. Comparing actual objects is tricky and not the point of this PR.

> What is the customer use case for this flow? Is this something we are being proactive and finding a new use case?

It is needed for expanding CloudAI usage, I'll share more details offline.

> Error prone for whom? The primary user of run.py and dgx.py were NVIS and others. We haven't seen any issues with this yet. More importantly let's not modify this until we get a clear user requirement from the acceptance tests folks. We are in process of getting clear requirements from them currently.

This is more philosophical topic, let's discuss it offline if you want. This PR doesn't change these files, so it can't be a blocker. As for use cases, we'll discuss it offline.

srivatsankrishnan (approved):

PR #350, opened 2025-01-25
Enforce either paralleism or mesh dimensions

srivatsankrishnan:
## Summary

For benchmarking in Jaxtoolbox, the users usually specify the `ici` and `dcn` mesh shapes. However, for DSE, the users will directly specify the parallelism mapping dimensions. For each parallelism dimension, there are many ways to construct the mesh shapes which will be handled by the agent in the DSE.

This PR enforces the requirement that for benchmarking/DSE, the user can directly specify parallelism mapping or mesh shapes directly but not both at the same time. 

## Test Plan
* CI/CD
* Pydantic validation on dry-run command

## Additional Notes

### DIFF
--- a/src/cloudai/test_definitions/jax_toolbox.py
+++ b/src/cloudai/test_definitions/jax_toolbox.py
@@ -16,7 +16,7 @@
 
 from typing import List, Optional, Union
 
-from pydantic import BaseModel, ConfigDict, field_serializer
+from pydantic import BaseModel, ConfigDict, field_serializer, root_validator
 
 from cloudai import CmdArgs, TestDefinition
 
@@ -29,6 +29,10 @@ class JaxFdl(BaseModel):
     checkpoint_policy: Union[str, list[str]] = "save_nothing"
     dcn_mesh_shape: Union[str, list[str]] = "'[1, 1, 1]'"
     fprop_dtype: Union[str, list[str]] = "bfloat16"
+    ep: Union[int, list[int]] = 1
+    tp: Union[int, list[int]] = 1
+    fsdp: Union[int, list[int]] = 1
+    dp: Union[int, list[int]] = 1
     ici_mesh_shape: Union[str, list[str]] = "'[1, 8, 1]'"
     max_steps: Union[int, list[int]] = 20
     num_gpus: Union[int, list[int]] = 64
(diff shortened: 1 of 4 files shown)


TaekyungHeo (dismissed):

src/cloudai/test_definitions/jax_toolbox.py
```diff
@@ -57,6 +61,29 @@ def _serialize_single_policy(self, value: str) -> str:
             return value.replace('"', '\\"')
         return f'\\"{value}\\"'
 
+    @root_validator(pre=True)
```
amaslenn, line 64: `@root_validator` is deprecated, please use `@model_validator(mode="after")`. `after` allows you running verification on already created model:
```py
    @model_validator(mode="after")
    def check_mesh_and_parallelism(self):
        ep = self.ep
```

Unless users specify `ep = none` in TOML, this field is always defined. The same is true for `ep`, `tp`, `dp` and `fsdp`.

I also find mutually exclusive and yet always defined options very confusing and suggest we change it, but first need to clarify its properties:
1. Is it fine NOT to use both shapes and `ep`, `tp`, `dp`, `fsdp` at all?
2. Do you think it is OK to remove default values for shapes and new args? This way it will be easier to control what is defined via TOML.

src/cloudai/test_definitions/nemotron.py
```diff
@@ -33,6 +33,8 @@ class NemotronFdl(JaxFdl):
     num_gpus: Union[int, list[int]] = 8
     percore_batch_size: Union[float, list[float]] = 0.25  # type: ignore
     use_repeated_layer: Union[bool, list[bool]] = True
+    ici_mesh_shape: Union[str, list[str]] = "[1, 8, 1]"
```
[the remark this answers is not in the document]
srivatsankrishnan: These two fields actually defined in the parent class (Jaxtoolbox). There are some unit tests with Nemo where this was needed and hence this is added.

src/cloudai/test_definitions/jax_toolbox.py
srivatsankrishnan: Yes, I realized the `@root_validator` is deprecated. I was playing with `@model_validator` but I couldn't get it working with what I want. Maybe I didn't try the `mode=after`. Let me try that and get back. 


Regarding the other comments, yes, the benchmarking folks don't exactly specify the `EP`, `DP`, `TP`, `FSDP`.  They only mention the `ici_mesh` and `dcn_mesh`. They directly infer the parallelization dimensions from the mesh shapes.

For DSE, folks typically talk about parallelization dimensions. For each parallelization dimension, we could have many different combinations of `ici_mesh` and `dcn_mesh`. I think this is just a framework quirkiness with Paxml. So I think it's reasonable to enforce this that we don't want to specify both at the same time.  We need to discuss this.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -272,6 +272,9 @@ def _generate_python_command(
                 fdl_key = cmd_arg.replace(f"{self.test_name}.fdl.", "")
                 fdl_args[fdl_key] = cmd_args[cmd_arg]
 
+        fake_flags = {"ep", "tp", "fsdp", "dp"}
```
amaslenn, line 275: We have the same code in tests. Is it to exclude new fields from the CMD? If we don't want these fields there, let's exclude them like this:
```py
    ep: Optional[Union[int, list[int]]] = Field(exclude=True, default=None)
```

What it does:
1. `obj.ep` is always accessible and works as we expect.
2. During `.model_dump()` (this is what we are doing in `.cmd_args_dict`) fields marked as "exclude" will not be exported.
3. This will allow us removing duplicating code that otherwise can lead to bugs.

PR #356, opened 2025-02-03
Support custom mounts for slurm container jobs

amaslenn:
## Summary
Support custom mounts for slurm container jobs.

## Test Plan
1. CI.
2. Manual run with custom configs.

## Additional Notes
—

### DIFF
--- a/src/cloudai/_core/test.py
+++ b/src/cloudai/_core/test.py
@@ -103,6 +103,7 @@ class TestDefinition(BaseModel, ABC):
     cmd_args: Any
     extra_env_vars: dict[str, str] = {}
     extra_cmd_args: dict[str, str] = {}
+    extra_container_mounts: list[str] = []
 
     @property
     def cmd_args_dict(self) -> Dict[str, Union[str, List[str]]]:
--- a/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
+++ b/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
@@ -65,7 +65,12 @@ class SlurmCommandGenStrategy(CommandGenStrategy):
         Function returns CommandGenStrategy specific container mounts as well as default ones
         that should always be used.
         """
-        return [f"{tr.output_path.absolute()}:/cloudai_run_results", *self._container_mounts(tr)]
+        tdef = tr.test.test_definition
+        return [
+            f"{tr.output_path.absolute()}:/cloudai_run_results",
+            *tdef.extra_container_mounts,
+            *self._container_mounts(tr),
+        ]
 
     def gen_exec_command(self, tr: TestRun) -> str:
         env_vars = self._override_env_vars(self.system.global_env_vars, tr.test.extra_env_vars)
--- a/tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
+++ b/tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
@@ -23,6 +23,7 @@ import pytest
 from cloudai import Test, TestDefinition, TestRun, TestScenario, TestTemplate
 from cloudai.systems import SlurmSystem
 from cloudai.systems.slurm.strategy import SlurmCommandGenStrategy
+from cloudai.test_definitions.nccl import NCCLCmdArgs, NCCLTestDefinition
 from tests.conftest import create_autospec_dataclass
 
 
@@ -276,3 +277,19 @@ def test_default_container_mounts(strategy_fixture: SlurmCommandGenStrategy, tes
     mounts = strategy_fixture.container_mounts(testrun_fixture)
     assert len(mounts) == 1
     assert mounts[0] == f"{testrun_fixture.output_path.absolute()}:/cloudai_run_results"
+
+
+def test_default_container_mounts_with_extra_mounts(strategy_fixture: SlurmCommandGenStrategy):
+    nccl = NCCLTestDefinition(
+        name="name",
+        description="desc",
+        test_template_name="tt",
+        cmd_args=NCCLCmdArgs(),
+        extra_container_mounts=["/host:/container"],
+    )
+    t = Test(test_definition=nccl, test_template=Mock())
+    tr = TestRun(name="t1", test=t, num_nodes=1, nodes=[], output_path=Path("./"))
+    mounts = strategy_fixture.container_mounts(tr)
+    assert len(mounts) == 2
+    assert mounts[0] == f"{tr.output_path.absolute()}:/cloudai_run_results"
+    assert mounts[1] == "/host:/container"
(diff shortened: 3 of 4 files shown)


TaekyungHeo (dismissed):

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -65,7 +65,12 @@ def container_mounts(self, tr: TestRun) -> list[str]:
         Function returns CommandGenStrategy specific container mounts as well as default ones
         that should always be used.
         """
-        return [f"{tr.output_path.absolute()}:/cloudai_run_results", *self._container_mounts(tr)]
+        tdef = tr.test.test_definition
```
srivatsankrishnan, line 68: As you are modifying these, please remove the comments as well. 

Also don't use * operator. It makes it unreadable. Is there a scenario where one of the element in list unpacks to `None`? In that case, this will not work. Please remove it and rewrite it to more readable version.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
```diff
@@ -31,6 +32,12 @@ def _container_mounts(self, tr: TestRun) -> List[str]:
         return []
 
 
+class MyTestDefinition(TestDefinition):
```
srivatsankrishnan, line 35: You already import NCCLTestDefinition anyway. Why can't you use that? Please use real objects as much as possible.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
```diff
@@ -40,18 +47,17 @@ def strategy_fixture(slurm_system: SlurmSystem) -> SlurmCommandGenStrategy:
 
 @pytest.fixture
 def testrun_fixture(tmp_path: Path) -> TestRun:
```
srivatsankrishnan, line 49: Please use real objects. You are already importing NCCLTestDefinition. You can test this via that?

amaslenn: This would make it a bit too NCCL-specific while logically it doesn't make sense. But let's try it, I agree with relying more on real objects.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
amaslenn: Make sense + see my another comment on this.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: Sorry, don't get which comments to do you mean?

`tdef.extra_container_mounts` is never `None` (same as the return value of `self._container_mounts`). Empty lists are expanded with `*` without any issues.

As for readability, I don't agree. It is widely used pythonic way for expanding lists, but if you have a recommendation how to re-write this code, please share the snippet.

src/cloudai/_core/test.py
```diff
@@ -103,6 +103,7 @@ class TestDefinition(BaseModel, ABC):
     cmd_args: Any
     extra_env_vars: dict[str, str] = {}
     extra_cmd_args: dict[str, str] = {}
+    extra_container_mounts: list[str] = []
```
srivatsankrishnan, line 106: Since we have type hinting and static type check, lets also use List from typing. We anyway import `List` but resort to built-in types. Any reason?

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
srivatsankrishnan: Sure, but fake TestDefinition doesn't server any purpose either. NCCLTestDefinition is something more accurate object we use in CloudAI.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: The doctring in function definition. We don't expect future PRs to have these. So better to clean this up from code base as we touch upon these. 

Its a list and can user enter a list with `None`? Do we have checks in place for the above use case? Using `extend` can be an alternative. nit. Not a blocker.

src/cloudai/_core/test.py
amaslenn: `typing.List` will be [deprecated](https://docs.python.org/3/library/typing.html#typing.List) at some point, `list` should be used instead. We do not restrict a particular python version for CloudAI yet, but we already use both. In future, prefer `list` over `List`.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: Not sure I follow. We still have linter that requires public methods of public classes to have documentation, plus this docstring is valid and not misleading.

If user passes `None`, it will be a type error.

You mean like this:
```py
    res = [f"{tr.output_path.absolute()}:/cloudai_run_results"]
    res.extend(tdef.extra_container_mounts)
    res.extend(self._container_mounts(tr))
    return res
```
?
I find it too verbose. Compare:
```py
    return [
        f"{tr.output_path.absolute()}:/cloudai_run_results",
        *tdef.extra_container_mounts,
        *self._container_mounts(tr),
    ]
```

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
amaslenn: Right. On the other hand, the reader of the test could ask "Why NCCL test and not the other? Will that work for other test definitions?". Let's try with NCCL, as you mentioned it is imported anyway.

src/cloudai/_core/test.py
srivatsankrishnan: Good to know. But our code base is full of using typing.* types. It seems by `oct 2025` (when it will be deprecated) we need to clean it up everywhere.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
srivatsankrishnan: We discussed and agreed to use existing NCCLTestDefinition.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: cc @srinivas212 @TaekyungHeo please share what option would you prefer here

TaekyungHeo: Either option is fine, but I prefer the second one. It seems there has been a miscommunication. Meetings can be more efficient.

srivatsankrishnan: This is not a blocker as I mentioned before. Looks like we have a majority vote on the latter. Lets close this.

srivatsankrishnan (approved):

TaekyungHeo (approved):

PR #358, opened 2025-02-05
Introduce configurable mounts for git repos

amaslenn:
## Summary
Introduce configurable mounts for git repos.

Git repos can now be specified separately from `cmd_args` and with an optional spec for mounting in a container (for ex. Slurm runs):
```toml
[[git_repos]]
url = "https://github.com/NVIDIA/cloudai"
commit = "sha1"
mount_as = "/work"

[[git_repos]]
url = "https://github.com/NVIDIA/cloudai-new"
commit = "sha1"
mount_as = "/opt/new"
```

NeMoLauncher-based definition changed to have a mandatory `lauchner_repo` field (keeping its default value). Users can override it like this:
```toml
[launcher_repo]
url = "https://github.com/NVIDIA/NeMo-Framework-Launcher.git"
commit = "597f83a4f3b04a7a60fdd7e1a51b5bfeb8d81ae8"
```

## Test Plan
1. CI
2. Testing via https://github.com/Mellanox/cloudaix/pull/161.

## Additional Notes
—


src/cloudai/_core/installables.py
```diff
@@ -70,26 +81,35 @@ def installed_path(self, value: Union[str, Path]) -> None:
         self._installed_path = value
 
 
-@dataclass
-class GitRepo(Installable):
+class GitRepo(Installable, BaseModel):
```
srivatsankrishnan, line 84: The GitRepo is basically needed for one or two frameworks right? For example DiT and maybe something like NemoLauncher (which is getting deprecated too). Does it makes sense to modify the installables so it basically affects all the test definitions? 


Could we have not just added this class within the `test.py` and instead not having to change all the imports in all the test definitions?

amaslenn: Import paths were changed because of `src/cloudai/installer/installables.py` → `src/cloudai/_core/installables.py`. Installables are needed by multiple parts of CloudAI, so it has to be in the "core".

Generic git repos are used only by DiT, but _any_ test can add own now. NemoLauncher has a dedicated field now. According to some discussions, arbitrary repos could be useful for NemoRun.

srivatsankrishnan (approved):

PR #364, opened 2025-02-11
Enable and configure Nsys tracing via test config

amaslenn:
## Summary
1. Add `NsysConfiguration` class for configuring Nsys.
2. Use Nsys through this new interface in the code.

## Test Plan
1. CI
2. TBD

## Additional Notes
—


TaekyungHeo:
Did you have a chance to get input from Zsolt to see if this design works for him?

TaekyungHeo (dismissed):

amaslenn:
> Did you have a chance to get input from Zsolt to see if this design works for him?

Yes, we've discussed that in [this thread](https://nvidia.slack.com/archives/C06LC3PRHDZ/p1737377893181459)

srivatsankrishnan (commented):
Based on the discussion we had with [Zsolt](https://docs.google.com/document/d/1Tojl86bmjDeoSqX3RfbDwpawJhiBMj83Lt5HYVqPom8/edit?tab=t.0#bookmark=id.q7rslb3hjnok), we need to check and validate on these. Primarily for NemoRun and mostly for other frameworks as well.

- Nsys trace and see if the start/end times are captured correctly.

- Checks the ranks collecting the trace

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
```diff
@@ -282,19 +286,22 @@ def _generate_python_command(
...
 
-        nsys_command = (
-            "nsys profile \\\n"
-            "    -s none \\\n"
-            f"    -o /opt/paxml/workspace/nsys_profile_{stage} \\\n"
-            "    --force-overwrite true \\\n"
-            "    --capture-range=cudaProfilerApi \\\n"
-            "    --capture-range-end=stop \\\n"
-            "    --cuda-graph-trace=node \\\n"
+        nsys = NsysConfiguration(
```
srivatsankrishnan, line 289: Not needed for Jaxtoolbox. This is getting deprecated and won't be used by anyone. You can remove this.

amaslenn: Right, the idea is to have generic interface for controlling Nsys. Even if a particular cmd_gen_strategy doesn't follow general flow (srun + nsys + test) like Jax and SlurmContainer, we can still rely on this class to build the nsys CLI.

amaslenn:
> * Nsys trace and see if the start/end times are captured correctly.
> * Checks the ranks collecting the trace

I've ran 8 nodes job, traces were collected. I don't know how to check start/stop markers, but from CloudAI point of view CLI is the same -> behavior is the same.

I believe we will have to update this code once we get to NemoRun plugin/callback support, so far it is a generic way to enable Nsys for all our workloads.

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
srivatsankrishnan: If you want to keep, please also confirm if you how you will validate if these Nsys flags can be tested for Jaxtoolbox-paxml. You plan on running these Paxml test on real system and validate the generated Nsys traces? This unfortunately need real system testing. I felt any new work on Jaxtoolbox-paxml is wasted effort at this point.

amaslenn: Nsys is always ON for Jax runs. But with this modification I didn't change our ref sbatch/run.sh content. So IMO this change doesn't require real HW testing. Which is hard to do at this point and likely just the waste of time as you've mentioned.

srivatsankrishnan (approved):

TaekyungHeo (approved):

src/cloudai/schema/test_template/jax_toolbox/slurm_command_gen_strategy.py
amaslenn: @srivatsankrishnan can we resolve this?

PR #373, opened 2025-02-24
Allow unknown fields for cmd_args in test definitions

amaslenn:
## Summary
Allow unknown fields for cmd_args in test definitions.

`verify-configs` got a new option `--strict` to let users know which fields are not known to existing schema.

## Test Plan
1. CI.
2. Manual runs of `verify-configs`.

## Additional Notes
—


TaekyungHeo (approved):

srivatsankrishnan (approved):
Lets start with this for now. Ideally what we want to check is the following. Something to check for future.
* Can we have an empty test definition file for newer models/frameworks? Is that possible?
* I assume we still need some boiler plate to define the `CmdArgs` class + add the `model_config` set to `allow` to allow passing unknown fields.

amaslenn:
> * Can we have an empty test definition file for newer models/frameworks? Is that possible?

We can create a dummy TestDefinition with almost empty CmdArgs, that will do what you describe. Cmd generation can be tricky though, but we can try it.

> * I assume we still need some boiler plate to define the `CmdArgs` class + add the `model_config` set to `allow` to allow passing unknown fields.

Right, exactly.

PR #375, opened 2025-02-25
Support extra_env_vars contains space

lilyw97:
## Summary
The new flag in NCCL test, NCCL_TEST_SPLIT, it's value will contains space, for example `MOD 8`
If we dump to script as `export NCCL_TEST_SPLIT = MOD 8` it will run into error
Need to be dumped as `export NCCL_TEST_SPLIT = "MOD 8"`

### DIFF
--- a/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
+++ b/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
@@ -359,5 +359,7 @@ class SlurmCommandGenStrategy(CommandGenStrategy):
         for key in sorted(env_vars.keys()):
             value = env_vars[key]
             formatted_value = str(value["default"]) if isinstance(value, dict) and "default" in value else str(value)
+            if " " in formatted_value:
+                formatted_value = f'"{formatted_value}"'
             formatted_vars.append(f"export {key}={formatted_value}")
         return "\n".join(formatted_vars)


src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -359,5 +359,7 @@ def _format_env_vars(self, env_vars: Dict[str, Any]) -> str:
         for key in sorted(env_vars.keys()):
             value = env_vars[key]
             formatted_value = str(value["default"]) if isinstance(value, dict) and "default" in value else str(value)
+            if " " in formatted_value:
```
amaslenn, line 362: This implementation contains a bug (see failing tests in CI). We need to check if the value is already quoted.

Please add this test into `tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py`:
```py
@pytest.mark.parametrize(
    "in_value,expected",
    [
        ("v1", "export VAR=v1"),
        ("unquoted with spaces", 'export VAR="unquoted with spaces"'),
        ("value with 'quotes'", "export VAR=\"value with 'quotes'\""),
        ('"quoted with spaces"', 'export VAR="quoted with spaces"'),
        ('"quoted_no_spaces"', 'export VAR="quoted_no_spaces"'),
    ],
)
def test_format_env_vars(in_value: str, expected: str, strategy_fixture: SlurmCommandGenStrategy):
    formatted_vars = strategy_fixture._format_env_vars({"VAR": in_value})
    assert formatted_vars == expected
```
Feel free to extend cases with more examples.

amaslenn (commented):
Have you tried setting the value as `"ENV_VAR": '"value with spaces"'` in the test TOML?

PR #381, opened 2025-02-26
Refactor Callbacks on CloudAI Nemo Run Script

srivatsankrishnan:
## Summary
Currently we have a `combined_callbacks` that combines all other plugins we have today. But there are cases where slowdown was observed in performance when we chain all of them together. For instance, garbage collection callbacks was slowing down the training however, the comms_overlap callback is needed to improve. While we can individually pass various callback, the `timing_callbacks` is needed for the trainer to print `training_step_time` in the `stdout`.

So ensuring that individual callback can be passed and timing_callbacks is enabled by default so it prints training metrics.

## Test Plan
CI
Dry-Run


TaekyungHeo (dismissed):

amaslenn (approved):
Just to understand it better:
1. All callbacks should include `timing_callback`, without it we won't get number in stdout, correct?
2. Should we prefer plural func name (`nsys_callbacks`) or singular (`garbage_collection_callback`)? First option highlights that even though it is one, it includes some other, but either is fine. Let's be consistent.

TaekyungHeo (approved):

PR #382, opened 2025-02-27
Use non-abs bin names for NCCL tests

amaslenn:
## Summary
Use non-abs bin names for NCCL tests.

## Test Plan
1. CI.
2. Run NCCL test scenario, all test passed and generated reports.

## Additional Notes
—

### DIFF
--- a/src/cloudai/workloads/nccl_test/slurm_command_gen_strategy.py
+++ b/src/cloudai/workloads/nccl_test/slurm_command_gen_strategy.py
@@ -46,7 +46,7 @@ class NcclTestSlurmCommandGenStrategy(SlurmCommandGenStrategy):
     def generate_test_command(
         self, env_vars: Dict[str, str], cmd_args: Dict[str, Union[str, List[str]]], tr: TestRun
     ) -> List[str]:
-        srun_command_parts = [f"/usr/local/bin/{cmd_args['subtest_name']}"]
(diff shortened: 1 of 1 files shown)


TaekyungHeo (approved):

srivatsankrishnan (commented):
This PR is related to this discussion here:
https://nvidia.slack.com/archives/C088QU7CE3S/p1740666239498379

``` If we use bin name instead of abs path + update PATH to prepend new tests location, it should simplify experimentation.```


This has to happen in the command generation right? Or you were thinking about defining the PATH as an environment variable and export that to container. How will we test this?

**Comment by amaslenn** [comment]
> ` If we use bin name instead of abs path + update PATH to prepend new tests location, it should simplify experimentation.`
> 
> This has to happen in the command generation right? Or you were thinking about defining the PATH as an environment variable and export that to container. How will we test this?

Existing NCCL tests are installed into `/usr/local/bin`, this path is already part of `PATH`, so we are good.

If someone want to use custom tests build, they could do (`test.toml`):
```toml
extra_container_mount = ["/my/nccl-test/build:/nccl-tests"]

[extra_env_vars]
PATH = "/nccl-tests:$PATH"
```

This will lead to using custom binaries instead of default ones.

srivatsankrishnan (approved):

PR #390, opened 2025-03-03
Ensure unique NeMo Launcher job names using timestamp to avoid conflicts

TaekyungHeo:
## Summary
When the NeMo Launcher submits multiple jobs, they have the same job name. However, on some Slurm systems, jobs with identical names remain pending even when sufficient nodes are available. To prevent this, I added a timestamp to the job name, allowing multiple jobs to run in parallel.

## Test Plan
1. CI passes
2. Ran on IL-1 and EOS

### DIFF
--- a/tests/test_acceptance.py
+++ b/tests/test_acceptance.py
@@ -17,12 +17,12 @@
 import argparse
 from functools import partial
 from pathlib import Path
-from typing import Callable, Dict, Optional, Tuple
+from typing import Callable, Dict, Optional, Tuple, Type
 from unittest.mock import Mock, patch
 
 import pytest
 
-from cloudai import Test, TestRun, TestScenario, TestTemplate
+from cloudai import CommandGenStrategy, Test, TestDefinition, TestRun, TestScenario, TestTemplate
 from cloudai.cli import handle_dry_run_and_run, setup_logging
 from cloudai.systems import SlurmSystem
 from cloudai.systems.slurm.strategy import SlurmCommandGenStrategy
@@ -115,8 +115,8 @@ def create_test_run(
     partial_tr: partial[TestRun],
     slurm_system: SlurmSystem,
     name: str,
-    test_definition,
-    command_gen_strategy,
+    test_definition: TestDefinition,
+    command_gen_strategy: Type[CommandGenStrategy],
 ) -> TestRun:
     tr = partial_tr(
         name=name,
--- a/tests/test_acceptance.py
+++ b/tests/test_acceptance.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import argparse
+import re
 from functools import partial
 from pathlib import Path
 from typing import Dict, Optional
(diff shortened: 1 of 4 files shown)


tests/test_acceptance.py
```diff
@@ -275,7 +280,12 @@ def test_sbatch_generation(slurm_system: SlurmSystem, test_req: tuple[TestRun, s
         sbatch_script = slurm_system.output_path / "generated_command.sh"
     curr = Path(sbatch_script).read_text().strip()
 
-    assert curr == ref
+    if "nemo-launcher" in test_req[1]:
```
amaslenn, line 283: Why do we need normalization?

Our acceptance test failures are already tricky to debug, diffs are huge and don't exactly match what we have in reference files (path are expanded). Normalization introduces even more differences and also changes formatting which can be important.

I assume this is needed because of timestamps. I suggest we add `NeMoLauncherSlurmCommandGenStrategy.job_prefix` and mock it during object creation.

TaekyungHeo: Thanks, please find the updated code.

tests/test_acceptance.py
```diff
@@ -280,15 +278,10 @@ def test_sbatch_generation(slurm_system: SlurmSystem, test_req: tuple[TestRun, s
...
-    curr_norm = normalize_script(curr)
-    ref_norm = normalize_script(ref)
-    assert curr_norm == ref_norm, f"Mismatch:\nExpected:\n{ref_norm}\n\nActual:\n{curr_norm}"
+    assert curr == ref
 
     run_script = test_req[-1]
     if run_script:
-        curr_run_script = Path(slurm_system.output_path / "run.sh").read_text()
-        ref_run_script = (Path(__file__).parent / "ref_data" / run_script).read_text()
+        curr_run_script = Path(slurm_system.output_path / "run.sh").read_text().strip()
```
amaslenn: Is `.strip()` actually needed?

tests/test_acceptance.py
```diff
@@ -259,15 +258,14 @@ def create_test_run(name, test_definition, command_gen_strategy):
...
-def normalize_script(text: str) -> str:
-    return "\n".join(line.strip() for line in text.splitlines())
-
-
 def test_sbatch_generation(slurm_system: SlurmSystem, test_req: tuple[TestRun, str]):
     slurm_system.output_path.mkdir(parents=True, exist_ok=True)
 
     tr = test_req[0]
 
+    if isinstance(tr.test.test_template.command_gen_strategy, NeMoLauncherSlurmCommandGenStrategy):
```
amaslenn, line 266: Let's move it into `def create_test_run` (inside `def test_req`).

TaekyungHeo: Please find the updated code. I had to spin off some helper functions because of lint errors.

tests/test_acceptance.py
amaslenn, line 1: Thanks for refactoring it!

amaslenn (approved):

PR #391, opened 2025-03-04
Refactor NeMoLauncher report generation to report avg, min, max, and median

TaekyungHeo:
## Summary
Refactor NeMoLauncher report generation to report avg, min, max, and median. NeMoLauncher currently does not generate the same report as NeMoRun ([link](https://github.com/NVIDIA/cloudai/blob/7bdc060dd70ebf7fb7322136c36f7315c09dac48/src/cloudai/workloads/nemo_run/report_generation_strategy.py#L25)). The goal of this PR is to align the report format with NeMoRun by incorporating these statistical measures.

## Test Plan
1. CI passes
2. Ran on IL-1

### DIFF
--- a/src/cloudai/workloads/nemo_launcher/report_generation_strategy.py
+++ b/src/cloudai/workloads/nemo_launcher/report_generation_strategy.py
@@ -14,47 +14,88 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
+import logging
+import re
 from pathlib import Path
+from typing import List
 
+import numpy as np
 import pandas as pd
 
 from cloudai import ReportGenerationStrategy
 from cloudai.report_generator.tool.bokeh_report_tool import BokehReportTool
-from cloudai.report_generator.tool.tensorboard_data_reader import TensorBoardDataReader
 
 
 class NeMoLauncherReportGenerationStrategy(ReportGenerationStrategy):
-    """
-    Strategy for generating reports from NeMo launcher directories.
-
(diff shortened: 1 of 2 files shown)


srivatsankrishnan (approved):
LGTM

src/cloudai/workloads/nemo_launcher/report_generation_strategy.py
```diff
@@ -14,47 +14,88 @@
...
+    """Strategy for generating reports from NeMo launcher directories."""
 
     def can_handle_directory(self) -> bool:
-        for _, __, files in os.walk(self.test_run.output_path):
-            for file in files:
-                if file.startswith("events.out.tfevents"):
-                    return True
-        return False
+        run_dir = self.test_run.output_path / "run"
+        return (
```
amaslenn, line 34: Chat with Hua showed that we might have issues with such approach: `can_handle` returns true, but there are not real data and report generation can crash. My suggestion is to check that data is available as part of `can_hadle`. It introduce double work overhead, but IMO functional correctness is more important.

TaekyungHeo: This PR checks the availability of data (log*.out). Do you think we should also check the output values, such as the list of training step timings?

TaekyungHeo: I encountered the corner case you described (the log file exists, but the train_step_timing list is missing), and the current code handles this case without raising an exception.

PR #393, opened 2025-03-04
Control workload settings from scenario

amaslenn:
## Summary
Usage models:
1. Existing flow in which we have Test TOMLs and reference it in Scenario TOML.
2. 🆕 Like previous, but we can override Test configuration inside test case block.
3. 🆕 Fully define Scenario inline without any Test TOMLs. In this case `--tests-dir` CLI arg is optional and can be omitted.

A case in a scenario file (option (2) above):
```toml
[[Tests]]
id = "nccl.all_gather"
test_name = "ncc_all_gather"  # defined is a separate TOML
num_nodes = 2
time_limit = "00:20:00"

  [Tests.cmd_args]
  ngpus = 1
```

A case in a scenario file (fully defined, option (3) above):
```toml
[[Tests]]
id = "nccl.all_gather"
num_nodes = 2
time_limit = "00:20:00"

name = "all_gather"
description = "all_gather"
test_template_name = "NcclTest"

  [Tests.cmd_args]
  subtest_name = "all_reduce_perf_mpi"
  ngpus = 1
```

## Test Plan
1. CI.
2. Runs
    1. `cloudai run --system-config system.toml --test-scenario conf/common/test_scenario/nccl_tests_inline.toml`
    2. `cloudai run --system-config system.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml`
    3. `cloudai run --system-config system.toml --tests-dir conf/common/test --test-scenario conf/
common/test_scenario/nccl_test.toml`

## Additional Notes
1. Pydantic models become first class citizens in CloudAI and now available via `cloudai.models.` package.
2. `TestTemplate.name` was removed as it was used only for `__repr__`.


TaekyungHeo (approved):

conf/common/test_scenario/nccl_tests_inline.toml
```diff
@@ -0,0 +1,53 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name = "nccl-tests"
+
+[[Tests]]
+id = "nccl.all_reduce"
```
srinivas212, line 20: It is a bit confusing to have id as well as name

srinivas212 (changes requested):

conf/common/test_scenario/nccl_tests_inline.toml
amaslenn: `id` is the name, `test_name` is a reference to Test TOML. Doesn't make it less confusing though.

srinivas212 (commented):
LGTM

TaekyungHeo (dismissed):
This is an essential feature that we should have had earlier. The user interface is straightforward and intuitive. Thank you for making these changes.

BTW, have we documented how to use this feature? If not, please write one for it. It can be included in this PR or a follow-up PR.

srivatsankrishnan (commented):
Please document or add more information (for us atleast) in the PR for the following tests and what kind of knobs can be exposed? I want to understand the priority order and which one takes precedence. 
* Nemo-Run

For this test you ran, can you please updload the logs [here]. Can you also please confirm the container version you tested this on?

(https://drive.google.com/drive/folders/1XP1alXX80AkLnMdGeb1AHOG9tHQ_LOIy):
`cloudai run --system-config system.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml`

srinivas212:
Can we rename `test_spec` to `override` or `test_override`?

conf/common/test_scenario/nccl_tests_inline.toml
```diff
@@ -0,0 +1,53 @@
...
+num_nodes = 2
+time_limit = "00:20:00"
+
+  [Tests.test_spec]
+  name = "all_reduce"
+  description = "all_reduce"
+  test_template_name = "NcclTest"
+
+    [Tests.test_spec.cmd_args]
```
srinivas212, line 29: Can we rename test_spec to override or test_override?

test_spec is a new term we should avoid

TaekyungHeo: My preference is `test_spec` over `test_override`. If it is controversial, we can remove `test_spec.` field completely.
```
    [Tests.cmd_args]
```

srivatsankrishnan:
> Please document or add more information (for us atleast) in the PR for the following tests and what kind of knobs can be exposed? I want to understand the priority order and which one takes precedence.
> 
> * Nemo-Run
> 
> For this test you ran, can you please updload the logs [here]. Can you also please confirm the container version you tested this on?
> 
> (https://drive.google.com/drive/folders/1XP1alXX80AkLnMdGeb1AHOG9tHQ_LOIy): `cloudai run --system-config system.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml`

[@Srivatsan Krishnan](https://nvidia.slack.com/team/U07663SPLDU)

-  I'm providing the info you are asking, but this is not efficient going forward. If you think that PR (this and any future one) can break NemoRun jobs, please describe how. And why acceptance test is not enough. Let's improve our coverage.
- 
Unfortunately, acceptance tests and unit tests only provides us so far. We had documented cases with Jaxtoolbox and now with Nemo. I think its just the nature of the LLM training itself.  We run into issues can can update the acceptance test/unit tests. But lot of times esp with Nemo, there are many moving changes. So the request here is to have a real system testing. If you establish a pattern based on real-system testing, I agree, we can add those to the acceptance test and coverage. The methodology you are proposing works for stable benchmarks that doesn't change much.

Also Nemo2.0 is pretty complex and unless we gets hands-on testing (understanding how it works and reviewing the logs), it would be hard to learn and understand how it works. I think its high time each one us gathers the experience of running and understanding how Nemo2.0 works. Then work backwards to see if a particular feature makes sense or not. Its not just about sbatch generation here. We constantly will update the containers (meaning our dependencies with Nemo-CLI etc will also needs to be validated). So unless we have a clear signal about these things, please test it on real system.

Specific to this PR, we have had lots of discussions on which takes precedences when we set a variable in test and test scenario. So the ask here is
* Document for us which value will be final and honored. What happens when we set the same variable 
   * Defined in test definition but not exposed in test and/or test scenario.
   * Defined in test definition but not exposed in test but defined in test scenario
   * Not defined in test definition (we allow this today btw), but exposed in test and test scenario.
Lets set one variable and test this?


-If you will be asking the same verification again and again anyway, please provide one-click automation for this.

Unfortantely, I don't have BW to provide you. I can certainly give recipes and my learning for what works for me. But if you have a better way to solve this problem through one-click, please feel to contribute. We are solving system software problem and not purely software. So we need to strengthen our real system testing and if there are established patterns, then we can move it to acceptance tests. IMO, it should not be the other way.

TaekyungHeo (approved):

conf/common/test_scenario/nccl_tests_inline.toml
srinivas212: THis was my original suggestion but I think @amaslenn  had few concerns.

srinivas212:
> This is an essential feature that we should have had earlier. The user interface is straightforward and intuitive. Thank you for making these changes.
> 
> BTW, have we documented how to use this feature? If not, please write one for it. It can be included in this PR or a follow-up PR.

I agree. I have learnt to love this. My main issue is the new name we are using. I'm not a fan of that. Any other ways we can do this would be great.

conf/common/test_scenario/nccl_tests_inline.toml
amaslenn: The issue with removing it is this:
```py
class TestRunModel(BaseModel):
    id: str = Field(min_length=1)
    test_name: Optional[str] = None
    num_nodes: Optional[int] = None
    nodes: list[str] = Field(default_factory=list)
    weight: int = 0
    iterations: int = 1
    sol: Optional[float] = None
    ideal_perf: float = 1.0
    time_limit: Optional[str] = None
    dependencies: list[TestRunDependencyModel] = Field(default_factory=list)
    test_spec: Optional[TestSpecModel] = None
```

Notice that it is already rather big. `TestSpecModel` is this:
```py
class TestSpecModel(BaseModel):
    name: Optional[str] = None
    description: Optional[str] = None
    test_template_name: Optional[str] = None
    cmd_args: Optional[CmdArgs] = None
    extra_container_mounts: list[str] = []
    git_repos: list[GitRepo] = []
    nsys: Optional[NsysConfiguration] = None
    agent: str = "grid_search"
    agent_steps: int = 1
```
It is basically a copy of TestDefinition model. If we inline one into another, `TestRunModel` will become longer and less organized.

amaslenn: I've removed `.test_spec`, guess it is not right to ruin UI because of code structure. Let me know what do you think about new approach.

P.S. @TaekyungHeo thanks so much for your reviews, but yet another will be needed :)

srivatsankrishnan (approved):
Thanks for the updating the logs. LGTM.

TaekyungHeo (approved):

srinivas212 (approved):

srinivas212:
Thank you so much for working thru all the issues and making this work, @amaslenn. This is amazing!

PR #394, opened 2025-03-04
Allow skipping cache validation

amaslenn:
## Summary
Introduce `--enable-cache-without-check` for `run` and `dry-run` to skip real check for cache existence, but still build all paths correctly. This feature will be used for Slurm systems where head node doesn't have access to shared file system with compute nodes. The flow is like this:
1. [on the head node] run `cloudai install`
2. [on the head node] copy cache to compute nodes
3. Modify system.toml to set compute nodes' installation root
4. Run `cloudai run --enable-cache-without-check ...`

## Test Plan
1. CI
2. Arik verified it on his system with mentioned properties.

## Additional Notes
—


TaekyungHeo (dismissed):
* Not urgent, but we may need to update USER_GUIDE.md accordingly.
* It is a minor update, and I agree that it is necessary, but this change affects interfaces. Let me mark it as a feature to be reviewed by multiple contributors.

amaslenn:
> Not urgent, but we may need to update USER_GUIDE.md accordingly. ...

Good catch, I think it is important to add it into our doc now, later it will be harder to recall what it is about.

TaekyungHeo (approved):

srivatsankrishnan (commented):
> Modify system.toml to set compute nodes' installation root

What exactly is this step? Which clusters/system do you think this is needed?

amaslenn:
> What exactly is this step? Which clusters/system do you think this is needed?

This is internal pre-setups and the way it is being built. We've discussed it with Arik.

srivatsankrishnan (approved):
LGTM

PR #398, opened 2025-03-05
Move signal handling from BaseRunner to Runner

amaslenn:
## Summary
Move signal handling from BaseRunner to Runner. Reason is to clear responsibilities, `BaseRunner` servers as job scheduler and `Runner` is a generic wrapper, a manager that is also responsible for actual runner creation. So it make more sense to handle signals in the manager.

## Test Plan
1. CI.
2. Manual runs + `Ctrl+C`.

## Additional Notes
—


TaekyungHeo (approved):
- The design makes sense. (BaseRunner is simply a generic wrapper.)
- It's good to know that this PR speeds up job cancellation. That has been a concern.

Let's have one more review.

srivatsankrishnan (approved):
Internal to CloudAI. I don't have a strong opinion on this. LGTM.

PR #404, opened 2025-03-07
Support installing PythonExecutable in a subpath

TaekyungHeo:
## Summary
Support installing PythonExecutable in a subpath

## Test Plan
1. CI passes
2. Tested on EOS with https://github.com/Mellanox/cloudaix/pull/177
```
$ python cloudaix.py install --system-config conf/common/system/eos.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/nccl_test.toml
```


TaekyungHeo:
@amaslenn, thanks for the comments. I have updated the code accordingly. I tested it on an internal server and confirmed that the target binaries are installed.

src/cloudai/installer/slurm_installer.py
```diff
@@ -236,13 +236,25 @@ def _install_python_executable(self, item: PythonExecutable) -> InstallStatusRes
...
-        res = self._install_requirements(venv_path, requirements_txt)
-        if not res.success:
-            return res
 
-        item.venv_path = venv_path
+        project_dir = item.git_repo.installed_path
+        if item.project_subpath:
+            project_dir = project_dir / item.project_subpath
+
+        requirements_txt = project_dir / "requirements.txt"
```
amaslenn, line 244: I don't think we should decide based on file existence: many projects today have both `requirements.txt` and `pyproject.toml` and assumption to use one or another might be wrong.

One solution here could be introducing `install_from_pyproject` flag for the installable object.

Also since `_install_requirements` and `_install_pyproject` are almost identical, can we merge them into one function and make a decision based on the flag?

P.S. sorry, this comment had to be part of previous review, I forgot to save it.

TaekyungHeo: I agree that modern python projects use both requirements.txt and pyproject.toml simultaneously. I removed the conditional statements accordingly. However, merging _install_requirements and _install_pyproject is not a good option because _install_requirements is already used by other functions.

src/cloudai/installer/slurm_installer.py
```diff
@@ -274,16 +282,30 @@ def _create_venv(self, venv_dir: Path) -> InstallStatusResult:
...
-        logging.debug(f"Installing requirements from {requirements_txt} using command: {install_cmd}")
         result = subprocess.run(install_cmd, capture_output=True, text=True)
+
+        if result.returncode != 0:
+            return InstallStatusResult(False, f"Failed to install dependencies from requirements.txt: {result.stderr}")
+
+        return InstallStatusResult(True)
+
+    def _install_pyproject(self, venv_dir: Path, project_dir: Path) -> InstallStatusResult:
+        python_bin = venv_dir / "bin" / "python"
```
amaslenn, line 297: IMO these checks are excessive too:
1. venv is created before, so python is available (+ we don't have this check while using requirements.txt).
2. project dir is a subdir of the repo, so it should exist. If not — error will be raised during pip run. But if you think that this error improves readability, I'm fine with keeping it. Although I think pip should show a good message in this case as well.

src/cloudai/installer/slurm_installer.py
amaslenn: But now we install twice, first with requirements and then with pyproject (when both exist). IMO we need to run installation only once.

TaekyungHeo: What do you mean by run installation only once?
1. Run `pip install .` and do not run `pip install -r requirements.txt`
2. Run `pip install -r requirements.txt & `pip install .` but merge them in a single function

src/cloudai/installer/slurm_installer.py
```diff
@@ -236,13 +236,25 @@ def _install_python_executable(self, item: PythonExecutable) -> InstallStatusRes
...
+        pyproject_toml = project_dir / "pyproject.toml"
+        requirements_txt = project_dir / "requirements.txt"
+
+        if pyproject_toml.exists():
+            res = self._install_pyproject(venv_path, project_dir)
+        elif requirements_txt.exists():
+            res = self._install_requirements(venv_path, requirements_txt)
+        else:
+            return InstallStatusResult(False, "No pyproject.toml or requirements.txt found for installation.")
+
```
TaekyungHeo, line 253: @amaslenn, please find the updated code. I am no longer calling _install_requirements if pyproject.toml exists. As a result, pip install -r requirements.txt is not executed twice.

Please note that I did not merge _install_pyproject and _install_requirements to support both NeMoLauncher and Maya. NeMoLauncher has a requirements.txt file but does not have a pyproject.toml file, whereas Maya has pyproject.toml. Merging these two functions would be somewhat awkward.

Instead, I would prefer merging all of the code into a single function (_install_python_executable). However, doing so would significantly increase the function’s size.

TaekyungHeo: Note: At this point, I don’t believe it is necessary to add a switch to choose between `requirements.txt` and `pyproject.toml`. Additionally, I am not certain if this is the best way to manage it. For these reasons, I have not included it in this PR.

TaekyungHeo: Sorry, I didn’t get your point clearly. I believe this order is valid: if `pyproject.toml` exists, we ignore `requirements.txt` and use `pyproject.toml` instead. What is your suggestion here?

> IMO later should have higher priority.

Are you suggesting that `requirements.txt` should have a higher priority? If that is the case, then even if `pyproject.toml` exists, only `requirements.txt` would be used.

If we conditionally use both `requirements.txt` and `pyproject.toml`, it would result in the same implementation as the original code (running the installation twice).

I think the current approach is reasonable, so we can proceed. Let’s revisit this later if it becomes an issue.

TaekyungHeo: If we install requirements.txt only when it is found, we will miss the binary aliases in pyproject.toml. Therefore, we need to go back to the original implementation: install requirements.txt if it exists and pyproject.toml if it exists.

TaekyungHeo: @amaslenn and I had an offline discussion. I will introduce `dependencies_from_pyproject` and will set it to True by default.

TaekyungHeo: @amaslenn Please find the updated code.

tests/test_slurm_installer.py
```diff
@@ -221,15 +219,22 @@ def test_requirements_not_installed(self, installer: SlurmInstaller):
...
+        assert res.message == "Failed to install dependencies from requirements.txt: err"
 
     def test_all_good_flow(self, installer: SlurmInstaller, git: GitRepo):
         py = PythonExecutable(git)
         py.git_repo.installed_path = installer.system.install_path / py.git_repo.repo_name
+
+        repo_dir = py.git_repo.installed_path
+        repo_dir.mkdir(parents=True, exist_ok=True)
+        pyproject_file = repo_dir / "pyproject.toml"
+        pyproject_file.write_text("[tool.poetry]\nname = 'dummy_project'")
```
amaslenn, line 231: Is it necessary to add this line?

tests/test_slurm_installer.py
```diff
@@ -286,6 +291,64 @@ def test_uninstall_venv_removed_ok(self, installer: SlurmInstaller, git: GitRepo
...
+    def test_install_python_executable_prefers_pyproject_toml(self, installer: SlurmInstaller, git: GitRepo):
+        repo_dir = installer.system.install_path / git.repo_name
+        repo_dir.mkdir(parents=True)
+        subdir = repo_dir / "subdir"
+        subdir.mkdir()
+
+        pyproject_file = subdir / "pyproject.toml"
+        requirements_file = subdir / "requirements.txt"
+
+        pyproject_file.write_text("[tool.poetry]\nname = 'project'")
```
amaslenn, line 303: Same as above, is it necessary to write data into files?

tests/test_slurm_installer.py
```diff
@@ -286,6 +291,64 @@ def test_uninstall_venv_removed_ok(self, installer: SlurmInstaller, git: GitRepo
...
+        py.git_repo.installed_path = repo_dir
+
+        res = installer._install_python_executable(py)
+
+        assert res.success
+        installer._install_pyproject.assert_called_once_with(installer.system.install_path / py.venv_name, subdir)
+        installer._install_requirements.assert_not_called()
+        assert py.venv_path == installer.system.install_path / py.venv_name
+
+    def test_install_python_executable_prefers_requirements_txt(self, installer: SlurmInstaller, git: GitRepo):
```
amaslenn, line 322: [minor, feel free to ignore] you can introduce a fixture and re-use it for two new tests.

tests/test_slurm_installer.py
TaekyungHeo: Replaced `write_text` with `touch` but anyway we need the file.

tests/test_slurm_installer.py
TaekyungHeo: Replaced `write_text` with `touch` but anyway we need the file.

tests/test_slurm_installer.py
TaekyungHeo: Please find the updated code.

amaslenn (approved):

PR #409, opened 2025-03-12
[DRAFT][RAY] Add slurm ray tests

AgrawalAmey:
## Summary
This PR adds a new test category which runs ray applications with slurm.


conf/common/test/slurm_ray_container_vllm.toml
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
amaslenn, line 2: For new files please set a single year value (diagnostic we have today is misleading)

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,78 @@
...
+
+from .slurm_ray_container import SlurmRayContainerTestDefinition
+
+
+class SlurmRayContainerCommandGenStrategy(SlurmContainerCommandGenStrategy):
+    """Command generation strategy for generic Slurm container tests."""
+
+    def _get_sbatch_directives(self, args: Dict[str, Any], output_path: Path) -> Dict[str, str]:
+        sbatch_directives = super()._get_sbatch_directives(args, output_path)
+        # TODO(Amey): We probably need to figure out what to do with cpus-per-task, mem-per-cpu
```
amaslenn, line 33: This can be set with `SlurmSystem.extra_sbatch_args`. The downside is that it is set per System, so all tests in a scenario will have it.

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
```diff
@@ -0,0 +1,78 @@
...
+        command_parts: list[str] = [tdef.cmd_args.cmd]
+        if tr.test.extra_cmd_args:
+            command_parts.append(tr.test.extra_cmd_args)
+
+        # load the jinja template file which is placed at the same directory as this file
+        script_dir = Path(__file__).parent
+        template_path = script_dir / "slurm_ray_container_template.sh.jinja"
+        template = Template(template_path.read_text())
+
+        conda_activate_command = f"conda activate {tdef.cmd_args.conda_env} && " if tdef.cmd_args.conda_env else ""
```
amaslenn, line 67: Please help me understanding this part. Isn't env for `ray` is ready inside a container? Why this extra env needed?

In CloudAI we have a concept of `installable`: items that should be "installed" before run (done with `cloudai install ...`). Examples: docker images, git repos with python scripts (in this case we can create venv for it), etc. Repos can be mount into a container to have files available.

src/cloudai/workloads/slurm_ray_container/slurm_ray_container_template.sh.jinja
amaslenn, line 1: Thanks! I was thinking of using jinja for our sbatch generation, this change will hopefully push me to work on it :)

tests/test_acceptance.py
```diff
@@ -261,6 +266,20 @@ def test_req(request, slurm_system: SlurmSystem, partial_tr: partial[TestRun]) -
             ),
             SlurmContainerCommandGenStrategy,
         ),
+        "slurm_ray_container": lambda: create_test_run(
```
amaslenn, line 269: Please also update fixture.params for this one, otherwise this case will not run.

amaslenn (commented):
@AgrawalAmey thanks a lot for your contribution! And sorry for the late feedback.

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
AgrawalAmey: Basically, i want this to be dynamic, as a fraction of total resources

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
AgrawalAmey: Essentially, this is supposed to be an optional parameter to activate a specific environment if required. For instance, in the Vajra nightly perf test container, we have multiple envs for vllm, vajra, sglang etc.

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
AgrawalAmey: Since we have to set the task per worker to 1 for ray, we need to ensure that all the resources are made available to the process.

src/cloudai/workloads/slurm_ray_container/slurm_command_gen_strategy.py
amaslenn: I'm concerned that `SlurmRayContainer` becomes too Vajra-specific. This shouldn't be a blocker, but if we can generalize it, would be great. I don't have a good idea so far.

PR #334, opened 2025-01-12
Safety valve to seperate the DSE execution with benchmarking execution

srivatsankrishnan:
## Summary
This PR separates the DSE execution with benchmarking execution by identifying the job type (`DSE` or Non-DSE (i.e., benchmarking) by dynamically parsing the TestRun object. If it's a DSE job, the agent and environment integration happens along with iterators. If it's a regular job, then it will not instantiate the agent or the environment.

### Input
* Test toml file

## Output
* if DSE ranges are provides, will insatiate agent/env <--> runner
* if not DSE, then original execution flow

## Test Plan
CI/CD

### Result structure for DSE and Non-DSE Job

```
results/
└── dse_jaxtoolbox_grok
    └── Tests.1
        └── 0
            ├── 1
            │   ├── cloudai_sbatch_script.sh
            │   └── run.sh
            ├── 2
            │   ├── cloudai_sbatch_script.sh
            │   └── run.sh
            ├── 3
            │   ├── cloudai_sbatch_script.sh
            │   └── run.sh
            ├── 4
            │   ├── cloudai_sbatch_script.sh
            │   └── run.sh
            ├── 5
            │   ├── cloudai_sbatch_script.sh
            │   └── run.sh
            └── 6
                ├── cloudai_sbatch_script.sh
                └── run.sh
```

Non-DSE Result Structure (Unchanged)

```
results/nccl-test_2025-01-17-00-43-34/
├── Tests.1
│   └── 0
│       ├── cloudai_sbatch_script.sh
│       ├── post_test
│       │   └── nccl_test_all_gather
│       └── pre_test
│           └── nccl_test_all_gather
├── Tests.10
│   └── 0
│       ├── cloudai_sbatch_script.sh
│       ├── post_test
│       │   └── nccl_test_all_gather
│       └── pre_test
│           └── nccl_test_all_gather
├── Tests.11
│   └── 0
│       ├── cloudai_sbatch_script.sh
│       ├── post_test
│       │   └── nccl_test_all_gather
│       └── pre_test
│           └── nccl_test_all_gather
├── Tests.12
│   └── 0
│       ├── cloudai_sbatch_script.sh
│       ├── post_test
│       │   └── nccl_test_all_gather
│       └── pre_test
│           └── nccl_test_all_gather
├── Tests.13
│   └── 0
│       ├── cloudai_sbatch_script.sh
│       ├── post_test
│       │   └── nccl_test_all_gather
│       └── pre_test
│           └── nccl_test_all_gather
[...]

PR #431, opened 2025-03-26
Scenario reporter for DSE jobs

amaslenn:
## Summary
1. Unified results output dir name for benchmark and DSE, last one will now also include timestamp.
2. Build correct scenario level report taking into account DSE steps.

## Test Plan
1. CI
3. Manual runs for report generation.

## Additional Notes
—


srivatsankrishnan (approved):
Thanks. Can you please update the result directory structure so we can visually see what it looks like for benchmarking and DSE job now?

The only difference before was the benchmarking job had a timestamp and the DSE job didn't have a timestam? So the latter would basically override the generated batch scripts. Correct?

The original design decision was agreed and discussed: https://github.com/NVIDIA/cloudai/pull/334

TaekyungHeo (approved):

PR #442, opened 2025-04-01
Remove redundant system initialization in LSF and Slurm installers

TaekyungHeo:
## Summary
Remove redundant system initialization in LSF and Slurm installers.
Please refer to https://github.com/NVIDIA/cloudai/blob/7e5056ad957f845992c0fc429ed3f35d5be0a992/src/cloudai/_core/base_installer.py#L48

## Test Plan
CI passes

### DIFF
--- a/src/cloudai/installer/lsf_installer.py
+++ b/src/cloudai/installer/lsf_installer.py
@@ -18,7 +18,6 @@ import logging
 
 from cloudai import BaseInstaller, DockerImage, File, GitRepo, Installable, InstallStatusResult, PythonExecutable
 from cloudai.installer.slurm_installer import SlurmInstaller
-from cloudai.systems import LSFSystem
 
 
 class LSFInstaller(BaseInstaller):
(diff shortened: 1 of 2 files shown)


src/cloudai/installer/lsf_installer.py
```diff
@@ -34,16 +33,6 @@ class LSFInstaller(BaseInstaller):
 
     PREREQUISITES = ("bsub", "bjobs", "bhosts", "lsid", "lsload")
 
-    def __init__(self, system: LSFSystem):
```
amaslenn, line 37: I think it was mostly for types, now LSFInstaller thinks that it works with generic System, while in fact can only be LSF. Same for Slurm.

TaekyungHeo: Does it mean that `self.system = system` is essential?

amaslenn: `system: LSFSystem` + `self.system = system` are essential. Looks like we don't use any sub-system specific features, hence pyright is silent.

TaekyungHeo: > I don't think we can remove this. Why is this PR necessary?

I wanted to fix it as I identified this while working on RunAI.

> system: LSFSystem + self.system = system are essential. 

Sounds good, then let's keep as it is.

PR #445, opened 2025-04-02
Register reports via Registry

amaslenn:
## Summary
To unify internals and provide extensibility move to use `Registry` for registering reports.

## Test Plan
CI

## Additional Notes
—

### DIFF
--- a/tests/test_test_scenario.py
+++ b/tests/test_test_scenario.py
@@ -21,11 +21,10 @@ from unittest.mock import Mock, patch
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/test_test_scenario.py
```diff
@@ -282,7 +283,8 @@ def test_default(self):
         assert len(reporters) == 0
 
     def test_default_reporters_size(self):
-        assert len(DEFAULT_REPORTERS) == 11
+        print(Registry().reports_map)
```
TaekyungHeo, line 286: Do we need this print here?

amaslenn: No, debug print, removed :) Thanks!

TaekyungHeo (approved):

PR #451, opened 2025-04-03
Check job completion before scheduling post-init dependent tests

TaekyungHeo:
## Summary
Bug fix for https://redmine.mellanox.com/issues/4388097.

Root cause: The second job depends on the first being initialized (i.e., scheduled and running). On Slurm, jobs aren’t scheduled immediately, so the initial dependency check returns False. The second job waits 60 seconds, then checks again—by then, the first job has already finished, so it goes back to sleep and blocks indefinitely.

## Test Plan
1. CI passes
2. Ran on IL-1
```cloudai run --system-config conf/common/system/israel_1.toml --tests-dir conf/common/test  --test-scenario conf/common/test_scenario/sleep.toml
[INFO] System Name: Israel-1
[INFO] Scheduler: slurm
[INFO] Test Scenario Name: test_scenario_example
[INFO] Test Scenario: test_scenario_example

Section Name: Tests.1
  Test Name: sleep_10
  Description: sleep_10
  No dependencies
Section Name: Tests.2
  Test Name: sleep_5
  Description: sleep_5
  Start Post Init: Tests.1
Section Name: Tests.3
  Test Name: sleep_5
  Description: sleep_5
  Start Post Comp: Tests.1
Section Name: Tests.4
  Test Name: sleep_20
  Description: sleep_20
  End Post Comp: Tests.1
[INFO] Creating SlurmRunner
[INFO] Starting test: Tests.1
[INFO] Running test: Tests.1
[INFO] Submitted slurm job: 448121
[INFO] Starting test: Tests.4
[INFO] Running test: Tests.4
[INFO] Submitted slurm job: 448122
[INFO] Delayed start for test Tests.2 by 5 seconds.
[INFO] Starting test: Tests.2
[INFO] Running test: Tests.2
[INFO] Submitted slurm job: 448123
[INFO] Job completed: Tests.1 (iteration 1 of 1)
[INFO] Delayed start for test Tests.3 by 5 seconds.
[INFO] Starting test: Tests.3
[INFO] Running test: Tests.3
[INFO] Submitted slurm job: 448124
[INFO] Scheduling termination of job 448122 after 0 seconds.
[INFO] Job completed: Tests.4 (iteration 1 of 1)
[INFO] Job completed: Tests.2 (iteration 1 of 1)
[INFO] Job completed: Tests.3 (iteration 1 of 1)
[INFO] All test scenario results stored at: results/test_scenario_example_2025-04-04_12-07-54
[INFO] Generated scenario report at results/test_scenario_example_2025-04-04_12-07-54/test_scenario_example.html
[WARNING] Skipping 'results/test_scenario_example_2025-04-04_12-07-54/Tests.1/0', can't handle with strategy=SleepReportGenerationStrategy.
[WARNING] Skipping 'results/test_scenario_example_2025-04-04_12-07-54/Tests.2/0', can't handle with strategy=SleepReportGenerationStrategy.
[WARNING] Skipping 'results/test_scenario_example_2025-04-04_12-07-54/Tests.3/0', can't handle with strategy=SleepReportGenerationStrategy.
[WARNING] Skipping 'results/test_scenario_example_2025-04-04_12-07-54/Tests.4/0', can't handle with strategy=SleepReportGenerationStrategy.
[INFO] All jobs are complete. For more details, please review the 'debug.log'..```

### DIFF
--- a/src/cloudai/_core/command_gen_strategy.py
+++ b/src/cloudai/_core/command_gen_strategy.py
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
-# Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -21,11 +21,7 @@ from .test_template_strategy import TestTemplateStrategy
 
 
 class CommandGenStrategy(TestTemplateStrategy, ABC):
-    """
-    Abstract base class defining the interface for command generation strategies across different system environments.
-
-    It specifies how to generate execution commands based on system and test parameters.
-    """
+    """Abstract base class defining the interface for command generation strategies across different systems."""
 
     @abstractmethod
     def gen_exec_command(self, tr: TestRun) -> str:
@@ -39,28 +35,3 @@ class CommandGenStrategy(TestTemplateStrategy, ABC):
             str: The generated execution command.
         """
         pass
-
-    @abstractmethod
-    def gen_srun_command(self, tr: TestRun) -> str:
-        """
-        Generate the Slurm srun command for a test based on the given parameters.
-
-        Args:
-            tr (TestRun): Contains the test and its run-specific configurations.
-
-        Returns:
-            str: The generated Slurm srun command.
-        """
-        pass
-
-    def gen_srun_success_check(self, tr: TestRun) -> str:
-        """
-        Generate the Slurm success check command to verify if a test run was successful.
-
-        Args:
-            tr (TestRun): Contains the test and its run-specific configurations.
-
-        Returns:
-            str: The generated command to check the success of the test run.
-        """
-        return ""
--- a/src/cloudai/_core/test_template.py
+++ b/src/cloudai/_core/test_template.py
@@ -91,6 +91,8 @@ class TestTemplate:
         return self.command_gen_strategy.gen_exec_command(tr)
 
     def gen_srun_command(self, tr: TestRun) -> str:
+        from ..systems.slurm.strategy.slurm_command_gen_strategy import SlurmCommandGenStrategy
+
         """
         Generate an Slurm srun command for a test using the provided command generation strategy.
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Review by amaslenn — COMMENTED** [review]
This is not a fix IMO. User system can set own limit and then we have this bug again.

**Comment by TaekyungHeo** [comment]
@amaslenn I agree. Do you have any suggestion?

**Comment by amaslenn** [comment]
> @amaslenn I agree. Do you have any suggestion?

Do a proper fix :) We need to understand in depth why it happens and fix it for good no mater what timeout is used.

**Comment by TaekyungHeo** [comment]
@amaslenn I’ve been thinking about this for a while, but I haven’t found a better solution than increasing the polling frequency. I agree that changing the default value is risky, but ultimately the solution is to enforce a lower interval. Other than that, I don’t see a viable alternative. There are two approaches: polling or interrupt. Since we interface with systems via commands (e.g., `squeue`) or REST APIs (e.g., RunAI), there’s no way for the systems to interrupt CloudAI. We also discussed using Slurm-native primitives, but that doesn’t apply to other systems and adds complexity. So for now, polling is the only option — and that means the only viable path is to increase the polling frequency.

**Comment by amaslenn** [comment]
> ... So for now, polling is the only option — and that means the only viable path is to increase the polling frequency.

I don't argue that polling is our only option — agreed.

But this is not the fix: user's system configs won't be affected by this change if they specify monitoring interval and bug will be re-opened.

We need to understand why interval affects this at all: we rely on `sacct` today, so job can't just disappear.

**Comment by TaekyungHeo** [comment]
> But this is not the fix: user's system configs won't be affected by this change if they specify monitoring interval and bug will be re-opened.

I agree.

My suggestion is to enforce a high-frequency monitoring interval (up to ~10 seconds). I agree that setting the default to 1 second isn’t the answer, but we should ensure frequent checks.

Here’s why: the second job depends on the first being initialized (i.e., scheduled and running). On Slurm, jobs aren’t scheduled immediately, so the initial dependency check returns False. The second job waits 60 seconds, then checks again—by then, the first job has already finished, so it goes back to sleep and blocks indefinitely.

One alternative is to treat job completion as a valid condition to unblock the second job. Let me know what you think.

**Comment by amaslenn** [comment]
> Here’s why: the second job depends on the first being initialized (i.e., scheduled and running). On Slurm, jobs aren’t scheduled immediately, so the initial dependency check returns False. The second job waits 60 seconds, then checks again—by then, the first job has already finished, so it goes back to sleep and blocks indefinitely.
> 
> One alternative is to treat job completion as a valid condition to unblock the second job. Let me know what you think.

I don't see it as a problem of timeout, but rather our internal logic. We know, that for "start_post_init" on slurm we don't need to check if main job is running — let's submit them one by one anyway. Here "init" is CloudAI's term anyway.

Frequent polling can be a problem on some systems, our logic should not depend on this value.

File: src/cloudai/_core/base_runner.py [hunk 1]
```diff
@@ -145,7 +145,7 @@ async def check_start_post_init_dependencies(self):
         items = list(self.testrun_to_job_map.items())
 
         for tr, job in items:
-            if self.mode == "dry-run" or self.system.is_job_running(job):
+            if self.mode == "dry-run" or self.system.is_job_running(job) or self.system.is_job_completed(job):
```
amaslenn, line 148: Really? :) Was issues because main job was completed already?

I still think that we need to schedule such jobs together, otherwise the whole point of such dependency is not met. But as a fix this is fine, let's have a second opinion.

TaekyungHeo: This is an idea and I need to test it. I will let you know if this works.

TaekyungHeo: It works. Please let me know if I can merge.

> I still think that we need to schedule such jobs together, otherwise the whole point of such dependency is not met.

I agree, that's why I proposed enforcing higher frequencies initially. But I also agree that DDOSing is a real risk as well. Let's move on with this fix if you are good.

amaslenn: Works for me, just let's here a second though on it.

PR #453, opened 2025-04-03
Replace pyright ignores with explicit type handling

TaekyungHeo:
## Summary
Replace pyright ignores with explicit type handling

## Test Plan
CI passes


TaekyungHeo:
@amaslenn We have two options:

1. Fix it first with this PR , discuss, and improve
2. Discuss and decide how to handle such cases, decide, and fix.

Which do you prefer? I agree this PR isn't ideal, but we need to make a decision. If you prefer option #2, I'll go ahead and close this. My preference is #1—otherwise, it might remain unresolved indefinitely.

amaslenn (approved):

PR #450, opened 2025-04-03
Move Slurm-specific methods to SlurmCommandGenStrategy

TaekyungHeo:
## Summary
I realized that there is a design flaw in CommandGenStrategy and SlurmCommandGenStrategy while working on [this issue](https://redmine.mellanox.com/issues/4388097). The methods gen_srun_command and gen_srun_success_check are currently defined in CommandGenStrategy instead of SlurmCommandGenStrategy. Since CommandGenStrategy is an abstract base class intended for any command generation strategy, defining Slurm-specific methods here is not appropriate. This design flaw led to an issue when running the sleep test.
```
$ pip install .
$ cloudai run --system-config conf/common/system/standalone_system.toml --tests-dir conf/common/test  --test-scenario conf/common/test_scenario/sleep.toml
Traceback (most recent call last):
  File "/Users/theo/venv/bin/cloudai", line 8, in <module>
    sys.exit(main())
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/__main__.py", line 24, in main
    rc = cloudai_cli.run()
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/cli/cli.py", line 163, in run
    return self.handlers[args.mode](args)
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/cli/handlers.py", line 147, in handle_dry_run_and_run
    system, tests, test_scenario = parser.parse(args.tests_dir, args.test_scenario)
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/parser.py", line 80, in parse
    tests = self.parse_tests(list(test_path.glob("*.toml")), system)
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/parser.py", line 148, in parse_tests
    tests: List[Test] = test_parser.parse_all()
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/_core/test_parser.py", line 71, in parse_all
    parsed_object = self._parse_data(data)
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/_core/test_parser.py", line 212, in _parse_data
    test_template = self._get_test_template(test_template_name, test_def)
  File "/Users/theo/venv/lib/python3.10/site-packages/cloudai/_core/test_parser.py", line 180, in _get_test_template
    self._fetch_strategy(CommandGenStrategy, type(obj.system), type(tdef), cmd_args),
[...]

PR #454, opened 2025-04-04
Move Slurm-specific methods to SlurmCommandGenStrategy

TaekyungHeo:
## Summary
Move Slurm-specific methods to SlurmCommandGenStrategy

Follow up for https://github.com/NVIDIA/cloudai/pull/450

Discussion: https://github.com/NVIDIA/cloudai/pull/450#discussion_r2027315601

## Test Plan
CI passes

### DIFF
--- a/src/cloudai/_core/test_template.py
+++ b/src/cloudai/_core/test_template.py
@@ -90,50 +90,6 @@ class TestTemplate:
             )
         return self.command_gen_strategy.gen_exec_command(tr)
 
-    def gen_srun_command(self, tr: TestRun) -> str:
-        from ..systems.slurm.strategy.slurm_command_gen_strategy import SlurmCommandGenStrategy
-
-        """
-        Generate an Slurm srun command for a test using the provided command generation strategy.
-
-        Args:
-            tr (TestRun): Contains the test and its run-specific configurations.
-
-        Returns:
-            str: The generated Slurm srun command.
-        """
-        if self.command_gen_strategy is None:
-            raise ValueError(
-                "command_gen_strategy is missing. Ensure the strategy is registered in the Registry "
-                "by calling the appropriate registration function for the system type."
-            )
-        if isinstance(self.command_gen_strategy, SlurmCommandGenStrategy):
-            return self.command_gen_strategy.gen_srun_command(tr)
-        else:
-            raise TypeError("command_gen_strategy is not of type SlurmCommandGenStrategy")
-
-    def gen_srun_success_check(self, tr: TestRun) -> str:
-        from ..systems.slurm.strategy.slurm_command_gen_strategy import SlurmCommandGenStrategy
-
-        """
-        Generate a Slurm success check command for a test using the provided command generation strategy.
-
-        Args:
-            tr (TestRun): Contains the test and its run-specific configurations.
-
-        Returns:
-            str: The generated command to check the success of the test run.
-        """
-        if self.command_gen_strategy is None:
-            raise ValueError(
-                "command_gen_strategy is missing. Ensure the strategy is registered in the Registry "
-                "by calling the appropriate registration function for the system type."
-            )
-        if isinstance(self.command_gen_strategy, SlurmCommandGenStrategy):
-            return self.command_gen_strategy.gen_srun_success_check(tr)
-        else:
-            raise TypeError("command_gen_strategy is not of type SlurmCommandGenStrategy")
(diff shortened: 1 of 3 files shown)


tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
```diff
@@ -131,132 +131,127 @@ def test_time_limit(time_limit: Optional[str], strategy_fixture: SlurmCommandGen
...
+            testrun_fixture.post_test = TestScenario(name="post", test_runs=post_runs)
+        else:
+            testrun_fixture.post_test = None
+
+        sbatch_command = strategy_fixture.gen_exec_command(testrun_fixture)
+        script_path = sbatch_command.split()[-1]
+        content = Path(script_path).read_text()
+
+        for expected in expected_lines:
+            assert any(expected in line for line in content.splitlines()), f"Missing line: {expected}"
```
amaslenn, line 254: Should it be `all`, not `any`?

src/cloudai/_core/test_template.py
```diff
@@ -90,50 +90,6 @@ def gen_exec_command(self, tr: TestRun) -> str:
             )
         return self.command_gen_strategy.gen_exec_command(tr)
 
-    def gen_srun_command(self, tr: TestRun) -> str:
```
amaslenn, line 93: Can we also remove `gen_exec_command`? Asking if that make sense to you, let's not expand the scope of this PR.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -145,16 +145,40 @@ def job_name(self, job_name_prefix: str) -> str:
             job_name = f"{self.system.account}-{job_name_prefix}.{datetime.now().strftime('%Y%m%d_%H%M%S')}"
         return job_name
 
+    def _prepare_test_run(self, base_output_dir: Path, tr: TestRun) -> Tuple[Path, "SlurmCommandGenStrategy"]:
```
amaslenn, line 148: This function make sense to me. Just let's move import to the top of the file, it can be relative.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
```diff
@@ -131,132 +131,127 @@ def test_time_limit(time_limit: Optional[str], strategy_fixture: SlurmCommandGen
...
+def make_test_run(slurm_system: SlurmSystem, name: str, output_dir: Path) -> TestRun:
+    test_def = NCCLTestDefinition(
+        name=name,
+        description=name,
+        test_template_name="nccl",
+        cmd_args=NCCLCmdArgs(),
+        extra_env_vars={"TEST_VAR": "VALUE"},
+    )
+    test_template = TestTemplate(slurm_system, "nccl")
+    from cloudai.workloads.nccl_test.slurm_command_gen_strategy import NcclTestSlurmCommandGenStrategy
```
amaslenn, line 143: Let's move the import.

src/cloudai/_core/test_template.py
TaekyungHeo: Removing gen_exec_command makes sense to me, especially because we have two diverging methods: gen_exec_command and gen_json. I don't want to have two methods with different purposes in the parent class. However, I am not sure whether it is straightforward to remove it. We need to explore and evaluate.

tests/slurm_command_gen_strategy/test_common_slurm_command_gen_strategy.py
amaslenn: Should that be simply `assert expected in content.splitlines(), ...` then?

TaekyungHeo: The semantics are slightly different. The proposed assertion expects an exact match, whereas the current impl allows the keyword to appear anywhere within a line.

If you're unsure whether the changes are valid, you can see that test_acceptance.py remains unchanged and still passes. Since it includes test hooks, this indicates that the changes do not break them.

amaslenn: Just hard to understand. Probably because of Friday :)

amaslenn (approved):

PR #458, opened 2025-04-04
Support NCCL_TOPO_FILE in all SlurmCommandGenStrategy classes

TaekyungHeo:
## Summary
Support NCCL_TOPO_FILE in all SlurmCommandGenStrategy classes. Currently, NCCL_TOPO_FILE is only supported in NCCL tests. However, this should not be the case, and it should be supported by any Slurm command generation strategy. This PR moves the logic to the parent class, allowing other SlurmCommandGenStrategy implementations to mount NCCL_TOPO_FILE by default. This is one of the LLMB requirements.

## Test Plan
CI passes
* Passes `tests/slurm_command_gen_strategy/test_nccl_slurm_command_gen_strategy.py`. It indicates that this change does not break NCCL_TOPO_FILE but only expands its support for general slurm command gen strategy.

### DIFF
--- a/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
+++ b/src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
@@ -51,12 +51,7 @@ class SlurmCommandGenStrategy(CommandGenStrategy):
 
     @final
     def container_mounts(self, tr: TestRun) -> list[str]:
-        """
-        Return the container mounts for the test run.
-
-        Function returns CommandGenStrategy specific container mounts as well as default ones
-        that should always be used.
-        """
+        """Return the container mounts for the test run."""
         tdef = tr.test.test_definition
 
         repo_mounts = []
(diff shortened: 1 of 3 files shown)


amaslenn (commented):
1. Why do we need it for all workloads? Name of the variable says it is for NCCL only.
2. Please expand unit tests (if we actually need this support).

TaekyungHeo:
Let's start by addressing the first question:

> Why do we need it for all workloads? The name of the variable suggests it is for NCCL only.

NCCL is a library that can be used in other workloads such as NeMoLauncher, NeMoRun, and Chakra Replay. However, it's also true that NCCL is not used in UCC tests. 

Rather than introducing a new class specifically for workloads that use the NCCL library, I felt it was reasonable and straightforward to mount it globally in the base class—since UCC tests won’t use or provide that option anyway.

amaslenn:
> Once we are on the same page for the first question, I will expand unit tests.

We are on the same page, thanks for explanation!

TaekyungHeo:
I will expand unit tests. FYI, existing unit tests also cover NCCL_TOPO_FILE but only for NCCL tests

amaslenn:
> I will expand unit tests. FYI, existing unit tests also cover NCCL_TOPO_FILE but only for NCCL tests

Great, let's move these tests to generic level.

TaekyungHeo:
@amaslenn Please find the updated code.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -51,27 +51,32 @@ def _container_mounts(self, tr: TestRun) -> list[str]:
...
             *repo_mounts,
             *self._container_mounts(tr),
             f"{self.system.install_path.absolute()}:/cloudai_install",
         ]
 
+        merged_env = self.system.global_env_vars.copy()
+        merged_env.update(tr.test.extra_env_vars)
+        if "NCCL_TOPO_FILE" in merged_env:
+            nccl_topo_file = merged_env["NCCL_TOPO_FILE"]
+            if isinstance(nccl_topo_file, str):
```
amaslenn, line 74: Do we actually need to check for `isinstance` here? Our tests don't verify it + if it comes directly from TOML, nothing else can be there, we limit env vars to str:str mapping.

BTW, if `src == dst`, I believe we can use shorten notation with just `src`. (not necessary in this PR, but if you have BW to check it, would be great).

TaekyungHeo: It is because an environment variable can be a DSEable argument, and it is checked against a str instance to ensure it is not a list.
* Reference: https://github.com/NVIDIA/cloudai/blob/821cad529633c2ce035b2c6b258527e555659b10/src/cloudai/_core/test.py#L75

PR #480, opened 2025-04-14
Set sequence length explicitly in NeMoRun config and data model

TaekyungHeo:
## Summary
Set sequence length explicitly in NeMoRun config and data model

CloudAIX PR: https://github.com/Mellanox/cloudaix/pull/210

## Test Plan
CI passes


src/cloudai/workloads/nemo_run/nemo_run.py
```diff
@@ -40,6 +40,7 @@ class Data(BaseModel):
 
     model_config = ConfigDict(extra="allow")
 
+    seq_length: int = 8192
```
amaslenn, line 43: Should it be DSE-able?

TaekyungHeo: It makes sense.

TaekyungHeo: Updated.

amaslenn (approved):

srinivas212 (approved):

PR #488, opened 2025-04-16
Updates for slurm metadata

amaslenn:
## Summary
1. `ompi_version` --> `mpi_version`
2. + `mpi_type`
3. - `nnodes` (duplicated of `num_nodes`).

## Test Plan
1. CI
2. Real run on CW, metadata looks valid.

## Additional Notes
—

### DIFF
--- a/src/cloudai/systems/slurm/metadata.py
+++ b/src/cloudai/systems/slurm/metadata.py
@@ -31,7 +31,8 @@ class MetadataSystem(BaseModel):
 class MetadataMPI(BaseModel):
     """Represents the MPI metadata."""
 
-    ompi_version: str
+    mpi_type: str
+    mpi_version: str
     hpcx_version: str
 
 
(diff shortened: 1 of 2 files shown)


TaekyungHeo (approved):

src/cloudai/systems/slurm/metadata.py
```diff
@@ -68,7 +69,6 @@ class MetadataSlurm(BaseModel):
     num_nodes: str
     ntasks_per_node: str
     ntasks: str
-    nnodes: str
     job_id: str
```
TaekyungHeo, line 72: I know this isn't directly related to the PR, but we may need to revisit SlurmSystemMetadata. It currently includes both system and job metadata, not just system-specific metadata.

amaslenn: Yes, this is pretty much a mess today, we also need to separate test data (args, env, actual CLI used, etc) from job data (job id, etc).

PR #496, opened 2025-04-18
Nemo2.0 Perf Features (next set)

srivatsankrishnan:
## Summary
This Umbrella PR for several features for Nemo-Perf efforts.

* Contstrain Check
In Nemo2.0, the `dp` dimension in internally calculated and sometimes can be 0 for cases where `cp`, `pp`, `tp` exceeds total number of GPUs. 

* Vboost with Sudo
The original implementation of Vboost wasn't working since we were passing the container mounts. THe container has root access inside and sudo wasn't working. These commands must run on the node without any container mounts, image, and should run on the node directly. So need for the extra frills that we have for other `srun`.

* FLOPsMeasurementCallback
FLOPsMeasurement Callback prints the flops count to the stdout. This was requested by malay and used in his training scripts. 

* Cross Entropy Fusion
Enables the cross entropy fusion. Common feature across H100/B200/GB200 for all models.

* Auto-set the num_train_samples
When we are using mock dataset, the training doesn't get enough samples and it causes NCCL error causing it to hang. We had earlier set this in the test toml. But since we support the DSE to specify gbs and max_steps as DSE'ble parameters, we need to calculate all the combinations and weed out the invalid ones. So now, we automatically set this based on teh value of gbs and max_steps just before the command_generation.

Note: This parameter unlike others cannot be a List. 

* Adding NUMA control
This is used primarily for performance recipe. We are not sure is this will be widely used across all the use cases. For this configurability, we will introduce an environment variable called `ENABLE_NUMA_CONTROL` to enable the generation of the NUMA control.

### Constraint Check Feature for DP-->0
CI/CD
Dry-Run
```
  num_nodes = 1

    [cmd_args.trainer.strategy]
    tensor_model_parallel_size = [16]
    pipeline_model_parallel_size = [1]
    context_parallel_size = [2]

```
```
  No dependencies
[INFO] Creating SlurmRunner
[INFO] Constraint check failed. Skipping step.
[INFO] Step 1: Observation: [-1.0], Reward: -1.0
```
### Vboost requiring Sudo
CI/CD
Run
```
cloudai run --system-config ../cloudaix/conf/common/system/xxxx.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml 
[INFO] System Name: xxxx
[INFO] Scheduler: slurm

  No dependencies
[INFO] Creating SlurmRunner
[INFO] Running step 1 with action {xxxxxxxxxx}
[INFO] Submitted slurm job: 2462923
```

### FLopsMeasurementCallback
The generatd stdout logs will have a new column for `tflops_per_sec_per_gpu`.
```
Training epoch 0, iteration 54/99 | lr: xxxx-06 | global_batch_size: xxx | global_step: xx | peak_memory_usage: xxx | memory_allocated: xxx | reduced_train_loss: 11.97 | train_step_timing in s: xxx | consumed_samples: 7040 | tflops_per_sec_per_gpu: xxx
Training epoch 0, iteration 55/99 | lr: xxx-xx | global_batch_size: xxx | global_step: xxx | peak_memory_usage: xxx | memory_allocated: xxx | reduced_train_loss: xxx | train_step_timing in s: xxx | consumed_samples: xxx | tflops_per_sec_per_gpu: xxx
```

### NUMA Control Test plan
* CI/CD
* Dry-run
```cloudai dry-run --
system-config ../cloudaix/conf/common/system/xxxx.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nem
o_run_llama3_8b.toml 
[INFO] System Name: xxx
[INFO] Scheduler: slurm

  No dependencies
[INFO] Creating SlurmRunner
[INFO] Running step 1 with action {'extra_env_vars.ENABLE_NUMA_CONTROL': '1', xxxx}
[INFO] Submitted slurm job: 0
[INFO] Job completed: dse_nemo_run_llama3_8b_1 (iteration 1 of 1)
[INFO] Step 1: Observation: [-1.0], Reward: -1.0
```
Generated sbatch script
```
export ENABLE_NUMA_CONTROL=1
...
...
srun --mpi=pmix numactl --cpunodebind=$((SLURM_LOCALID/4)) --membind=$((SLURM_LOCALID/4))
```
## Additional Notes


src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -103,11 +103,31 @@ def _validate_recipe_name(self, recipe_name: str) -> str:
 
         return recipe_name
 
+    def update_num_train_samples(self, tr: TestRun) -> None:
```
amaslenn, line 106: Can we make it a part of Test Definition model for NemoRun? Like post validation. Benefit: this will work for all systems at the same time, no changes for cmd gen will be required.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -328,6 +329,17 @@ def _enable_vboost_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
             ]
         )
 
+    def _enable_numa_control_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
+        return " ".join(
+            [
+                "srun",
+                "--mpi=pmix",
```
amaslenn, line 336: `--mpi=...` should be taken from `system.mpi`.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -328,6 +329,17 @@ def _enable_vboost_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
...
+                "numactl",
+                "--cpunodebind=$((SLURM_LOCALID/4))",
```
amaslenn, line 338: Why 4? Is it always like that?

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -319,7 +319,8 @@ def _metadata_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
     def _enable_vboost_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
         return " ".join(
             [
-                *self.gen_srun_prefix(slurm_args, tr),
+                "srun",
```
amaslenn, line 322: Here we loose `--export=ALL` and other flags. Likely not a problem here, but would be better to have use `gen_srun_prefix` and append `--ntasks=1` after it.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -355,6 +367,8 @@ def _write_sbatch_script(
 
         if env_vars.get("ENABLE_VBOOST") == "1":
             batch_script_content.extend([self._enable_vboost_cmd(slurm_args, tr), ""])
+        if env_vars.get("ENABLE_NUMA_CONTROL") == "1":
+            batch_script_content.extend([self._enable_numa_control_cmd(slurm_args, tr), ""])
```
TaekyungHeo, line 371: Let's confirm whether we agree to adopt this approach for all command generation. Since we are modifying `src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py`, this implies that the approach will be applied to all Slurm command generation (including VBOOST and NUMA_CONTROL with environment variables). 

An alternative solution would be to inherit from this class and enable the behavior specifically in the NeMoRun Slurm command generation strategy.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: You don't need the other flags nor additional `--export=ALL` for this command. You don't need container mounts and image all. The previous bug with `sudo` not working with vboost is due to using the all the cloudAI prefix to it. Thats how this command is supposed to be. No additional frills required.

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
srivatsankrishnan: What do you mean all system. Its supposed only for Nemo. Not generic. Without this fix, training won't finish

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: Yes

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
amaslenn: Systems are: slurm, standalone, lsf. This code change is introduced inside cmd gen strategy which is per system. So we will have to duplicate it per system. But if we move it to the model level, that won't be required.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
amaslenn: I see, thanks. Please add unit tests for these functions so we solidify this behavior.

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
srivatsankrishnan: Not urgent for this PR. Nemo is currently only suppored for slurm only.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
TaekyungHeo: **Apr 22nd, 2025.** 
Briefly discussed. We are on the same page. Not an issue.

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
amaslenn: What exactly is "urgent" here? This is a new feature and we need to decide how to best fit it into our design. IMO my proposal gives us benefits of not implementing this again and again for other systems, plus the cost of change is copy-paste to another file. RunAI/K8s will need NemoRun at some point.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srivatsankrishnan: There is already a test for that. Thats good coverage for cmd generation
https://github.com/NVIDIA/cloudai/blob/main/tests/ref_data/nemo-run-vboost.sbatch

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
srinivas212: Is the idea ENABLE_NUMA_CONTROL will only be turned on for GB200? Not for H100/B100?

srivatsankrishnan: For H100/B200/GB200. This is a common [utility] (https://github.com/NVIDIA/NeMo/blob/9a0592c3a9c67100155d5a5cc8bfb942a5eef544/scripts/performance/utils.py#L82) used across all  and can be controlled by the platform flag. But same as what we can do as well. We have an environment variable and will let that define it.

You mean assuming 2 NUMA nodes not cores. Something like this

assuming 8 takes running on 2 Numa nodes. SLURM_LOCALID[0-3]--> 1st NUMA node
SLURM_LOCALID[4-7] --> 2nd NUMA node

 But at present this is exactly how they define it.

srinivas212: Makes sense. Yes, I meant two Numa nodes.

I'm not sure if it would be same for GB200, right? Because node is 4 gpus? 

This is mainly for my understanding. We can resolve this comment.

src/cloudai/systems/slurm/strategy/slurm_command_gen_strategy.py
```diff
@@ -328,6 +329,17 @@ def _enable_vboost_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
             ]
         )
 
+    def _enable_numa_control_cmd(self, slurm_args: dict[str, Any], tr: TestRun) -> str:
+        return " ".join(
+            [
+                "srun",
+                f"--mpi={self.system.mpi}",
```
amaslenn, line 336: Is MPI needed here?

srinivas212: Let's break this down into few parts - 
1. Do we need mpi here? And why? 
2. If not required, let's remove this in #500 since this PR has already gone thru significant testing. 

In future, we can try to resolve this in one shot. This approach is mainly to make progress on this PR.

srivatsankrishnan: We already addressed this command here:
https://github.com/NVIDIA/cloudai/pull/496#discussion_r2059324983

srivatsankrishnan: Relevant pointer: https://github.com/NVIDIA/NeMo/blob/9a0592c3a9c67100155d5a5cc8bfb942a5eef544/scripts/performance/utils.py#L82

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
amaslenn: Discussed offline, to be addressed later

amaslenn (approved):

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
srivatsankrishnan: Andrei to share the exact code snippet for `post-validation` using pydantic. Probably needs additional testing as well.

PR #498, opened 2025-04-21
[WIP] Enable nsys trace collection for NeMo2.0

TaekyungHeo:
## Summary
Enable nsys trace collection for NeMo2.0

## Test Plan
1. CI passes
2. Ran on EOS.


src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -501,6 +501,11 @@ def cloudai_llama3_70b_recipe() -> run.Partial:
                     overlap_param_gather=True,
                 ),
                 timing_callback(),
+                run.Config(
+                    NsysCallback,
+                    start_step=20,
+                    end_step=30,
+                ),
```
TaekyungHeo, line 508: 1. Check if it collects steps. between 20 and 30
2. Check if controlling start / end is supported.

TaekyungHeo: 2. https://nvidia.slack.com/archives/C083Y1XED43/p1745351434360739

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -446,6 +446,29 @@ def cloudai_llama3_8b_recipe() -> run.Partial:
 # LLAMA3 70B Recipe
 @run.cli.factory(target=llm.pretrain)
 def cloudai_llama3_70b_recipe() -> run.Partial:
+    nsys_enabled = os.getenv("NSYS", "0") == "1"
```
amaslenn, line 449: For task and recipe we use `CLOUDAI_...` env vars format, do you think it make sense following this approach for nsys as well? IMO it make sense.

amaslenn (approved):

srinivas212:
@TaekyungHeo - can we enable this for all models? It is going to be confusing for users to get this only for Llama3 70B. I understand the LLMB requirement - I'm asking a more general question. We can do this in a separate PR but needs to be considered.

PR #500, opened 2025-04-22
Nemo2.0 Perf Recipes (Set 2)

srivatsankrishnan:
## Summary
This is continuation of the [PR](https://github.com/NVIDIA/cloudai/pull/496).  Seperating it out so not all features are blocking the cut of new release branch. While the previous PR also has lots of features for next release and other projects, each of the features were tested on system.

This PR has the following feature support
* For H100/B200/GB200 (BF-16) only.
This flag needs to be enabled. So adding this in the test definition with the correct class hierarchy
1recipe.optim.config.use_precision_aware_optimizer = True (only for bf16)`

### use_precision_aware_optimizer
* CI/CD
* Dry-Run
* Real System
```
cloudai run --system-config ../cloudaix/conf/common/system/xxxx.toml 
--tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml 
[INFO] System Name: xxxx
[INFO] Scheduler: slurm

  Test Name: dse_nemo_run_perf_llama3_8b_bf16
  Description: dse_nemo_run_perf_llama3_8b_bf16
  No dependencies
[INFO] Creating SlurmRunner
[INFO] Running step 1 with action {}
[INFO] Submitted slurm job: xxx
```
Output

```
│ optim            │ MegatronOptimizerModule(                                  │
│                  │   config=OptimizerConfig(                                 │
│                  │     optimizer='xxx',                                     │
│                  │     lr=xxx,                                            │
│                  │     weight_decay=xxx                                     │
│                  │     fp16=xxx,                                           │
│                  │     bf16=xxx,                                            │
│                  │     params_dtype=torch.bfloat16,                          │
│                  │     use_precision_aware_optimizer=True,                   │
│                  │     adam_beta1=xxx,                                       │
│                  │     adam_beta2=xxx,                                      │
│                  │     adam_eps=1exxxx,                                       │
│                  │     use_distributed_optimizer=xx,                       │
│                  │     clip_grad=1xx),                                       │
│                  │   lr_scheduler=CosineAnnealingScheduler(                  │
│                  │     warmup_steps=2000,                                    │
│                  │     constant_steps=0,                                     │
│                  │     min_lr=2.9999999999999997e-05))
...
...
...
raining epoch 0, iteration 0/99 | lr: xxx-07 | global_batch_size: xxx | global_step: 0 | peak_memory_usage: xxx | memory_allocated: xxx | reduced_train_loss: xxx | train_step_timing in s: xxx
Training epoch 0, iteration 1/99 | lr: xxx-07 | global_batch_size: 128 | global_step: 1 | peak_memory_usage: xxx | memory_allocated: xxx | reduced_train_loss: 11.97 | train_step_timing in s: xxx |                  
```

## Additional Notes


amaslenn:
As part of this PR please make `Data.num_train_samples` DSE-able, otherwise it breaks https://github.com/Mellanox/cloudaix/blob/main/conf/staging/nemo_perf/test/dse_nemo_run_llama3_70b_fp8.toml#L31

srivatsankrishnan:
> As part of this PR please make `Data.num_train_samples` DSE-able, otherwise it breaks https://github.com/Mellanox/cloudaix/blob/main/conf/staging/nemo_perf/test/dse_nemo_run_llama3_70b_fp8.toml#L31

I will open this PR and you can review @amaslenn . Else you will be wasting your time on reviewing WIP or fixes that are already in pipeline. Currently this is draft and the focus is on getting feature parity with Nemo2.0 standalone mostly on the B200.

This variable cannot be DSE'ble. I will explain more in the call next time. This has to be internally set and the toml file you linked is specifically for nemo-perf effort which this PR adds features for.

srivatsankrishnan:
> > I will open this PR and you can review @amaslenn . Else you will be wasting your time on reviewing WIP or fixes that are already in pipeline. Currently this is draft and the focus is on getting feature parity with Nemo2.0 standalone mostly on the B200.
> 
> The thing is that current CloudAI is incompatible with current CloudAIx. So if anyone would try to bump dependency version, `num_train_samples` will be a problem. Should be either removed from config, make scalar, or fixed on CloudAI side.
> 
> I'm not reviewing this PR, this misconfiguration was introduced in #496 and this PR is a follow up, so I'm adding this comment here.

Sure, I ams saying this is a known issue with nemo-perf (the reason why we had this variable in toml file was for a fix for Daria. there are many moving pieces here) .and once we align on the performance value this iwll be updated. This value will not be exposed to user in toml file. This shouldn't be DSE'ble.

srivatsankrishnan:
Outstanding issue for this PR -- How to apply post-validation pydantic to this `num_train_samples`.  Currently the way it is this pydantic approach won't work for DSE (where the values that `num_train_samples` depends upon a list `gbs`). For scalar values 

Andrei;s suggestions
Here is a though on how we can approach that:
* post-validation method ignores all list values, so we do not change anything if any of two fields we depend on is a list.

[Sri]: But we do want to update the value of `num_train_samples` dynamically once the particular value is selected from a list. This happen way after the model initialization. It happens in gym environment and eventually get scalar value before command generation. 

* post-validation method runs actual verification and modifies field value if all fields we depend on are scalars. We override any existing value user provided with what should be set.
[Sri]: `gbs` is DSE'b;e knobs. But the `num_train_samples` needs to be updated even for DSE'ble knobs. This assumption might not be true for us.

* To enable it, we need to make sure that when we create apply a new set of args, model is re-created and post-validation methods are called.
[Sri]: I am not following but from the looks of it seems complex. Please give an example of what you mean by this.

amaslenn:
> * post-validation method runs actual verification and modifies field value if all fields we depend on are scalars. We override any existing value user provided with what should be set.
>   [Sri]: `gbs` is DSE'b;e knobs. But the `num_train_samples` needs to be updated even for DSE'ble knobs. This assumption might not be true for us.

The issue is the same with constraint checks: we do not run it for initial definitions with lists, only when we know that only scalar values are set.

> * To enable it, we need to make sure that when we create apply a new set of args, model is re-created and post-validation methods are called.
>   [Sri]: I am not following but from the looks of it seems complex. Please give an example of what you mean by this.

Here is an example: https://github.com/NVIDIA/cloudai/blob/64f6cd3c65134cb70f5fcd5267045b08cdff2d08/src/cloudai/_core/test_scenario.py#L148

amaslenn:
> Constrain check we are not doing post-validation pydantic right? Its currently defined as a property. I would actually consider this approach. It will help with the original goal. But this is not post-validation. Am I missing something?

Right, cause the semantics is different here. We construct the model, but some param sets we do not want to run. Model itself is fine, but run logic wants to exclude some combinations.

> * Move the logic from slurm command generation to test definition.
> * Call this property to update the `num_train_samples`.

1. The main intent here is to avoid code duplication, that is why cmd gen is considered bad here (we will have to do the same for other cmd gen strategies).
2. Semantic is different: based on some fields we want to set another field value. For simplicity, let's imagine we don't have DSE. Create model -> some fields are auto-computed. With DSE I still suggest that we ignore lists and do it only if all fields we depend on are scalars.
    - BTW, are we going to allow user setting such fields explicitly? What if the value is invalid according to our check?

srivatsankrishnan:
> Right, cause the semantics is different here. We construct the model, but some param sets we do not want to run. Model itself is fine, but run logic wants to exclude some combinations.

For DSE, if we have property in the test definition. Does that work? This will take care of both DSE and non-DSE case. We will just call this property where it is required where we are always guaranteed that it will be scalar. This is my original proposal as well. It satisfy the main goal (move this from slurm command generation --> test definition)


> 2. Semantic is different: based on some fields we want to set another field value. For simplicity, let's imagine we don't have DSE. Create model -> some fields are auto-computed. With DSE I still suggest that we ignore lists and do it only if all fields we depend on are scalars.

With DSE I still suggest that we ignore lists and do it only if all fields we depend on are scalars.
This will break for DSE because we still need to calcualte and set this value.

For Non-DSE case

gbs = 2
max_steps = 100

`num_train_samples` = gbs * max_steps (this needs to be set to 200). Else we training will fail.


For DSE
gbs = [2, 4, 8]
max_steps = 100

If we ignore this calculation? But even for DSE, this needs to be set. Else training will fail. Thats why I feel post-validation might not be the right approach for this. If our goal is not have code-duplication (we don't today since we only support Slurm for this model). Lets assume if we have multiple backend, then I still feel doing someting like property would be a better solution.

`num_train_samples` = gbs * max_steps (this needs to be set to 200). Else we training will fail.
>    * BTW, are we going to allow user setting such fields explicitly? What if the value is invalid according to our check?

We don't want to expose this value. It should be internally calculated.

amaslenn:
> For DSE, if we have property in the test definition. Does that work? This will take care of both DSE and non-DSE case. We will just call this property where it is required where we are always guaranteed that it will be scalar. This is my original proposal as well. It satisfy the main goal (move this from slurm command generation --> test definition)

Yes, that satisfies the requirements, but semantically is confusing + see below.
 
> This will break for DSE because we still need to calcualte and set this value.

1. What will it break? Fields can have `None` values by default.
1. When we construct DSE (`.model_validate()`) we go bypass way in post-validation, cause `gbs` is a list.
2. Why we construct singular run (again `.model_validate()`, see my PR), we actually cet calculated value.
3. Semantically this makes more sense because we are talking about model construction. Yes, we can use property, but calling this property will be on users' side. What it is is forgotten for a new system. Using post-validation doesn't have this issue cause it is done automatically.


> We don't want to expose this value. It should be internally calculated.

Good, then we need to develop a way to warn/error users if the try to set in in TOML or during object creation.

srivatsankrishnan:
> For DSE gbs = [2, 4, 8] max_steps = 100
> 
> If we ignore this calculation? But even for DSE, this needs to be set. Else training will fail. Thats why I feel post-validation might not be the right approach for this. If our goal is not have code-duplication (we don't today since we only support Slurm for this model). Lets assume if we have multiple backend, then I still feel doing someting like property would be a better solution.
> 
> `num_train_samples` = gbs * max_steps (this needs to be set to 200 for gbs=2). Else we training will fail. `num_train_samples` = gbs * max_steps (this needs to be set to 400 for gbs=4). Else we training will fail. `num_train_samples` = gbs * max_steps (this needs to be set to 800 for gbs=8). Else we training will fail.

Can you help me understand with the above example how you think we can apply post-validation in pydantic while also ensuring semantically it is right? I am trying to understand how we can ignore when `gbs` it is specified as a list? When you say bypass post-validation, how will we set this value? If we don't set the `num_train_samples` value, the training will fail. 

> Why we construct singular run (again .model_validate(), see my PR), we actually cet calculated value.
In your PR, do you also encounter one of the values as list? 

Please use the exact example I have used here. It will help.

TaekyungHeo:
@amaslenn , @srivatsankrishnan , Let's have a call. I will schedule one.

srivatsankrishnan:
> @amaslenn , @srivatsankrishnan , Let's have a call. I will schedule one.

Thanks @TaekyungHeo for facilitating the discusison call. As [discussed](https://docs.google.com/document/d/18IETf4oCmo0IUhx69sBgR_LBWlLK-qv6au8LrTErvCY/edit?tab=t.0#heading=h.txsjoasqntek), the pydnatic approach needs more modifications and once Andrei's [PR](https://github.com/NVIDIA/cloudai/pull/513) and is tested, we will apply the post-validation pydantic feature.

TaekyungHeo:
I shared my comments with @srivatsankrishnan over a call. Let's wait for @amaslenn 's review.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -541,6 +704,74 @@ def cloudai_llama3_70b_recipe() -> run.Partial:
...
     )
     recipe.trainer.strategy.cross_entropy_fusion_impl = "te"
+    gpu_type = os.getenv("CLOUDAI_GPU_TYPE")
+    compute_dtype = os.getenv("CLOUDAI_GPU_DTYPE")
+    enable_cuda_graphs = os.getenv("CLOUDAI_ENABLE_CUDA_GRAPHS", "0") == "1"
+    if enable_cuda_graphs:
+        recipe.model.config.enable_cuda_graph = True
+        recipe.trainer.strategy.use_te_rng_tracker = True
+
+    if gpu_type == "h100" and compute_dtype == "bf16":
```
amaslenn, line 714: These 4 blocks are differ only in `tp_comm_overlap_cfg` field. Let's highlight that by extracting common part. We can also combine it with default:
```py
    callback = run.Config(
        MegatronCommOverlapCallback,
        tp_comm_overlap=False,
        overlap_param_gather_with_optimizer_step=True,
        defer_embedding_wgrad_compute=True,
        wgrad_deferral_limit=22,
    )
    if gpu_type == "h100" and compute_dtype == "bf16":
        callback.tp_comm_overlap = True
        callback.tp_comm_overlap_cfg = llama3_70b_bf16_h100_tp_overlap_config()
    elif ...
    
    recipe.trainer.callbacks.append(callback)
```

Will that work?

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -449,6 +599,10 @@ def cloudai_llama3_8b_recipe() -> run.Partial:
             model_name="llama3",
         )
     )
+    enable_cuda_graphs = os.getenv("CLOUDAI_ENABLE_CUDA_GRAPHS", "0") == "1"
```
amaslenn, line 602: Let's wrap it into a function that accepts recipe and updates it if needed. The same block is used below.

src/cloudai/workloads/nemo_run/nemo_run.py
```diff
@@ -38,6 +38,20 @@ class Plugin(BaseModel):
     grad_reduce_in_fp32: Optional[bool] = None
 
 
+class OptimConfig(BaseModel):
+    """Configuration for NeMoRun."""
+
+    model_config = ConfigDict(extra="allow")
+    use_precision_aware_optimizer: Optional[bool] = None
```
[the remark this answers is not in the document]
srivatsankrishnan: No not [tracked](https://docs.google.com/spreadsheets/d/1xQHVw5wHTkbCMiOnQheBC-pFU64xf0iDo1NNyPV6An4/edit?gid=0#gid=0) as DSE'ble parameter. yet.

tests/test_test_definitions.py
```diff
@@ -361,3 +361,45 @@ def test_load_is_mounted_and_src_exists(self, field: str, mount: str):
...
+    test_def = NeMoRunTestDefinition(
+        name="nemo_test",
+        description="desc",
+        test_template_name="nemo",
+        cmd_args=cmd_args,
+    )
+    assert test_def.num_train_samples == 1600
+
+
+def test_nemorun_num_train_samples_list():
```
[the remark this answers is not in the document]
srivatsankrishnan: Makes sense.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
srivatsankrishnan: Not sure and Ideally don't want to muck around with custom recipes. Sometimes callbacks are passed as lists of callbacks. So have to ensure the idx matches as well. Would need some more testing

1. functionality (easier)
2. Performance-level (if deviations founds, need more debugging)
3. optional: DSE (if 2 works, should be fine imo).

src/cloudai/workloads/nemo_run/nemo_run.py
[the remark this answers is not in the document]
srivatsankrishnan: Thought more about this. Lets make it DSE'ble. End-user shouldn't care about this and they can specify as static if needed.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
amaslenn: Don't fully get your point. How this
```py
if ...
    recipe.trainer.callbacks.append(...)
elif ...
    recipe.trainer.callbacks.append(...)
```
is different from
```py
callback = ...
if ...
    callback.field = ...
elif ...
    callback.field = ...
recipe.callbacks.append(callback)
```
in terms of idx?

My suggestion is about (1) reducing duplication and (2) improving readability by highlighting actual diff between if-else branches.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -278,7 +268,167 @@ def llama3_70b_bf16_tp_overlap_config() -> run.Config[TransformerLayerTPOverlapC
...
-def llama3_70b_fp8_tp_overlap_config() -> run.Config[TransformerLayerTPOverlapCfg]:
+def llama3_70b_bf16_b200_tp_overlap_config() -> run.Config[TransformerLayerTPOverlapCfg]:
+    return run.Config(
+        TransformerLayerTPOverlapCfg,
+        qkv_dgrad=run.Config(
+            BulkOverlapCfg,
+            cga_size=2,
+            method="bulk",
+            num_sm=16,
+            set_sm_margin=False,
```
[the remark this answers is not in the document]
srivatsankrishnan: I actually find this easy to read top to bottom and single line hard to read (left to right). Lets leave out subjective comments.

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -103,30 +103,12 @@ def _validate_recipe_name(self, recipe_name: str) -> str:
...
-
-        tdef.cmd_args.data.num_train_samples = global_batch_size * max_steps
-
     def generate_test_command(
         self, env_vars: Dict[str, Union[str, List[str]]], cmd_args: Dict[str, Union[str, List[str]]], tr: TestRun
     ) -> List[str]:
         tdef: NeMoRunTestDefinition = cast(NeMoRunTestDefinition, tr.test.test_definition)
 
-        self.update_num_train_samples(tr)
+        tdef.cmd_args.data.num_train_samples = tdef.num_train_samples
```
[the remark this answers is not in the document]
srivatsankrishnan: Yes. I renamed the property as `update_num_train_samples`. There is unit test for this

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
srivatsankrishnan: what do you mean? for enabling cuda_graphs?

amaslenn: Yes

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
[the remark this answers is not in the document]
amaslenn: It is quantitative: number of total line in file will be lower. But up to you.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
srivatsankrishnan: I don't follow the ask here. What exactly are you expecting here? Can you share a code snippet?

amaslenn:
```py
def set_enable_cuda_graphs_params(recipe):
    enable_cuda_graphs = os.getenv("CLOUDAI_ENABLE_CUDA_GRAPHS", "0") == "1"
    if enable_cuda_graphs:
        recipe.model.config.enable_cuda_graph = True
        recipe.trainer.strategy.use_te_rng_tracker = True

recipe = ...
set_enable_cuda_graphs_params(recipe)
```

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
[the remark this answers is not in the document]
amaslenn: Does it make sense to convert it into post-validation in this PR or you will do a follow up PR for that?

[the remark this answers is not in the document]
srivatsankrishnan: I will follow it up in the next PR.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -613,7 +828,35 @@ def cloudai_llama3_405b_recipe() -> run.Partial:
             model_name="llama3",
         )
     )
+
+    set_enable_cuda_graphs_params(recipe)
     recipe.trainer.strategy.cross_entropy_fusion_impl = "te"
+
+    enable_fsdp = os.getenv("CLOUDAI_ENABLE_FSDP", "0") == "1"
```
amaslenn, line 835: Do we have a list of all such env variables? Should we track it? Not a request for this PR.

amaslenn (approved):

TaekyungHeo (approved):

PR #505, opened 2025-04-29
Remove requirements files to rely on pyproject for dependencies

amaslenn:
## Summary
Remove requirements files to rely on pyproject for dependencies. This removes duplication of dependencies and allow modern tools (like `uv`) to be used.

## Test Plan
1. CI
2. `pip install [-e] <local path>`, `pip install git+https://github.com/amaslenn/cloudai/@am/no-reqs`

## Additional Notes
—


srivatsankrishnan (commented):
Please document the exact steps now we have to follow to move away from the `requirements.txt` and how this help with `uv`.

What happens on the private repo side of things? Are we getting rid of that as well?

It can just be your `history` command from your terminal.

srivatsankrishnan:
> > Please document the exact steps now we have to follow to move away from the `requirements.txt` and how this help with `uv`.It can just be your `history` command from your terminal.
> 
> Not sure I understand this request. Why do you need my commands' history? Today we have to mention requirements in both requirements.txt and pyproject.toml. This PR removes one of these sources. `pip` will work as mentioned in README. `uv` can work in both `uv sync` and `uv pip` modes depends on how user wants it.

How will this command work if we remove the `requirements.txt` file? Can we document it please?
image

srivatsankrishnan:
> > > Please document the exact steps now we have to follow to move away from the `requirements.txt` and how this help with `uv`.It can just be your `history` command from your terminal.
> > 
> > 
> > Not sure I understand this request. Why do you need my commands' history? Today we have to mention requirements in both requirements.txt and pyproject.toml. This PR removes one of these sources. `pip` will work as mentioned in README. `uv` can work in both `uv sync` and `uv pip` modes depends on how user wants it.
> 
> How will this command work if we remove the `requirements.txt` file? Can we document it please? image

NVM. I saw the readme now.

srivatsankrishnan (approved):

PR #510, opened 2025-05-02
Move data repository USER GUIDE to CloudAIX

TaekyungHeo:
## Summary
Move data repository USER GUIDE to CloudAIX

https://github.com/Mellanox/cloudaix/pull/242

## Test Plan
CI passes


USER_GUIDE.md
amaslenn, line 1: Why removing from here? IMO it makes sense to have in both repos.

TaekyungHeo: Yes, it makes sense. Then let's leave it here. I will close this PR.

PR #513, opened 2025-05-02
Move sweeps logic to TestRun

amaslenn:
## Summary
Move sweeps logic to TestRun. Sweeps are needed in multiple parts of CloudAI:
1. Reporting, to know the size of action space and potential unrolling all the cases.
2. Runners, for unrolling all cases when it is needed, example is Single Sbatch Runner.

**CHANGE:**
Previous implementation returned all model fields for action space even if the were not defined as lists. New implementation returns only fields with lists as action space. This was done to make markings explicit:
1. `field1 = [1, 2], field2 = [1]` — both fields are DSE fields
2. `field1 = [1, 2], field2 = 1` — only `field1` is a DSE field.

## Test Plan
1. CI (updated)
3. Run `dse_nemo_run_llama3_8b.toml` on real system:
```bash
$ cloudai run --system-config system.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml                                                                                                                                     
[INFO] System Name: coreweave
[INFO] Scheduler: slurm

  No dependencies
[INFO] Creating SlurmRunner
[INFO] Running step 1 with action {'trainer.strategy.tensor_model_parallel_size': 4, 'trainer.strategy.pipeline_model_parallel_size': 1, 'trainer.strategy.context_parallel_size': 2, 'data.micro_batch_size': 1, 'data.global_batch_size': 128}
[INFO] Submitted slurm job: 2665996
[INFO] Job completed: dse_nemo_run_llama3_8b_1 (iteration 1 of 1)
[INFO] Step 1: Observation: [20.60966101694915], Reward: 0.04852093390461936
```

## Additional Notes
—


TaekyungHeo:
Thanks for the PR.
1. I agree with moving the logic to TestRun, as the action space is a trait of a test run, not of the gym itself.
2. I disagree with defining the action space using only lists. I will share more detail in the chat.
3. We need @srivatsankrishnan's review.

amaslenn:
@srivatsankrishnan please review this PR and my comments in the [internal doc](https://docs.google.com/document/d/1fqD_hBXmj0ikXX91iAVGvCLZk2tZdUKqK08TZdT7E0g/edit?pli=1&tab=t.0#heading=h.ca10pmxep9de)

srivatsankrishnan (changes requested):
As discussed in the document, action space part of the environment and therefore in gym.  `cmd_args` where the user specifies the parameter space is part of test run. So lets not move move the action space to test run.

The end goal you are trying to achieve is not have all parameters as list and only expose parameters that user specify as DSE'ble.  For that, we can still achieve that by not moving this logic from the gym environment to test run object.

Gym is ideally the interface where agents will also configure its action space. For learning based agent, the last layer in the policy (usually NN model) will define its action space.

amaslenn:
> For your end goal, we can still meet in the middle by modifying the action space to list only those that users specify as list and not move or define a new concept called `action_space` inside test run today.

Good we agree on that.

For the sweeps logic placement. I explained already that this functionality is needed by other parts of CloudAI: reporting and single sbatch runner for NCCL perf. DSE and Gym are part of CloudAI (not the other way around) and they are using CloudAI core (not the other way around). Sweeps is essential functionality for many use cases, not only DSE.

amaslenn:
@TaekyungHeo @srivatsankrishnan I've renamed `action_space --> param_space` and to for consistency `apply_action --> apply_params_set` on TR level. Please review.

TaekyungHeo (approved):

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
         return report(system, self).get_metric(self.test.test_definition.agent_metric)
 
+    @property
+    def param_space(self) -> dict[str, Any]:
+        action_space: dict[str, Any] = {}
+        cmd_args_dict = TestTemplateStrategy._flatten_dict(self.test.test_definition.cmd_args.model_dump())
+        extra_env_vars_dict = self.test.test_definition.extra_env_vars
+
+        action_space = {
+            **{f"{key}": value for key, value in cmd_args_dict.items() if isinstance(value, list)},
```
srivatsankrishnan, line 112: Aren't the keys already strings? Do you need the F-strings here? I think you can use the key directly. wdyt?

```
**{key: value for key, value in cmd_args_dict.items() if isinstance(value, list)},
```

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
+            **{f"extra_env_vars.{key}": value for key, value in extra_env_vars_dict.items() if isinstance(value, list)},
```
srivatsankrishnan, line 113: What happens when the keys can also have deep hierarchy? Can you please test this for corner cases?

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
+
+        action_space: dict[str, Any] = self.param_space
+
+        parameter_values: list[Any] = []
+        for _, values in action_space.items():
+            parameter_values.append(values)
+        action_combinations = list(itertools.product(*parameter_values))
+
+        keys = list(action_space.keys())
+        all_combinations = [dict(zip(keys, combination, strict=True)) for combination in action_combinations]
```
srivatsankrishnan, line 131: `strict=True` is python 3.10+. Will this run on older environments? Have we officially conveyed to everyone that we will be migrating to 3.10 now?

srivatsankrishnan, line 131: The cartesian product can be very huge. if it gets in the order of 10,000-1000,000, this can casue this method to be very slow. Can we test it with unit test for large space in that order.

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
+    def all_combinations(self) -> list[dict[str, Any]]:
+        if not self.test.test_definition.is_dse_job:
+            return []
+
+        action_space: dict[str, Any] = self.param_space
+
+        parameter_values: list[Any] = []
+        for _, values in action_space.items():
+            parameter_values.append(values)
+        action_combinations = list(itertools.product(*parameter_values))
```
srivatsankrishnan, line 128: If parameter_values is empty, itertools.product(*[]) yields one empty tuple — is this the desired behavior?

Should we check this condition to make the code robust?

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
+                tdef.extra_env_vars[key[len("extra_env_vars.") :]] = value
+            else:
+                attrs = key.split(".")
+                obj = tdef.cmd_args
+                for attr in attrs[:-1]:
+                    obj = getattr(obj, attr)
+                setattr(obj, attrs[-1], value)
+
+        new_tr = copy.deepcopy(self)
+        new_tr.test.test_definition = type(tdef)(**tdef.model_dump())  # re-create the model to enable validation
```
srivatsankrishnan, line 148: if model_copy or model_dump fails, the object may be left in a partially mutated state.

Can you add extra checks around this to ensure this operation always passes?

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
...
+            if key.startswith("extra_env_vars."):
+                tdef.extra_env_vars[key[len("extra_env_vars.") :]] = value
+            else:
+                attrs = key.split(".")
+                obj = tdef.cmd_args
+                for attr in attrs[:-1]:
+                    obj = getattr(obj, attr)
+                setattr(obj, attrs[-1], value)
+
+        new_tr = copy.deepcopy(self)
```
srivatsankrishnan, line 147: Since tr is basically encapsulats everything in CLoudAI. Are there any non-pickeable objects in there where this could fail? Maybe Taekyung can confirm about the datastore handlers. Are those Pickeable?

src/cloudai/_core/test_template_strategy.py
```diff
@@ -74,7 +74,8 @@ def construct_args(
 
         return construct_args(self.cmd_args)
 
-    def _flatten_dict(self, d: Dict[str, Any], parent_key: str = "", sep: str = ".") -> Dict[str, Any]:
+    @classmethod
```
srivatsankrishnan, line 77: Why is this added as class method? Can you explain what are the reason for doing this instead of instance method?

tests/test_cloudaigym.py
```diff
@@ -258,3 +122,79 @@ def test_tr_output_path(setup_env: tuple[TestRun, Runner]):
     env.step(action)
 
     assert env.test_run.output_path.name == "42"
+
```
srivatsankrishnan, line 125: Lots of duplicated code in the tests. Can you parameterize these test into a single test that takes 
* `cmd_args`
* `extra_cmd_args`
*  `expected_results`

Something like this
```
@pytest.mark.parametrize(
    "cmd_args, extra_env_args, expected_results",
    # Case 1: test_populate_action_space
    
```

```
def test_populate_action_space_parametrized(cmd_args, extra_env_args, expected_results):
    env = CloudAIGymEnv(test_run=MagicMock(), runner=MagicMock())
    action_space = {}
    combined_dict = {**cmd_args.model_dump(), **extra_env_args}
    env.populate_action_space("", combined_dict, action_space)
    for key, value in expected_results.items():
        assert action_space[key] == value
```

tests/test_cloudaigym.py
```diff
@@ -258,3 +122,79 @@ def test_tr_output_path(setup_env: tuple[TestRun, Runner]):
...
+    tr, _ = setup_env
+    nemorun.cmd_args.trainer = Trainer(max_steps=[1000])
+    tr.test.test_definition = nemorun
+    action_space = tr.param_space
+    action_space["trainer.max_steps"] = "invalid"
+
+    with pytest.raises(UserWarning) as excinfo:
+        tr.apply_params_set(action_space)
+
+    assert excinfo.type is UserWarning
```
srivatsankrishnan, line 198: Is it possbible to have a general method for assertion? We seem to have multiple assertions. It makes it hard to read. Do you think having a generic wrapper for assertions would make it better?

srinivas212:
As discussed in meeting, this is a good short-term solution for immediate deliverables. Not specifically in this PR but we should align towards one grid search implementation. This will require more code re-factoring. Let's plan.

srinivas212 (approved):

srivatsankrishnan (commented):
PIC has approved this design choice for getting this implemention out sooner. So please see the minor comments on the code in those light based on the urgency.

amaslenn:
> PIC has approved this design choice for getting this implemention out sooner. So please see the minor comments on the code in that light based on the urgency. The PR already has two approvals.

Since we have two approval, I'd like to merge it as is to make progress. @srivatsankrishnan please approve or any other way remove "request change" status, can't merge now.

But I found your comments very useful and will create a follow up PR with updates this week. Thanks for the effort, I think your suggestions are making our code better.

src/cloudai/_core/test_scenario.py
```diff
@@ -99,6 +102,52 @@ def get_metric_value(self, system: System) -> float:
 
         return report(system, self).get_metric(self.test.test_definition.agent_metric)
 
+    @property
+    def param_space(self) -> dict[str, Any]:
+        action_space: dict[str, Any] = {}
```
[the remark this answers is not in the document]
amaslenn: This can be removed, thanks for catching it!

src/cloudai/_core/test_scenario.py
amaslenn: `list(itertools.product(*[])) --> [()]`. I think it make sense to add such check, right. But in reality we unlikely get this issue, since in case of no DSE args, we won't call this function. Anyway, I'll add this check.

src/cloudai/_core/test_scenario.py
amaslenn: CloudAI can't be installed with <3.10 today. Here is an [announcement](https://nvidia.slack.com/archives/C06LC3PRHDZ/p1746173706818809) about it.

amaslenn: Let's test it when we get there. Otherwise the question is why wasn't it tested before.

src/cloudai/_core/test_scenario.py
amaslenn: Are you referring to `pickle` module? Is it related to this PR somehow? I think TR should be pickle-able, but need to try.

tests/test_cloudaigym.py
amaslenn: Not sure what you mean here, `assert` is the standard way to test values in `pytest`. If you have something in mind that is better, please share.

src/cloudai/_core/test_template_strategy.py
amaslenn: Sure, if you look at this method, it doesn't use `self` and can be used without instance creation. I'm using this method to flatten dicts, so it is pretty useful (and common) technic to make this method classmethod. An alternative solution would be to move this function into `utily` module, but that would introduce more unrelated to this PR changes. IMO moving it is still a good idea and I can do it in a follow up PR that will address other your suggestions. WDYT?

src/cloudai/_core/test_scenario.py
amaslenn: By this time dict is already flattened. `test_action_space` (and other) are based on NemoRun where we have nested configuration, so this is covered.

srivatsankrishnan (approved):
Might need to revisit few parts again in the follow-up PR.

amaslenn:
> Might need to revisit few parts again in the follow-up PR.

Agree. Could you please reply to my comments where applicable? Or I can open a follow up PR and you can add requests there. Both options work for me.

PR #516, opened 2025-05-07
fix can't import FLOPsMeasurementCallback issue for nemo:24.12.rc3

lilyw97:
## Summary
Acceptance Test for NeMo run is using **nemo:24.12.rc3**, which I suppose should not be changed so it can maintain same performance, But that version hasn't introduced **FLOPsMeasurementCallback**, so it will raise import error 

So change to import FLOPsMeasurementCallback when it's actually being used


lilyw97:
Hi @srivatsankrishnan , please help to check it

srivatsankrishnan:
> Hi @srivatsankrishnan , please help to check it

Hi Lily,

FlopsMeasurement callback is not supported in `24.12.rc3`. This callback is supported from 25.04. 

You can see the support here: https://github.com/NVIDIA/cloudai/blob/main/src/cloudai/workloads/nemo_run/cloudai_nemorun.py#L45

lilyw97:
> > Hi @srivatsankrishnan , please help to check it
> 
> Hi Lily,
> 
> FlopsMeasurement callback is not supported in `24.12.rc3`. This callback is supported from 25.04.
> 
> You can see the support here: https://github.com/NVIDIA/cloudai/blob/main/src/cloudai/workloads/nemo_run/cloudai_nemorun.py#L45

Thanks for the confirmation, could you review and approve this PR? thanks!

lilyw97:
Hi @srinivas212  / @srivatsankrishnan / @TaekyungHeo  / @amaslenn ,  could you help to review and merge this PR? the Nemo tests in dgx_acceptance_test uses `24.12.rc3.` which will raise error of `ModuleNotFoundError: No module named 'nemo.lightning.pytorch.callbacks.flops_callback'`

PR #526, opened 2025-05-14
Use lazy imports for slow modules

amaslenn:
## Summary
Use lazy imports for slow modules. Types are saved to still use IDE autocompletion and static type checker.

Some silly numbers from my laptop for `cloudai -h`:
||old version|new version|
|-|-|-|
|Cold cache (no `__pycache__`)|6sec+ (sometimes 14sec+)|1.8sec|
|Hot cache (w\ `__pycache__`|1.9sec|0.8s|

## Test Plan
CI

## Additional Notes
—

### DIFF
--- a/src/cloudai/util/lazy_imports.py
+++ b/src/cloudai/util/lazy_imports.py
@@ -46,7 +46,7 @@ class LazyImports:
         self._bokeh_pallettes: ModuleType | None = None
 
     @property
-    def np(self) -> "np":  # type: ignore[no-any-return]
+    def np(self) -> np:  # type: ignore[no-any-return]
         """Lazy import of numpy."""
         if self._np is None:
             import numpy as np
@@ -55,7 +55,7 @@ class LazyImports:
         return cast("np", self._np)
 
     @property
-    def pd(self) -> "pd":  # type: ignore[no-any-return]
+    def pd(self) -> pd:  # type: ignore[no-any-return]
         """Lazy import of pandas."""
         if self._pd is None:
             import pandas as pd
@@ -64,7 +64,7 @@ class LazyImports:
         return cast("pd", self._pd)
 
     @property
-    def k8s(self) -> "k8s":  # type: ignore[no-any-return]
+    def k8s(self) -> k8s:  # type: ignore[no-any-return]
         """Lazy import of kubernetes."""
         if self._k8s is None:
             import kubernetes as k8s
@@ -74,7 +74,7 @@ class LazyImports:
         return cast("k8s", self._k8s)
 
     @property
-    def bokeh(self) -> "bokeh":  # type: ignore[no-any-return]
+    def bokeh(self) -> bokeh:  # type: ignore[no-any-return]
         """Lazy import of bokeh."""
         if self._bokeh is None:
             import bokeh as bokeh
@@ -84,7 +84,7 @@ class LazyImports:
         return cast("bokeh", self._bokeh)
 
     @property
-    def bokeh_plotting(self) -> "bokeh_plotting":  # type: ignore[no-any-return]
+    def bokeh_plotting(self) -> bokeh_plotting:  # type: ignore[no-any-return]
         """Lazy import of bokeh.plotting."""
         if self._bokeh_plotting is None:
             import bokeh.plotting as bokeh_plotting
@@ -94,7 +94,7 @@ class LazyImports:
         return cast("bokeh_plotting", self._bokeh_plotting)
 
     @property
-    def bokeh_models(self) -> "bokeh_models":  # type: ignore[no-any-return]
+    def bokeh_models(self) -> bokeh_models:  # type: ignore[no-any-return]
         """Lazy import of bokeh.models."""
         if self._bokeh_models is None:
             import bokeh.models as bokeh_models
@@ -104,7 +104,7 @@ class LazyImports:
         return cast("bokeh_models", self._bokeh_models)
 
     @property
-    def bokeh_layouts(self) -> "bokeh_layouts":  # type: ignore[no-any-return]
+    def bokeh_layouts(self) -> bokeh_layouts:  # type: ignore[no-any-return]
         """Lazy import of bokeh.layouts."""
         if self._bokeh_layouts is None:
             import bokeh.layouts as bokeh_layouts
@@ -114,7 +114,7 @@ class LazyImports:
         return cast("bokeh_layouts", self._bokeh_layouts)
 
     @property
-    def bokeh_transform(self) -> "bokeh_transform":  # type: ignore[no-any-return]
+    def bokeh_transform(self) -> bokeh_transform:  # type: ignore[no-any-return]
         """Lazy import of bokeh.transform."""
         if self._bokeh_transform is None:
             import bokeh.transform as bokeh_transform
@@ -124,7 +124,7 @@ class LazyImports:
         return cast("bokeh_transform", self._bokeh_transform)
 
     @property
-    def bokeh_pallettes(self) -> "bokeh_pallettes":  # type: ignore[no-any-return]
+    def bokeh_pallettes(self) -> bokeh_pallettes:  # type: ignore[no-any-return]
         """Lazy import of bokeh.palettes."""
         if self._bokeh_pallettes is None:
             import bokeh.palettes as bokeh_pallettes


src/cloudai/util/lazy_imports.py
```diff
@@ -0,0 +1,137 @@
...
+        self._k8s: ModuleType | None = None
+        self._bokeh: ModuleType | None = None
+        self._bokeh_plotting: ModuleType | None = None
+        self._bokeh_models: ModuleType | None = None
+        self._bokeh_layouts: ModuleType | None = None
+        self._bokeh_transform: ModuleType | None = None
+        self._bokeh_pallettes: ModuleType | None = None
+
+    @property
+    def np(self) -> "np":  # type: ignore[no-any-return]
```
TaekyungHeo, line 49: Please confirm if it is the best practice to have `# type: ignore[no-any-return]` in this usage scenario. If yes, I am good.

TaekyungHeo (approved):
Let's have one more approval. It looks good

src/cloudai/util/lazy_imports.py
amaslenn: I'll double check, thanks for highlighting it.

TaekyungHeo:
@amaslenn By "from dev experience," do you mean as a developer importing "lazy"? My main concern is more about testing rather than developer experience. Importing packages will be handled automatically by Cursor, which looks fine to me. My concern is whether this change breaks any existing experiments / codes. I assume it does not, assuming our unit tests provide good coverage. If we want to invest more time, I would prefer to first set up nightly runs with good coverage first and then merge this to see if it breaks any of them.

src/cloudai/util/lazy_imports.py
amaslenn: The problem is in this: `Module cannot be used as a type (reportGeneralTypeIssues)`.

But without it we won't be able to keep type checks for code that uses these properties.

But I removed unneeded quotes for return types :)

amaslenn:
> @amaslenn By "from dev experience," do you mean as a developer importing "lazy"? My main concern is more about testing rather than developer experience. Importing packages will be handled automatically by Cursor, which looks fine to me. My concern is whether this change breaks any existing experiments / codes. I assume it does not, assuming our unit tests provide good coverage. If we want to invest more time, I would prefer to first set up nightly runs with good coverage first and then merge this to see if it breaks any of them.
> 
> If pulling and trying this PR helps, I will do. However given the same amount of resources, I would rather invest in improving coverage with nightly runs and then merge it.

For testing we are covered with `pyright` and unit tests.

Cursor and VSCode are fine, but PyCharm's autocompletion for these properties seems to be broken.

Sure, like I said, this PR is not urgent, take your time.

src/cloudai/util/lazy_imports.py
TaekyungHeo: Thanks for the update. **Another question**: is this considered best practice in the python community? I believe lazy import is widely adopted in large python projects. What was the main motivation for introducing this methodology? Do you have any examples?

amaslenn: I've noticed that startup time is not good. Also users reported that. So I started investigating it. CloudAI itself doesn't do much for initialization, but importing packages like `pandas` and `numpy` took a lot of time. Even for `cloudai -h` where we don't do anything with these packages.

The only way that I know (did some googling and AI search) is lazy imports. So I started per file, but as some point it became clear that approach is re-usable.

Once the heaviest modules were wrapped, I noticed that `k8s` and `bokeh` also consume significant amount of time so I wrapped them too.

Tools I've used:
1. `python -m benchmark_imports cloudai`
2. `python -m cProfile -o profile.out -m cloudai dry-run ...` + `python -c "import pstats; p = pstats.Stats('profile.out'); p.sort_stats('cumtime').print_stats(15)"`

Like you've said, lazy imports are widely adopted in python community. I guess we can call it "best practice", but in reality there are no other options. Or at least I don't know about them.

TaekyungHeo:
@amaslenn I have two questions:
1. What is the best way to measure the execution time of cloudaix.py? Do you have any recommendations?
2. I’ve noticed that cloudaix.py still seems slow, even after installing this branch. Have you had a chance to measure any performance improvements with cloudaix.py? I suspect the axplatform package might be a contributing factor.

amaslenn:
> 1. What is the best way to measure the execution time of cloudaix.py? Do you have any recommendations?

I'd recommend the same tools I've shared in comments above.

> 2. I’ve noticed that cloudaix.py still seems slow, even after installing this branch. Have you had a chance to measure any performance improvements with cloudaix.py? I suspect the axplatform package might be a contributing factor.

This is expected. I believe the slowest part is related to BO agent as it depends on multiple heavy libraries (about 800MB in size). Plus, `pandas` and `numpy` are still used directly, not via lazy imports. To summarize: unfortunately, CloudAIx won't benefit from this branch automatically, we need to change its code.

You can work on this if you have time: inherit lazy import class, extend it and use for all heavy objects in CloudAIx. Once we merge this PR, will quickly merge an update for CloudAIx too.

TaekyungHeo:
FYI - Regarding the PyCharm experience: it's helpful to understand the impact of this PR on developers, but I believe updating PyCharm is the responsibility of the PyCharm team.

amaslenn:
I've tested it on real HW.

DSE run:
```bash
$ cloudai run --system-config ../cw.toml --tests-dir conf/common/test --test-scenario conf/common/test_scenario/dse_nemo_run_llama3_8b.toml 
[INFO] System Name: coreweave
[INFO] Scheduler: slurm

  No dependencies
[INFO] Creating SlurmRunner
[INFO] Running step 1 with action {'docker_image_url': 'nvcr.io/nvidia/nemo:25.04.rc2', 'task': 'pretrain', 'recipe_name': 'cloudai_llama3_8b_recipe', 'num_layers': 32, 'trainer.max_steps': 100, 'trainer.val_check_interval': 100, 'trainer.num_nodes': 1, 'trainer.strategy.tensor_model_parallel_size': 4, 'trainer.strategy.pipeline_model_parallel_size': 1, 'trainer.strategy.context_parallel_size': 2, 'trainer.strategy.virtual_pipeline_model_parallel_size': None, 'trainer.plugins.fp8': 'hybrid', 'trainer.plugins.fp8_margin': 0, 'trainer.plugins.fp8_amax_history_len': 1024, 'trainer.plugins.fp8_amax_compute_algo': 'max', 'trainer.plugins.fp8_wgrad': None, 'trainer.plugins.fp8_params': True, 'trainer.plugins.grad_reduce_in_fp32': True, 'trainer.callbacks': None, 'log.ckpt': None, 'log.tensorboard': None, 'data.seq_length': 8192, 'data.micro_batch_size': 1, 'data.global_batch_size': 128, 'data.num_train_samples': 1000, 'extra_env_vars.ENABLE_VBOOST': '1', 'extra_env_vars.ENABLE_NUMA_CONTROL': '1', 'extra_env_vars.NCCL_P2P_NET_CHUNKSIZE': '2097152', 'extra_env_vars.TORCHX_MAX_RETRIES': '0', 'extra_env_vars.TRANSFORMERS_OFFLINE': '0', 'extra_env_vars.NCCL_NVLS_ENABLE': '0', 'extra_env_vars.NVTE_DP_AMAX_REDUCE_INTERVAL': '0', 'extra_env_vars.NVTE_ASYNC_AMAX_REDUCTION': '1', 'extra_env_vars.NVTE_FUSED_ATTN': '1', 'extra_env_vars.NVTE_FLASH_ATTN': '1', 'extra_env_vars.NEMO_LOG_MEMORY_USAGE': '1', 'extra_env_vars.CUDA_DEVICE_MAX_CONNECTIONS': '1', 'extra_env_vars.NVTE_FWD_LAYERNORM_SM_MARGIN': '16', 'extra_env_vars.NVTE_BWD_LAYERNORM_SM_MARGIN': '16'}
[INFO] Submitted slurm job: 2900043
[ERROR] Job 2900043 for test dse_nemo_run_llama3_8b_1 failed: Missing success indicators in results/nemo_run_llama3_8b_2025-05-15_04-15-01/dse_nemo_run_llama3_8b_1/0/1/stderr.txt: 'max_steps=', 'reached'. These keywords are expected to be present in stderr.txt when the NeMo training job completes successfully. Please review the full stderr output. Ensure that the NeMo training ran to completion and the logger output wasn't suppressed. If the issue persists, contact the system administrator.
[INFO] Terminating all jobs...
[INFO] Terminating job 2900043 for test dse_nemo_run_llama3_8b_1
[INFO] All jobs have been killed.
[INFO] Step 1: Observation: [21.350169491525424], Reward: 0.046838035660416305
```

Regular run:
```bash
$ cloudai run --system-config ../cw.toml --tests-dir conf/common/test --test-scenario conf/common/test_scen
ario/nccl_tests_inline.toml             
[INFO] System Name: coreweave
[INFO] Scheduler: slurm               
[INFO] Test Scenario Name: nccl-tests  
[INFO] Test Scenario: nccl-tests
                                                  
Section Name: nccl.all_reduce
  Test Name: nccl-all_reduce                                                                                                                                                                             
  Description: all_reduce                                                                                                                                                                                
  No dependencies                                                                                                                                                                                        
Section Name: nccl.all_gather                                                                                                                                                                            
  Test Name: nccl-all_gather                                                                                                                                                                             
  Description: all_gather                                                                                                                                                                                
  No dependencies                                                                                                                                                                                        
[INFO] Creating SlurmRunner                   
[INFO] Starting test: nccl.all_reduce        
[INFO] Running test: nccl.all_reduce
[INFO] Submitted slurm job: 2899980                                                                                                                                                                      
[INFO] Starting test: nccl.all_gather                                                                                                                                                                    
[INFO] Running test: nccl.all_gather                                                                
[INFO] Submitted slurm job: 2899981
[INFO] Job completed: nccl.all_reduce (iteration 1 of 1)        
[INFO] Job completed: nccl.all_gather (iteration 1 of 1)
[INFO] All test scenario results stored at: results/nccl-tests_2025-05-15_04-07-42
[INFO] Generated scenario report at results/nccl-tests_2025-05-15_04-07-42/nccl-tests.html                                                                                                               
[INFO] All jobs are complete.
```

IMO since we are at the beginning of our release, we can merge it and fix issues (if any) later.
@TaekyungHeo WDYT?

TaekyungHeo:
Sounds good. This is a good starting point, and you may or may not need additional approval. It's up to you.

PR #554, opened 2025-06-04
Fix invalid type for image when cache is disabled

amaslenn:
## Summary
Fix issue when `Path` was returned instead of `None` for the case when caching is not enabled on the system.

## Test Plan
1. CI (extended)
2. Local dry-run that showed the problem

## Additional Notes
—


amaslenn:
> @amaslenn I know it's merged, but it's good to test on a real system (e.g. NCCL test). I am not sure whether the consumer classes are using the returned values (the second field of the returned obj).

I was able to reproduce it with dry-run for NCCL. With the fix, I don't see this issue anymore.

TaekyungHeo:
@amaslenn  Sounds good.

PR #571, opened 2025-06-17
Cleanup configs in conf/common

amaslenn:
## Summary
The goal is to make this `conf/common/` actionable for QA: all configs should just work. To reduce the maintenance, I've simplified how these configs are defined.

Additionally, NCCL definition got `stepfactor` argument that is often used is configs via `extra_cmd_args`.

## Test Plan
1. CI.
2. See comments for real runs.

## Additional Notes
—


TaekyungHeo:
Basically, I agree with the motivation, but this needs discussion and @srinivas212's review.

TaekyungHeo (approved):

PR #589, opened 2025-06-27
Am/no job status strategy

amaslenn:
## Summary
Provide a concise summary of the changes introduced by this pull request. Detail the purpose and scope of the changes, referencing any relevant issues or discussions. Explain how these changes address the problem or improve the project.

## Test Plan
In this section, describe the testing you have performed to verify the changes. Include:
- A clear description of the testing environment.
- The steps you followed to test the new features or bug fixes.
- Any specific commands used during testing, along with their outputs.
- A description of the results and observations from your testing.
This information is crucial for reviewers to understand how the changes have been validated.

## Additional Notes
Include any other notes or comments about the pull request here. This can include challenges faced, future considerations, or context that reviewers might find helpful.

PR #609, opened 2025-07-22
Fix to Buggy Implemention of PR 589

srivatsankrishnan:
## Summary
DSE feature in CloudAI is completely broken due to buggy/faulty implementation of https://github.com/NVIDIA/cloudai/pull/589. 

DSE customers impacted by this
* Perf team blocked for finding good configuration
* AI Dynamo DSE feature also dysfunctional

### The Problem
DSE functionality was completely broken - only Step 1 would submit jobs to Slurm, while Steps 2-4 would appear to run but never actually submit jobs, resulting in immediate -1.0 rewards instead of proper parameter exploration.

Root Cause: PR #589 

Separated job status checking into two levels:

- Runner level: Slurm job status (COMPLETED, FAILED, etc.)

- Workload level: Application-specific success indicators (e.g., NeMo looking for "max_steps=" in stderr)

Started calling shutdown() for workload failures:

### Two Design Flaws Exposed
Flaw 1: Exception Swallowing
```
# In Runner.run() - masks the real problem
try:
    await self.runner.run()
except JobFailureError as exc:
    logging.debug(f"Runner failed: {exc}")  # Just logs it!
```
Flaw 2: Persistent Runner State

- shutting_down = True persists between DSE steps

- No reset mechanism for exploration scenarios

- Assumes single test scenarios, not multi-step exploration

## Test Plan
-CI/CD
- Real system

## Additional Notes

### DIFF
--- a/src/cloudai/_core/base_runner.py
+++ b/src/cloudai/_core/base_runner.py
@@ -259,6 +259,7 @@ class BaseRunner(ABC):
                                 f"Job {job.id} for test {job.test_run.name} failed: {job_status_result.error_message}"
                             )
                             logging.error(error_message)
+                            await self.handle_job_completion(job)
                             await self.shutdown()
                             raise JobFailureError(job.test_run.name, error_message, job_status_result.error_message)
                     else:
--- a/src/cloudai/configurator/cloudai_gym.py
+++ b/src/cloudai/configurator/cloudai_gym.py
@@ -22,6 +22,7 @@ from typing import Any, Dict, Optional, Tuple
 
 from cloudai.core import METRIC_ERROR, Registry, Runner, TestRun
 from cloudai.util.lazy_imports import lazy
+from cloudai._core.exceptions import JobFailureError
 
 from .base_gym import BaseGym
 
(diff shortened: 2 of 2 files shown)


malay-nagda:
I tested this PR on pretyche. I can now see sub-directories created for jobs scheduled for BO.

malay-nagda (approved):

amaslenn:
@srivatsankrishnan thanks for looking into this issue! Have some follow up questions:
1. The case your are describing is: first DSE case fails, others appear to run, but don't really run because of `.shutdown` flag is preserved? Please note, that this is not introduced by #589, but rather kept for backward compatibility. Such behavior was introduced by original implementation. We never saw first step to failure, so it wasn't an issue.
2. What do you mean by comment in:
    ```
    # In Runner.run() - masks the real problem
    try:
        await self.runner.run()
    except JobFailureError as exc:
        logging.debug(f"Runner failed: {exc}")  # Just logs it!
    ```
    This is again part of original design and wasn't ever changed: `Runner` treats scenario as a single unit. For defining actual test status we use two parts: runner's (always OK now) and `was_run_successful()` that is defined perf workflow (previously it was done by `JobStatusRetrievalStrategy`).
    So there is no "masking" of real problem.
3. What does this code actually checks?
    ```py
        if (
            self.runner.runner.test_scenario.test_runs
            and self.runner.runner.test_scenario.test_runs[0].output_path.exists()
        ):
            self.test_run = self.runner.runner.test_scenario.test_runs[0]
        else:
            self.test_run = copy.deepcopy(self.original_test_run)
            self.test_run.step = new_tr.step
            self.test_run.output_path = new_tr.output_path
    ```
    If this is a check for successful run based on per-workflow logic, that should be done via `test_run.test.test_definition.was_run_successful`. Is that what you wanted to check here?

PR #653, opened 2025-08-15
Follow-up for PR641 (Support for DeepSeekR1 model with SGLang / AI Dynamo)

TaekyungHeo:
## Summary
Follow up for [PR641](https://github.com/NVIDIA/cloudai/pull/641) (Support for DeepSeekR1 model with SGLang / AI Dynamo)

**Resolved by this PR**
1. Use _gpus_per_node ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2278015139))
2. prefill-initialized-regex from TOML ([link1](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2278015867), [link2](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2278016229))
3. Partially resolved - Initializing prefill-cmd ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277998091))
    * My response
      * It is okay to have an initial value. This does not prevent users from specifying prefill-cmd.
      * Updated the sglang.toml file to set prefill-cmd as we did for vllm.toml (https://github.com/Mellanox/cloudaix/pull/332).

**TODOs in follow-up PRs**
1. Differentiate HF_PATH from MODEL_PATH ([link1](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277993254), [link2](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277996028)). Currently, these two concepts are mixed up.
    * Related PR: https://github.com/NVIDIA/cloudai/pull/655
2. Deprecate prefill.num_nodes and decode.num_nodes ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277999243))
3. Multi-node support for SGLang ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2278008177))

**Potentially Rejected**
1. Mounting dynamo_repo_path ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277990146))
    * My response: There is no strong reason to make this mount optional. I also need dynamo_repo for K8s support. Therefore, it should always be mounted. There is no downside.
2. SGLang-specific container mounts ([link](https://github.com/NVIDIA/cloudai/pull/641#discussion_r2277997134))
    * My response: It is acceptable to use if-else statements to support different backends. There's no strong reason not to have backend-specific container mounts.

**Discussion Needed**
* https://github.com/NVIDIA/cloudai/pull/641#discussion_r2278002812
[...]

PR #657, opened 2025-08-15
Refactor GPU count to use _gpus_per_node in vllm and env validation

TaekyungHeo:
## Summary
Refactor GPU count to use _gpus_per_node in vllm and env validation

Follow-up PR for https://github.com/NVIDIA/cloudai/pull/653
Comment https://github.com/NVIDIA/cloudai/pull/653#discussion_r2279537492

## Test Plan
1. CI passes
2. Run on EOS

**VLLM**
```
$ python cloudaix.py run --system-config conf/common/system/eos.toml --tests-dir conf/staging/ai_dynamo/test --test-scenario conf/staging/ai_dynamo/test_scenario/deepseek_r1_distill_llama_8b.toml 

[INFO] System Name: EOS
[INFO] Scheduler: slurm
[INFO] Test Scenario Name: deepseek_r1_distill_llama_8b
[INFO] Test Scenario: deepseek_r1_distill_llama_8b

Section Name: Tests.1
  Test Name: vllm
  Description: vllm
  No dependencies
[INFO] Creating SlurmRunner
[INFO] Starting test: Tests.1
[INFO] Running test: Tests.1
[INFO] Submitted slurm job: 3453750
[INFO] Job completed: Tests.1 (iteration 1 of 1)
```
https://drive.google.com/drive/folders/1To8hLD7h9aGyVvEboXV3Qm44TGgEo6Ko?usp=drive_link

### DIFF
--- a/src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
+++ b/src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
@@ -277,7 +277,7 @@ _compute_worker_allocation_sglang() {
 _compute_worker_allocation_vllm() {
   local tp_arg_name="--${dynamo_args["tp-arg-name"]}"
   local pp_arg_name="--${dynamo_args["pp-arg-name"]}"
-  local num_gpus=$(echo "${CUDA_VISIBLE_DEVICES:-}" | tr ',' '\n' | grep -c .)
+  local num_gpus="$(_gpus_per_node)"
 
   if [[ $num_gpus -eq 0 ]]; then
     log "ERROR: No GPUs found in CUDA_VISIBLE_DEVICES"
@@ -569,8 +569,7 @@ validate_environment() {
   fi
 
   # GPU count sanity
-  local num_gpus
-  num_gpus=$(echo "${CUDA_VISIBLE_DEVICES}" | tr ',' '\n' | grep -c . || true)
+  local num_gpus="$(_gpus_per_node)"
   if [[ "$num_gpus" -le 0 ]]; then
     log "ERROR: Parsed zero GPUs from CUDA_VISIBLE_DEVICES='${CUDA_VISIBLE_DEVICES}'"
     exit 1


src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
```diff
@@ -569,8 +569,7 @@ validate_environment() {
   fi
 
   # GPU count sanity
-  local num_gpus
-  num_gpus=$(echo "${CUDA_VISIBLE_DEVICES}" | tr ',' '\n' | grep -c . || true)
+  local num_gpus="$(_gpus_per_node)"
```
srivatsankrishnan, line 572: Keep this. Based on the comment, you are doing sanity check by grepping by looking at the number of CUDA devices with what user specifies in `_gpus_per_node`. Sounds like t o good check to keep?

TaekyungHeo: @karya0 asked me to replace this with the function. `$(echo "${CUDA_VISIBLE_DEVICES}" | tr ',' '\n' | grep -c . || true)` is equivalent to `_gpus_per_node`

srivatsankrishnan: Sure. I meant, since you have some leeway to decide how you want to shape your FR. I am suggesting that, you don't have to update the num_gpus again with user provided vlue (in way key issue of code duplication is exists but its less verbose now).

You can add the check and see if that value matches with what user provides. Up to you and kapil.

srivatsankrishnan: Sounds good.

minor nit: IMO I don't see how this is sanity check at all. You are just assigning a variable that user provides and redefine it again. I am not sure if thats what @karya0 agreed as well.

TaekyungHeo: By performing a sanity check, it determines whether the number of GPUs is zero. If the number of GPUs is zero, the check fails. @karya0, please confirm. You can arrange a call with @karya0 if you have any concerns.

PR #692, opened 2025-09-21
ucc_perftest_add_gen_and_a2av

yaeliyac:
As explained in release documentation [UCC v1.5](https://github.com/openucx/ucc/releases/tag/v1.5.0)
UCC v1.5 added a perftest generator [PR #1147](https://github.com/openucx/ucc/pull/1147) 
This PR added the --gen flag as added to ucc and an example for alltoallv perftest using this.

Important: UCC v1.5 will be only in the next pytorch container so this wont work in current versions.

### DIFF
--- a/src/cloudai/workloads/ucc_test/ucc.py
+++ b/src/cloudai/workloads/ucc_test/ucc.py
@@ -71,7 +71,6 @@ class UCCCmdArgs(CmdArgs):
     gen: Union[str, list[str], None] = None
 
 
-
 class UCCTestDefinition(TestDefinition):
     """Test object for UCC."""
 


src/cloudai/workloads/ucc_test/slurm_command_gen_strategy.py
```diff
@@ -39,6 +39,12 @@ def generate_test_command(self) -> List[str]:
         srun_command_parts.append(f"-c {tdef_cmd_args.collective}")
         srun_command_parts.append(f"-b {tdef_cmd_args.b}")
         srun_command_parts.append(f"-e {tdef_cmd_args.e}")
+        if tdef_cmd_args.gen is not None:
```
amaslenn, line 42: List in CmdArgs classes are for sweeps, they are unrolled during the run.

Inside `generate_test_command()` we assume that list for an arg is not possible, so let's do simple `srun_command_parts.append(f"--gen {tdef_cmd_args.gen}")` only when `gen` is set.

Unless setting them all in a single cmd is what you need, of course. But in this case we need another way to approach, please let me know if this is the case.

conf/common/test/ucc_test_alltoallv.toml
amaslenn, line 1: Can we get rid of this file and use `ucc_base_test` + args inside scenario?

conf/common/test_scenario/ucc_test.toml
```diff
@@ -61,3 +61,13 @@ num_nodes = 2
...
+
+[[Tests]]
+id = "Tests.alltoallv"
+test_name = "ucc_test_alltoallv"
+description = "UCC alltoallv"
+time_limit = "00:20:00"
+num_nodes = 2
+  [Tests.cmd_args]
+  collective = "alltoallv"
+  gen = "file:name=/workspace/ucc_transfer_matrices_new/bench0_intput_matrices.txt"
```
amaslenn, line 73: Does this file exist as part of container?

conf/common/test/ucc_test_alltoallv.toml
```diff
@@ -0,0 +1,9 @@
+name = "ucc_test_alltoallv"
+description = "alltoallv"
+test_template_name = "UCCTest"
+
+extra_container_mounts = ["path_to_host_matrices:path_to_container_matrices_as_used_in_gen"]
```
amaslenn, line 5: We can't use such values in configs under `common/` folder: they are supposed to work out of the box without any modifications.

Does this benchmark require some input before running it? Do we have an input example as part of container?

amaslenn: I see. Let's try something like `extra_container_mounts = ["$PWD/conf/common:/workspace"]` to ensure it will work for everyone. If not possible, we have two options:
1. Put configs into another folder (not `common/`) and add description how to run it.
2. Not having an example at all (not good).

amaslenn (dismissed):

PR #716, opened 2025-11-04
Simplify internal hierarchy of classes

amaslenn:
## Summary
The goal of this change is to de-clutter some legacy components that were originally useful, but lost their meaning during ongoing modifications.

1. Aligned `JsonGenStrategy` with `CmdGenStrategy`
    a. Added `TestRun` object as a class member
    b. `TestTemplateStrategy` was completely removed to simplify hierarchy.
2. Aligned `Grader` with other strategies
    a. `Grader` is a subject for removal, so it doesn't fully follow the JsonGen/CommandGen approach
    b. Removed unused `strategy_registry.py` (what was this for?)
3. Removed `TestTemplate` concept. Previously it held all the strategies, but now it is obsolete.
4. Removed `Test` concept. This reduces nesting for accessing cmd args via `TestDefinition` and simplifies overall hierarchy.

## Test Plan
1. CI (updated)
2. Real runs ~~TBD~~ see in comments

## Additional Notes
—

## Summary by CodeRabbit

* **Refactor**
  * Simplified test structure by eliminating intermediate wrapper abstraction and using test definitions directly throughout the system.
  * Reorganized strategy registry system with separate, dedicated mappings for JSON generation and grading strategies for improved clarity and maintainability.
  * Updated strategy instantiation to capture test context at construction time rather than method invocation, enabling more flexible strategy design.

### DIFF
--- a/src/cloudai/util/__init__.py
+++ b/src/cloudai/util/__init__.py
@@ -90,7 +90,7 @@ def prepare_output_dir(path: Path) -> Optional[Path]:
         return None
 
 
-def deep_merge(a: dict, b: dict):
+def deep_merge(a: dict, b: dict) -> dict:
     result = a.copy()
     for key in b:
         if key in result:


srivatsankrishnan (changes requested):
Please do this in two stages to enable gerrit code review as well.  Currently the limit is set to 100 files.  Either increase the limit or break it two stages. Also which user is asking for this refactor now?

amaslenn:
> Please do this in two stages to enable gerrit code review as well. Currently the limit is set to 100 files. Either increase the limit or break it two stages. Also which user is asking for this refactor now?

I don't see how this limit can be configured. Splitting this change doesn't make much sense, plus it was tested via real runs.

srivatsankrishnan:
> @srivatsankrishnan please unblock this PR

Seems like a architectural modifications. What would the changes here? There should be a respective PR for CloudAIx correct?
https://github.com/Mellanox/cloudaix/tree/main/workloads

@manleyalex  Please see how much changes need to happen for the prompting templates now. FYI.

manleyalex:
> > @srivatsankrishnan please unblock this PR
> 
> Seems like a architectural modifications. What would the changes here? There should be a respective PR for CloudAIx correct? https://github.com/Mellanox/cloudaix/tree/main/workloads
> 
> @manleyalex Please see how much changes need to happen for the prompting templates now. FYI.

Looks good to me, minimal changes are needed to my prompts.

PR #756, opened 2025-12-16
AIConfigurator in CloudAI

srivatsankrishnan:
## Summary
This PR adds AIConfigurator (aiconfig) to CloudAI workload. AIConfigurator is a standalone tool and doesn't require slurm or any GPU nodes. The workload runs prediction via the installed aiconfigurator package and runs disaggregated modes. generates a reproducible run_simple_predictor.sh per test-run, and parses predictor JSON output into CloudAI metrics via a dedicated report strategy.

## Test Plan
* CI/CD
* AIconfigurator runs on Llama3_70b 

```
cloudai run --system-config conf/common/system/standalone_system.toml --tests-dir 
conf/experimental/aiconfigurator/test --test-scenario conf/experimental/aiconfigurator/test_scenario/aiconfigurator_disa
gg.toml 
[INFO] System Name: standalone_system
[INFO] Scheduler: standalone
[INFO] Test Scenario Name: aiconfigurator-disagg-scenario
[INFO] Checking if workloads components are installed.
[INFO] Test Scenario: aiconfigurator-disagg-scenario

Section Name: aiconfigurator.disagg.demo
  Test Name: dse_aiconfigurator_disagg_demo_Llama3.1_70B
  Description: Aiconfigurator disaggregated predictor DSE sweeps
  No dependencies
[INFO] Initializing Runner [RUN] mode
[INFO] Creating StandaloneRunner
[INFO] Scenario results will be stored at: results/aiconfigurator-disagg-scenario_2025-12-16_14-40-00
[INFO] Running step 1 (of 32) with action {'disagg.p_tp': 1, 'disagg.p_pp': 1, 'disagg.p_dp': 1, 'disagg.p_workers': 1, 'disagg.d_tp': 1, 'disagg.d_pp': 1, 'disagg.d_dp': 1, 'disagg.d_bs': 8, 'disagg.d_workers': 1}
[INFO] Starting test: aiconfigurator.disagg.demo (results at: results/aiconfigurator-disagg-scenario_2025-12-16_14-40-00/aiconfigurator.disagg.demo/0/1)
[INFO] Running test: aiconfigurator.disagg.demo
[INFO] Executing command for test aiconfigurator.disagg.demo: bash /Users/srivatsank/workspace/aiconfigurator/cloudai/results/aiconfigurator-disagg-scenario_2025-12-16_14-40-00/aiconfigurator.disagg.demo/0/1/run_simple_predictor.sh
[...]

PR #762, opened 2025-12-19
Fix docs build failure by ensuring stable SCM versioning in CI workflow

srivatsankrishnan:
## Summary

Fix docs build failure caused by SCM version mismatch during uv sync. This PR also fixes the bugs in the orignal CI/CD when using uv as a default way to use CloudAI. For instance, [PR 701](https://github.com/NVIDIA/cloudai/pull/701) and [PR 702](https://github.com/NVIDIA/cloudai/pull/702) makes the first version to upload the uv lock file.

This PR also fixes the bug introduced by [PR 684](https://github.com/NVIDIA/cloudai/pull/684) where deploys the CI flow for building and publishing the GH document but does not introduce necessary checks to ensure the `uv.lock` file is pushed everytime the python package dependencies are modified and developer uses `uv` to run.

### Root cause
The failing step was building/installing cloudai itself during the docs workflow’s uv sync.

- cloudai uses SCM-based dynamic versioning (pdm-backend + source = "scm").

- In CI, the repo was checked out in a way that could be shallow / missing full git history/tags, which makes SCM-version computation flaky.

- Under that condition, the build backend can produce inconsistent versions between wheel filename and wheel metadata, leading to:
--metadata version: 0.1.dev…+g<sha>
--wheel filename version: 0.1.dev…+g<sha>.dYYYYMMDD

Tools reject this as an invalid wheel (“metadata version does not match wheel filename”), so the docs job fails during “Set up envir

## Test Plan
- CI/CD
- Manual Github action workflow (fork)
image

## Additional Notes
Keeping the pytest with pip based flow since we also support pip as described in the documentation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,29 +30,29 @@ jobs:
         run: uv lock --check
 
       - name: Install dependencies
-        run: uv sync --extra dev --extra docs
+        run: uv sync --frozen --extra dev --extra docs
 
       - name: Run ruff linter
-        run: ruff check
+        run: uv run ruff check
 
       - name: Run ruff formatter
-        run: ruff format --check --diff
+        run: uv run ruff format --check --diff
 
       - name: Run pyright
-        run: pyright
+        run: uv run pyright
 
       - name: Run vulture check
-        run: vulture src/ tests/
+        run: uv run vulture src/ tests/
 
       - name: Import linter
-        run: PYTHONPATH=src lint-imports
+        run: PYTHONPATH=src uv run lint-imports
 
       - name: Check TOML formatting
         run: |
           set -eE
           set -o pipefail
 
-          taplo fmt --check --diff
+          uv run taplo fmt --check --diff
 
       - name: Build documentation
         run: |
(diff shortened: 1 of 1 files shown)


amaslenn (changes requested):
This PR adds a workaround addressing symptoms, not the root cause. The real issue is that #760 didn't update `uv.lock` file.

Please add the following command to avoid such issues in future: `uv lock --check` (right after "Install dependencies" step under "lint" job).

@srivatsankrishnan please do not create new PRs for the same topic as you did with (#756 and #760), that ruins review process.

alexmanle (approved):

.github/workflows/ci.yml
```diff
@@ -13,42 +13,54 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Python
```
amaslenn: Since you've also migrated to `uv`, setting up python separately is not needed, `uv sync` will do it automatically.

srivatsankrishnan:
Andrei will provide the fix to the CI bug.

PR #766, opened 2026-01-05
Remove hardcoded `--distribution=arbitrary`

juntaowww:
## Summary
Remove hardcoded `--distribution=arbitrary`, this mode can lead to weaker affinity/pinning behavior unless explicit binding is added, causing a noticeable performance drop.

Also the distribution mode could be specified in the system configuration file, could safely remove it here.

## Testing Notes
Compared walltime/throughput before vs after, regression disappears when removing `--distribution=arbitrary`


amaslenn (commented):
Thank you for your contribution! Please address CI issues and we will proceed.

juntaowww:
Thanks! The CI issue seems to be the copyright year, should I fix it in this PR or create a new one?

PR #784, opened 2026-01-21
Make tests more stable on systems with slurm binaries

amaslenn:
## Summary
Make tests more stable on systems with slurm binaries by disabling real system check.

## Test Plan
1. CI

## Additional Notes
—


alexmanle (approved):
Looks good!

PR #795, opened 2026-02-02
Fix `nsys` subfield merging behavior

juntaowww:
## Summary
When specifying `[Tests.nsys]` in a scenario file to override only specific `nsys` configuration fields (like output), the entire `nsys` object was being overwritten instead of just the specified fields.

Fix:
```python
"nsys": self.nsys.model_dump() if self.nsys else None,
```
to:
```python
"nsys": self.nsys.model_dump(exclude_unset=True) if self.nsys else None,
```
Using `exclude_unset=True` ensures that only fields explicitly set by the user in the scenario file are included in the dump. The `deep_merge` function then correctly merges only those specified fields with the test configuration, preserving all other `nsys` settings from the test definition.

## Test Plan
`TestNsysMerging` is added to `tests/test_test_scenario.py` for testing the correctness of overwriting of `nsys` fields from scenario configuration.

### DIFF
--- a/tests/test_test_scenario.py
+++ b/tests/test_test_scenario.py
@@ -25,6 +25,7 @@ from cloudai._core.exceptions import MissingTestError
 from cloudai.core import (
     CmdArgs,
     GitRepo,
+    NsysConfiguration,
     PredictorConfig,
     Registry,
     ReportGenerationStrategy,
@@ -521,8 +522,6 @@ class TestNsysMerging:
     def test_nsys_partial_override_preserves_base_config(
         self, test_scenario_parser: TestScenarioParser, slurm_system: SlurmSystem
     ):
-        from cloudai.core import NsysConfiguration
-
         test_scenario_parser.test_mapping = {
             "nccl": NCCLTestDefinition(
                 name="nccl",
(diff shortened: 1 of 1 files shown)


tests/test_test_scenario.py
```diff
@@ -515,3 +515,160 @@ def test_get_reporters_nccl(self):
...
+
+        assert tdef.nsys is not None
+        assert tdef.nsys.output == "/scenario/output"
+        assert tdef.nsys.nsys_binary == "/custom/nsys"
+        assert tdef.nsys.trace == "cuda,nvtx"
+        assert tdef.nsys.sample == "cpu"
+        assert tdef.nsys.enable is True
+
+    def test_nsys_multiple_fields_override(self, test_scenario_parser: TestScenarioParser, slurm_system: SlurmSystem):
+        from cloudai.core import NsysConfiguration
```
amaslenn: Can it be imported on top of this file? (same for all tests)

juntaowww: Put them back on top of the file.

Indeed AI is used, could add some rules to prevent this in the future.

amaslenn (approved):

PR #888, opened 2026-05-12
Append trajectory row on cache hits

rutayan-nv:
## Summary

`CloudAIGymEnv.step()` short-circuits when the requested action has already been evaluated, returning the cached reward/observation. Currently this path **does not** append a row to `trajectory.csv` or to `self.current_trajectory`, so the visible trajectory contains only the executed (non-cached) steps even though the agent observed `agent_steps` interactions.

Downstream consequences:
- Step counts in reports / `analysis.json` diverge from the agent's actual interaction count.
- The CSV row index is no longer a stable proxy for "agent step", breaking offline analysis tools that assume one row per call to `step()`.
- Any future offline-RL / behavior-cloning corpus built from `trajectory.csv` is missing transitions.

This PR writes a `TrajectoryEntry` for cache hits using the **current** step index, reusing the cached reward and observation. The log line is also expanded to record the originating step the cached result was taken from.

## Changes

- `src/cloudai/configurator/cloudai_gym.py`: append a `TrajectoryEntry` in the cache-hit branch of `step()`; log the source step.
- `tests/test_cloudaigym.py`: add `test_cached_step_appends_trajectory_row` that primes `env.trajectory`, invokes `step()`, and asserts a new row appears in both the in-memory trajectory and `trajectory.csv`.

## Test plan

- [x] `pytest tests/test_cloudaigym.py` — 27 passed locally.
- [x] `ruff check` on the touched files — clean.
- [ ] CI green.

### DIFF
--- a/src/cloudai/configurator/cloudai_gym.py
+++ b/src/cloudai/configurator/cloudai_gym.py
@@ -123,8 +123,17 @@ class CloudAIGymEnv(BaseGym):
         cached_result = self.get_cached_trajectory_result(action)
         if cached_result is not None:
             logging.info(
-                "Retrieved cached result from trajectory with reward %s. Skipping step.",
+                "Retrieved cached result from trajectory with reward %s (from step %s). Skipping execution.",
                 cached_result.reward,
+                cached_result.step,
+            )
+            self.write_trajectory(
+                TrajectoryEntry(
+                    step=self.test_run.step,
+                    action=action,
+                    reward=cached_result.reward,
+                    observation=cached_result.observation,
+                )
             )
             return cached_result.observation, cached_result.reward, False, {}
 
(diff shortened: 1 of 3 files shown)


rutayan-nv:
@alexmanle FYI

rutayan-nv:
@coderabbitai resolved

done - af2e9c4a2d5f7d3b90fcb6cbf7ef4489878d48a0

Addressed in [af2e9c4](https://github.com/NVIDIA/cloudai/pull/888/commits/af2e9c4a2d5f7d3b90fcb6cbf7ef4489878d48a0) — added `runner.run.assert_not_called()` immediately after `env.step(cached_action)` in `test_cached_step_appends_trajectory_row`, before the remaining reward / observation / trajectory / csv assertions.

podkidyshev (approved):

PR #891, opened 2026-05-14
Add UCC/NCCL alltoallv, deepEP v1/v2 and moe benchmark

ybenvidia:
Add UCC/NCCL alltoallv vs DeepEP
- add ucc alltoallv test
- add nccl alltoallv test
- add report generation for DeepEP vs NCCL vs UCC


podkidyshev:
please resolve coderabbit comments and make CI pass before review

ping me again directly once it's done please

src/cloudai/workloads/deepep/slurm_command_gen_strategy.py
```diff
@@ -44,96 +110,107 @@ def _append_head_node_detection(self, batch_script_content: List[str]) -> None:
...
 
     def _append_sbatch_directives(self, batch_script_content: List[str]) -> None:
-        """
-        Append SBATCH directives and head node detection setup for DeepEP.
-
-        Args:
-            batch_script_content: The list of script lines to append to.
-        """
-        super()._append_sbatch_directives(batch_script_content)
+        num_nodes, node_list = self.get_cached_nodes_spec()
```
podkidyshev: is this implementation anyhow different from the base class? I can't find a difference...

src/cloudai/workloads/deepep/slurm_command_gen_strategy.py
```diff
@@ -44,96 +110,107 @@ def _append_head_node_detection(self, batch_script_content: List[str]) -> None:
...
+        batch_script_content.append(
+            "\nexport SLURM_JOB_MASTER_NODE=$(scontrol show hostname $SLURM_JOB_NODELIST | head -n 1)"
+        )
         self._append_head_node_detection(batch_script_content)
 
-    def _container_mounts(self) -> List[str]:
-        """Return container mounts specific to DeepEP benchmark."""
-        tdef: DeepEPTestDefinition = cast(DeepEPTestDefinition, self.test_run.test)
-        cmd_args: DeepEPCmdArgs = tdef.cmd_args
+    def _gen_srun_command(self) -> str:
```
podkidyshev, line 123: and this one

src/cloudai/workloads/nccl_test/slurm_command_gen_strategy.py
```diff
@@ -14,29 +14,113 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pathlib import Path
 from typing import List, cast
 
 from cloudai.systems.slurm import SlurmCommandGenStrategy
+from cloudai.workloads.moe_benchmark.combined_report import MOE_BENCHMARK_PREV_MOUNT, moe_benchmark_root
```
podkidyshev: what looks to me somewhat wrong, is that moe_benchmark python package imports something from nccl_test and here nccl_test imports something from moe_benchmark

in that case I suggest moving one of the components to `src/cloudai/workloads/common` so that it is a shared place with one-directional import

src/cloudai/workloads/ucc_test/slurm_command_gen_strategy.py
```diff
@@ -14,18 +14,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pathlib import Path
 from typing import List, cast
 
 from cloudai.systems.slurm import SlurmCommandGenStrategy
+from cloudai.workloads.moe_benchmark.combined_report import MOE_BENCHMARK_PREV_MOUNT, moe_benchmark_root
```
podkidyshev: same here

podkidyshev (changes requested):

podkidyshev:
please also resolve remaining coderabbit comments 🙏

src/cloudai/workloads/deepep/slurm_command_gen_strategy.py
ybenvidia: what do u mean ?

podkidyshev: https://github.com/NVIDIA/cloudai/pull/891#discussion_r3386996845

podkidyshev: I can see that this re-implementation of _gen_srun_command is different from the base class

ybenvidia: Good catch — the body is indeed a copy of the base _gen_srun_command. There's exactly one intentional difference: the srun geometry.

Base: gen_srun_prefix(use_pretest_extras=True) → emits -N{N} and lets the system-wide --ntasks-per-node (= gpus_per_node, i.e. 8) drive the task count.
DeepEP: gen_srun_prefix(..., with_num_nodes=False) then pins --nodes=N --ntasks=N --ntasks-per-node=1.
This is required because DeepEP launches one torchrun per node (torchrun --nproc_per_node forks the per-GPU workers itself). With the base geometry, srun would start one torchrun per GPU → 8 launchers/node → GPU oversubscription (we hit NCCL's "too many XML nodes" error this way).

I tried moving this into a gen_srun_prefix override to drop the duplicated tail, but that broke things: gen_srun_prefix is also used by the mapping and metadata sruns in the sbatch, which must keep the default -N{N} geometry. So the geometry tweak has to stay local to the launcher _gen_srun_command.

podkidyshev (approved):

PR #892, opened 2026-05-15
[Configurator] Make select_action observation-aware

rutayan-nv:
- `BaseAgent.select_action(observation: Optional[list[float]] = None)` so observation-conditioned agents (RL, contextual bandits) can use the latest env observation; stateless agents (grid search, BO) keep the same behavior.
- `handle_dse_job` seeds the loop with `env.reset()`, threads `observation` into `select_action`, and forwards the gymnasium-style transition (`observation`, `prev_observation`, `action`, `done`) alongside the existing `trial_index` / `value` feedback to `update_policy`.
- `GridSearchAgent.select_action` updated to the new signature.

### DIFF
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -20,7 +20,7 @@ import logging
 import signal
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Callable, List, Optional
+from typing import Callable, List, Optional
 from unittest.mock import Mock
 
 import toml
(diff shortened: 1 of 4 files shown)


src/cloudai/cli/handlers.py
```diff
@@ -157,16 +183,26 @@ def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
...
+            observation, reward, done, *_ = env.step(action)
+            agent.update_policy(
+                _build_step_feedback(
+                    step=step,
+                    action=action,
+                    reward=reward,
+                    observation=observation,
+                    prev_observation=prev_observation,
+                    done=done,
+                )
```
podkidyshev: nit: I think this separate function only reduce readibility as it does just a single op. you could simply replace `_build_step_feedback` here with `dict` and behavior won't change

tests/test_handlers.py
```diff
@@ -60,7 +60,7 @@ def get_config_class() -> type[StubAgentConfig]:
     def configure(self, config: dict[str, Any]) -> None:
         raise NotImplementedError
 
-    def select_action(self) -> tuple[int, dict[str, Any]]:
+    def select_action(self, observation: Optional[list[float]] = None) -> tuple[int, dict[str, Any]]:
```
podkidyshev: nit:

1. If we introduce an additional typing hint var (`Observation`) - let's use it consistently. To my taste I would not add it at all and just use `list[float]` but it's personal, you could leave it be :)
2. please use `observation: Observation | None = None` notation instead of Optional

podkidyshev (dismissed):

src/cloudai/cli/handlers.py
rutayan-nv: Done — inlined as a dict literal in `ca40cdc4`.

tests/test_handlers.py
rutayan-nv: Done — dropped the `Observation` alias and switched to `list[float] | None` everywhere in `ca40cdc4`.

rutayan-nv:
@podkidyshev — feedback addressed in `ca40cdc4`; ready for another look when you have a moment.

podkidyshev (approved):

PR #893, opened 2026-05-15
[CLI/Core] Polymorphic agent dispatch + TestRun owns trial counter

rutayan-nv:
## Issue
- Dispatch: `handle_dse_job` branched on agent identity (flag + Protocol + TypeGuard) to support custom training loops.
- Trial counter: `TestRun.step` had multiple writers; RLlib's frequent `reset()` collapsed every trial onto `step=1`, overwriting `trajectory.csv`/`env.csv` rows.
- Failure handling: the DSE loop didn't distinguish recoverable failures from bugs.

## Fix
- Dispatch: `BaseAgent.run()` is the polymorphic entry point and `handle_dse_job` collapses to `err |= agent.run()` (deletes the flag/Protocol/TypeGuard/helper, −89 lines).
- Trial counter: `TestRun.increment_step()` is the sole mutator, called only by `CloudAIGymEnv.step()`.
- Failure handling: hybrid contract — a non-zero rc accumulates and continues; an exception hard-fails (reports still generated, error written to `<scenario_root>/dse_failure.txt`, then re-raised).

## Testing
- `pytest tests`: 1607 passed, 4 skipped; ruff + pyright + vulture clean.
- Downstream cloudaix RL workloads (PPO + DQN) reinstalled on this branch — full suite green; contract tests confirm `tr.step` monotonicity through the adapter and `env.csv`.

**Stack:** `main` ← **#893** ← #900 ← #901 (merge bottom-up).

### DIFF
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -146,18 +146,30 @@ def _has_custom_training_loop(agent: object) -> TypeGuard[CustomTrainingLoopAgen
 
 
 def _run_custom_training_loop(agent: CustomTrainingLoopAgent, agent_type: str) -> int:
-    """Drive an agent's self-contained training loop and return a process-style exit code."""
+    """
+    Drive an agent's self-contained training loop and return a process-style exit code.
+
+    ``shutdown()`` runs inside its own ``try/except`` so a faulty teardown cannot
+    suppress the exit code from ``train()`` nor propagate out of this helper:
+    ``handle_dse_job`` relies on the returned ``rc`` to accumulate ``err |= rc``
+    and continue with the remaining test runs.
+    """
     logging.info(f"Agent {agent_type} drives its own training loop; delegating to agent.train().")
+    rc = 0
     try:
         agent.train()
-        return 0
     except Exception:
         logging.exception(f"Custom training loop failed for agent {agent_type}.")
-        return 1
+        rc = 1
     finally:
         shutdown = getattr(agent, "shutdown", None)
         if callable(shutdown):
-            shutdown()
+            try:
+                shutdown()
+            except Exception:
+                logging.exception(f"Shutdown failed for agent {agent_type}.")
+                rc = 1
+    return rc
 
 
 def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -18,6 +18,7 @@ import argparse
 import copy
 import logging
 import signal
+import traceback
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Callable, List, Optional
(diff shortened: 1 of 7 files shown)


src/cloudai/cli/handlers.py
```diff
@@ -118,6 +118,40 @@ def prepare_installation(
...
+    try:
+        agent.train()
+        return 0
+    except Exception:
+        logging.exception(f"Custom training loop failed for agent {agent_type}.")
+        return 1
+    finally:
+        shutdown = getattr(agent, "shutdown", None)
+        if callable(shutdown):
+            shutdown()
```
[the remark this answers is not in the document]
rutayan-nv: Done — `3ffe8935`.

src/cloudai/cli/handlers.py
```diff
@@ -118,6 +118,60 @@ def prepare_installation(
     return installables, installer
 
 
+@runtime_checkable
```
podkidyshev: let's move this code into base_agent.py. handlers.py is already too long

as for the tests against `_run_custom_training_loop`: I'm starting to make the tests folder structure replicate the main code structure. so in this case, I'd place all the relevant tests you added into tests/configurator/test_base_agent.py

(not related to tests against `handle_dse_job`)

src/cloudai/cli/handlers.py
```diff
@@ -157,6 +211,10 @@ def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
 
         agent = agent_class(env, agent_config)
 
+        if _has_custom_training_loop(agent):
+            err |= _run_custom_training_loop(agent, agent_type)
```
podkidyshev: shouldn't we exit (immediate `return err`) if `err` is greater than zero? The existing code above doesn't really treat the `err` well but maybe it's the time to start doing so :D

podkidyshev (changes requested):

rutayan-nv:
Downstream cloudaix PR that consumes this contract: https://github.com/Mellanox/cloudaix/pull/589

It ports `GymnasiumAdapter` to cloudaix (where its only consumer — the RL agent stack — lives), introduces a `rllib_run` dispatch helper for shared RLlib orchestration code, and refactors PPO/DQN onto the new `BaseAgent.run()` polymorphism. The cloudaix PR also pins the user-visible `env.csv` monotonicity contract — that's the test that would have caught the original cluster-run bug at unit-test time.

Together the two PRs are the architectural fix for the trial-counter collapse: cloudai owns the trial-counter contract (`TestRun.increment_step()` + `CloudAIGymEnv.step()` as sole mutator), cloudaix owns the RL-specific glue (adapter + dispatch helper + agent overrides).

src/cloudai/cli/handlers.py
rutayan-nv: makes sense, will do.

src/cloudai/cli/handlers.py
rutayan-nv: good point. Let me check that and change it!

src/cloudai/cli/handlers.py
rutayan-nv: Done in the `BaseAgent.run()` polymorphism refactor (latest push). The custom-loop dispatch, the `CustomTrainingLoopAgent` Protocol, and the TypeGuard are all gone from `handlers.py` — `handle_dse_job` now collapses to `err |= agent.run()`, and agents that drive their own training loop just override `run()`. The dispatcher/helper unit tests were replaced with polymorphic tests asserting `handle_dse_job` delegates to `agent.run()` and propagates the return code (`tests/test_handlers.py`).

src/cloudai/cli/handlers.py
rutayan-nv: I looked into this and decided to keep accumulate-and-continue rather than early-return. A DSE scenario can contain multiple `test_runs` (multiple agents) that run consecutively, and a failure in one shouldn't prevent the remaining independent runs from executing — we still want their results and reports. The `err |= agent.run()` accumulation preserves the failure signal: the handler returns non-zero if *any* run failed, so CI still fails on error. Early-returning would mask the other runs' outcomes. Happy to switch to fail-fast for the whole scenario if you'd prefer that semantics.

rutayan-nv:
@podkidyshev rebased onto latest `main` and addressed your review:

- **"move dispatch out of handlers.py"** — done via the `BaseAgent.run()` polymorphism refactor: `handle_dse_job` now collapses to `err |= agent.run()`; the Protocol/TypeGuard/helpers are gone from `handlers.py`. Dispatcher unit tests replaced with polymorphic tests in `tests/test_handlers.py`.
- **"early-return on err > 0"** — replied inline: keeping accumulate-and-continue so independent `test_runs` in a scenario still execute; the OR'd `err` preserves a non-zero exit if any run fails. Glad to switch to fail-fast if you prefer.

Also note a contract fix surfaced by the rebase: `main` had evolved the dispatch loop to thread observations (`env.reset()` → `select_action(observation=...)`), so I ported that threading into `BaseAgent.run()` rather than regressing observation-conditioned agents. Full suite green (1637 passed). Re-review would be appreciated when you have a moment.

src/cloudai/cli/handlers.py
```diff
@@ -157,17 +157,7 @@ def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
...
-            if result is None:
-                break
-            step, action = result
-            env.test_run.step = step
-            logging.info(f"Running step {step} (of {agent.max_steps}) with action {action}")
-            observation, reward, *_ = env.step(action)
-            feedback = {"trial_index": step, "value": reward}
-            agent.update_policy(feedback)
-            logging.info(f"Step {step}: Observation: {[round(obs, 4) for obs in observation]}, Reward: {reward:.4f}")
+        err |= agent.run()
```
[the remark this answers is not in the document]
rutayan-nv: Fixed in `11f8625f`. `handle_dse_job` now wraps `err |= agent.run()` in a try/except: an exception from one `TestRun` is logged with the agent identity, converted to `err |= 1`, and the loop continues with the remaining independent runs. This matches the sibling `agent_class is None` branch and honors the dependency-free "all cases run consecutively" contract enforced at the top of the same function — and it ensures the accumulated `err` (and `generate_reports`) actually survive to become the process exit code instead of being lost to an unwound stack.

Added two regression tests in `tests/test_handlers.py`:
- `test_handle_dse_job_guards_agent_run_exception` — a raising `run()` yields exit code 1 (no propagation) and logs the agent name.
- `test_handle_dse_job_continues_remaining_runs_after_exception` — two `TestRun`s; the first agent raises, the second still executes (`run_calls == 2`).

Both fail without the guard (exception propagates) and pass with it.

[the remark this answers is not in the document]
rutayan-nv: Thanks @coderabbitai — after discussing with the maintainers we're reverting this guard in favor of a hybrid failure contract, so flagging the reasoning for the record.

`handle_dse_job` now standardizes on:

- **Recoverable failures return a non-zero rc.** These accumulate via `err |= agent.run()` and the loop continues to the next `TestRun` — exactly the `err` accumulation this thread asked to preserve. It's the established convention in the codebase: `rllib_run` catches training errors and returns `rc=1`, and `CloudAIGymEnv.step` maps a failed/missing metric to `rewards.metric_failure` rather than raising. Workload-level failures therefore never reach `run()` as exceptions.
- **Exceptions out of `run()` are treated as framework/agent bugs and hard-fail.** Catching them and converting to `err |= 1; continue` would mask genuine defects behind a non-zero exit code (and silently skew an entire sweep), which we'd rather surface loudly.

Your secondary concern — that an abort bypasses report generation — is valid and we've kept it: `handle_dse_job` now captures the aborting exception, still runs `generate_reports`, documents the error in `<scenario_root>/dse_failure.txt`, and then re-raises. So reports are no longer skipped on a hard-fail.

The contract is documented in `BaseAgent.run()` and pinned by tests: a non-zero rc accumulates and continues; a raising `run()` propagates and aborts remaining runs; and on a hard-fail the reports + failure record are produced before re-raising.

rutayan-nv:
Closed by a branch rename (`custom-training-loop-dispatch` → `agent-run-polymorphism`). The work continues unchanged in #933, with the history rebuilt into 2 clean commits (the intermediate add-then-remove custom-loop-dispatch churn was squashed away). The net diff is identical.

PR #894, opened 2026-05-15
[Configurator] Add GymnasiumAdapter for CloudAI envs

rutayan-nv:
- `GymnasiumAdapter` wraps a CloudAI `BaseGym` as a `gymnasium`-compatible env: `Dict` of `Discrete` actions over the tunable params, `float32` `Box` observation derived from `define_observation_space()`, gymnasium 5-tuple from `step` / `step_raw`.
- Fixed (single-value) params are injected on every step so agents only ever see the tunable subset; `step_raw` bypasses index decoding and fixed injection for callers that already have a fully-formed parameter dict.
- The adapter mirrors `handle_dse_job`'s 1-based `test_run.step` so artifact paths match whether the env is driven by the DSE loop or by an external training loop wrapping the adapter.
- `CloudAIGymEnv.define_observation_space()` now returns one slot per agent metric (at least one), giving adapters the correct `Box` shape.
- `gymnasium` is an optional dependency: lazy-imported inside the adapter and exposed via a new `rl` extra (also added to `dev`) pinned to `~=1.2`.

### DIFF
--- a/src/cloudai/configurator/gymnasium_adapter.py
+++ b/src/cloudai/configurator/gymnasium_adapter.py
@@ -1,0 +1,137 @@
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Any, ClassVar, Optional
+
+from .base_gym import BaseGym
+
+_GYMNASIUM_INSTALL_HINT = "gymnasium is required for GymnasiumAdapter. Install it with: pip install gymnasium"
+
+
+def _import_gymnasium():
+    """
+    Import gymnasium + numpy lazily; raise a clear, actionable error when absent.
+
+    Kept as a single seam so that:
+    * cloudai installs without ``gymnasium`` continue to work for users that don't
+      need this adapter (the import is gated behind ``GymnasiumAdapter()``);
+    * tests can patch this helper to simulate a missing install.
+    """
+    try:
+        import gymnasium
+        import numpy as np
+        from gymnasium import spaces
+
+        return gymnasium, spaces, np
+    except ImportError as exc:
+        raise ImportError(_GYMNASIUM_INSTALL_HINT) from exc
+
+
+class GymnasiumAdapter:
+    """
+    Expose a CloudAI :class:`BaseGym` as a standard ``gymnasium.Env``-shaped object.
+
+    The adapter:
+
+    * builds a ``gymnasium.spaces.Dict`` of ``Discrete`` action spaces over the
+      *tunable* parameters (those with more than one candidate value), and
+      injects the *fixed* parameters (single candidate) automatically on every
+      step so agents never see them.
+    * converts observations to ``float32`` ``numpy`` arrays sized by
+      ``env.define_observation_space()``.
+    * returns the gymnasium 5-tuple ``(obs, reward, terminated, truncated, info)``
+      from :meth:`step` and :meth:`step_raw`.
+    * keeps ``env.test_run.step`` in sync (1-based) so artifact paths produced by
+      ``CloudAIGymEnv`` match those produced by ``handle_dse_job`` (i.e.
+      ``<scenario>/<test>/<iteration>/<step>/`` for every evaluation), which is
+      required when a custom training loop (e.g. RLlib) front-ends the env.
+
+    ``gymnasium`` and ``numpy`` are optional dependencies; importing this module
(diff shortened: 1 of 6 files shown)


src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,137 @@
...
+        obs, info = self._env.reset(seed=seed, options=options)
+        return self._as_obs_array(obs), info
+
+    def step(self, action: dict[str, int]) -> tuple[Any, float, bool, bool, dict[str, Any]]:
+        params = {**self._fixed_params, **self.decode_action(action)}
+        return self._step_with_params(params)
+
+    def step_raw(self, params: dict[str, Any]) -> tuple[Any, float, bool, bool, dict[str, Any]]:
+        """Step the env with an already-decoded parameter dict; bypasses index decoding."""
+        return self._step_with_params(params)
```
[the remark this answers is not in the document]
rutayan-nv: Done — `9937ef33`.

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Any, ClassVar, Optional
+
+from .base_gym import BaseGym
+
+_GYMNASIUM_INSTALL_HINT = "gymnasium is required for GymnasiumAdapter. Install it with: pip install gymnasium"
```
podkidyshev, line 23:
```suggestion
_GYMNASIUM_INSTALL_HINT = "gymnasium is required for GymnasiumAdapter. Install it with: uv sync --extra rl"
```

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _import_gymnasium():
```
podkidyshev, line 26: please introduce/move it into `src/cloudai/util/lazy_imports.py`. numpy is already there, just add gymnasium and maybe optionally `spaces` as a separate property

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+    ``gymnasium`` and ``numpy`` are optional dependencies; importing this module
+    is cheap, but instantiating the adapter without them raises ``ImportError``.
+    """
+
+    metadata: ClassVar[dict[str, Any]] = {"render_modes": ["human"]}
+
+    def __init__(self, env: BaseGym) -> None:
+        _, spaces, np = _import_gymnasium()
+
+        self._np = np
```
podkidyshev, line 73: please do not keep module as a variable, use it when needed from `lazy.np.<smth>` with `from cloudai.util.lazy_imports import lazy`

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+        Mirror ``handle_dse_job``'s 1-based ``test_run.step`` so artifact paths match.
+
+        The first step is written under ``…/<iteration>/1/``, matching how
+        ``handle_dse_job`` numbers steps; this keeps reports and trajectory
+        analysis consistent regardless of whether the env is driven by the
+        DSE loop or by an external training loop wrapping the adapter.
+        """
+        test_run = getattr(self._env, "test_run", None)
+        if test_run is not None:
+            test_run.step = self._step_count + 1
```
podkidyshev, line 162: who's responsible for steps budget now? for standard argents it's `handle_dse_job` loop with `agent.max_steps`

podkidyshev: apparently this is the only places that anyhow touches cloudai internals. if we solve it somehow - I suggest the adapter is to be put not in cloudai but in the private repo.

if the only consumer is our private RL agents, I’d prefer this live with those agents, while CloudAI exposes only the minimal env/execution hooks they need. Moving it to core makes Gymnasium compatibility look like a supported CloudAI API, but we do not yet have the integration tests or API ownership story for that

podkidyshev (changes requested):

rutayan-nv:
Closing unmerged. Superseded by a forthcoming cloudaix-side PR that introduces `GymnasiumAdapter` in its only real consumer (RL agents in `cloudaix.utils.configurator`).

## Why move to cloudaix

A live cluster run on ilyk surfaced that this adapter was mutating `test_run.step` from `reset()` and `step()`, which — combined with RLlib's contextual-bandit reset pattern (`agent_steps=1`) — collapsed every trial onto `step=1`, corrupting `trajectory.csv` and `env.csv`.

The fix is architectural: `TestRun.step` has a single owner now (`CloudAIGymEnv.step()`, via `TestRun.increment_step()` — see PR #893). Once the adapter stopped mutating `test_run.step`, the only cloudai-coupled thing it touched was `BaseGym` (input) and `gymnasium.Env` (output) — both stable interfaces. Nothing about it required living in cloudai. The only consumer is cloudaix's RL agent stack.

Closing rather than merging:
1. Avoids landing the buggy step-mutation code on `nv/main` even briefly.
2. Lets the cleaned-up version (no step mutation, plus negative-invariant contract tests asserting the adapter must not write `test_run.step`) land in its proper home in one move.
3. Keeps `cloudai/configurator/` minimal — only generic infrastructure (`BaseGym`, `CloudAIGymEnv`, `BaseAgent`), not RL-specific glue.

The architectural split going forward:
- **cloudai** owns the trial-counter contract (`TestRun`, `CloudAIGymEnv.step()`) and the dispatch contract (`BaseAgent.run()` polymorphism — PR #893).
- **cloudaix** owns RL-specific glue (`GymnasiumAdapter`, RLlib agent classes, custom `run()` overrides).

The cloudaix-side PR will link back here. Apologies for the churn on review comments — happy to address the substantive ones (action/param key validation, fail-fast on space mismatches) in the new PR.

PR #901, opened 2026-05-26
fix(configurator): make env_params first-class to fix the trajectory cache key

rutayan-nv:
## Issue
- Cache correctness: cache keyed on `action` only → revisiting an action under a different `env_params` sample serves the stale reward, corrupting the signal (the bug #900 pins).
- Artifact alignment: `env.csv` lived in workload code and was skipped on cache hits — sparse and misaligned with `trajectory.csv`.

## Fix
- Cache correctness: `env_params` becomes first-class schema (`TestDefinition.env_params`, `TestRun.current_env_params`, `TrajectoryEntry.env_params`); the cache key is now `(action, current_env_params)`.
- Artifact alignment: `CloudAIGymEnv` builds an `EnvParamsObserver` that samples deterministically and writes `env.csv` before the cache lookup, so it aligns 1:1 with `trajectory.csv`; workloads without `env_params` pay zero overhead.

## Testing
- #900's 4 TDD-red tests now pass; +9 unit (`test_env_params.py`) +2 integration (`test_cloudaigym.py`); full `pytest tests`: 1641 passed, 4 skipped; ruff + pyright clean.

**Stack:** `main` ← #893 ← #900 ← **#901** (top). Depends on #900 → #893.

### DIFF
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -155,7 +155,7 @@ class TestRun:
             return False
 
         agent = Registry().agents_map.get(self.test.agent)
-        return self.is_dse_job and (agent is None or agent.samples_env_params)
+        return self.is_dse_job and (agent is None or agent.supports_variable_environment)
 
     @property
     def nnodes(self) -> int:
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -203,10 +203,7 @@ class TestRun:
             else:
                 setattr(obj, attrs[-1], value)
 
-        # The agent's chosen action and this trial's sampled env_params are both just concrete
-        # values written onto cmd_args, applied through the one path here. env_params keys name
-        # cmd_args fields; sampling (the RNG) happens in the env before this call, so this method
-        # stays deterministic.
+        # RNG runs in the env before this call; applying only concrete values keeps this deterministic.
         for key, value in action.items():
             _apply(key, value)
         for key, value in (env_params or {}).items():
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -97,7 +97,6 @@ class TestRun:
     reports: Set[Type[ReportGenerationStrategy]] = field(default_factory=set)
     extra_srun_args: str | None = None
     num_nodes_explicit: bool = False
-    current_env_params: dict[str, Any] = field(default_factory=dict)
 
     def __hash__(self) -> int:
         return hash(self.name + self.test.name + str(self.iterations) + str(self.current_iteration))
@@ -140,6 +139,24 @@ class TestRun:
     @property
     def is_dse_job(self) -> bool:
         return self.test.is_dse_job or isinstance(self.num_nodes, list)
+
+    @property
+    def is_domain_randomization_active(self) -> bool:
+        """
+        Whether this run will actually env-sample (domain-randomize) per trial.
+
+        True only when domain randomization is declared (``env_params`` present), the run is a DSE
+        job (so a per-trial loop exists - including a ``num_nodes`` sweep), and the agent opts into
+        sampling. An unknown agent is treated as opted-in so the dedicated agent-resolution error
+        surfaces instead of this one.
+        """
+        if not self.test.is_domain_randomization_enabled:
+            return False
+
+        from .registry import Registry
+
+        agent = Registry().agents_map.get(self.test.agent)
+        return self.is_dse_job and (agent is None or agent.samples_env_params)
 
     @property
     def nnodes(self) -> int:
@@ -190,24 +207,21 @@ class TestRun:
     def apply_params_set(self, action: dict[str, Any], env_params: dict[str, Any] | None = None) -> "TestRun":
         tdef = self.test.model_copy(deep=True)
 
-        def _apply(key: str, value: Any) -> None:
+        # RNG runs in the env before this call; applying only concrete values keeps this deterministic.
+        # action and env_params target disjoint keys, so a plain merge applies both in one pass.
+        full_action = action | (env_params or {})
+        for key, value in full_action.items():
             if key.startswith("extra_env_vars."):
                 tdef.extra_env_vars[key[len("extra_env_vars.") :]] = value
-                return
-            attrs = key.split(".")
-            obj = tdef.cmd_args
-            for attr in attrs[:-1]:
-                obj = obj[attr] if isinstance(obj, dict) else getattr(obj, attr)
-            if isinstance(obj, dict):
-                obj[attrs[-1]] = value
             else:
-                setattr(obj, attrs[-1], value)
-
-        # RNG runs in the env before this call; applying only concrete values keeps this deterministic.
-        for key, value in action.items():
-            _apply(key, value)
-        for key, value in (env_params or {}).items():
-            _apply(key, value)
+                attrs = key.split(".")
+                obj = tdef.cmd_args
+                for attr in attrs[:-1]:
+                    obj = obj[attr] if isinstance(obj, dict) else getattr(obj, attr)
+                if isinstance(obj, dict):
+                    obj[attrs[-1]] = value
+                else:
+                    setattr(obj, attrs[-1], value)
 
         # env_params is validated at parse time; after the overlay its target cmd_args fields hold
         # concrete scalar draws, so re-validating it here would reject weighted specs. Drop it for
@@ -220,7 +234,6 @@ class TestRun:
         new_tr.test = tdef
         if "NUM_NODES" in action:
             new_tr.num_nodes = action["NUM_NODES"]
-        new_tr.current_env_params = dict(env_params or {})
         return new_tr
 
 
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -157,7 +157,9 @@ class TestRun:
             **{
                 key: value
                 for key, value in cmd_args_dict.items()
-                if isinstance(value, list) and not self.test.is_dse_excluded_arg(key)
+                if isinstance(value, list)
+                and not self.test.is_dse_excluded_arg(key)
+                and key.split(".", 1)[0] not in self.test.env_params
             },
             **{f"extra_env_vars.{key}": value for key, value in extra_env_vars_dict.items() if isinstance(value, list)},
         }
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, List, Optional, Set, Type, TypeAlias, Union
 
 from ..util import flatten_dict
+from .registry import Registry
 from .system import System
 
 if TYPE_CHECKING:
@@ -152,8 +153,6 @@ class TestRun:
         """
         if not self.test.is_domain_randomization_enabled:
             return False
-
-        from .registry import Registry
 
         agent = Registry().agents_map.get(self.test.agent)
         return self.is_dse_job and (agent is None or agent.samples_env_params)
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -97,6 +97,7 @@ class TestRun:
     reports: Set[Type[ReportGenerationStrategy]] = field(default_factory=set)
     extra_srun_args: str | None = None
     num_nodes_explicit: bool = False
+    current_env_params: dict[str, Any] = field(default_factory=dict)
 
     def __hash__(self) -> int:
         return hash(self.name + self.test.name + str(self.iterations) + str(self.current_iteration))
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -159,7 +159,7 @@ class TestRun:
                 for key, value in cmd_args_dict.items()
                 if isinstance(value, list)
                 and not self.test.is_dse_excluded_arg(key)
-                and key.split(".", 1)[0] not in self.test.env_params
+                and not self.test.is_env_sampled(key)
             },
             **{f"extra_env_vars.{key}": value for key, value in extra_env_vars_dict.items() if isinstance(value, list)},
         }
@@ -187,20 +187,30 @@ class TestRun:
 
         return all_combinations
 
-    def apply_params_set(self, action: dict[str, Any]) -> "TestRun":
+    def apply_params_set(self, action: dict[str, Any], env_params: dict[str, Any] | None = None) -> "TestRun":
         tdef = self.test.model_copy(deep=True)
-        for key, value in action.items():
+
+        def _apply(key: str, value: Any) -> None:
             if key.startswith("extra_env_vars."):
                 tdef.extra_env_vars[key[len("extra_env_vars.") :]] = value
+                return
+            attrs = key.split(".")
+            obj = tdef.cmd_args
+            for attr in attrs[:-1]:
+                obj = obj[attr] if isinstance(obj, dict) else getattr(obj, attr)
+            if isinstance(obj, dict):
+                obj[attrs[-1]] = value
             else:
-                attrs = key.split(".")
-                obj = tdef.cmd_args
-                for attr in attrs[:-1]:
-                    obj = obj[attr] if isinstance(obj, dict) else getattr(obj, attr)
-                if isinstance(obj, dict):
-                    obj[attrs[-1]] = value
-                else:
-                    setattr(obj, attrs[-1], value)
+                setattr(obj, attrs[-1], value)
+
+        # The agent's chosen action and this trial's sampled env_params are both just concrete
+        # values written onto cmd_args, applied through the one path here. env_params keys name
+        # cmd_args fields; sampling (the RNG) happens in the env before this call, so this method
+        # stays deterministic.
+        for key, value in action.items():
+            _apply(key, value)
+        for key, value in (env_params or {}).items():
+            _apply(key, value)
 
         type(tdef)(**tdef.model_dump())  # trigger validation
 
@@ -208,6 +218,7 @@ class TestRun:
         new_tr.test = tdef
         if "NUM_NODES" in action:
             new_tr.num_nodes = action["NUM_NODES"]
+        new_tr.current_env_params = dict(env_params or {})
         return new_tr
 
 
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -209,7 +209,12 @@ class TestRun:
         for key, value in (env_params or {}).items():
             _apply(key, value)
 
-        type(tdef)(**tdef.model_dump())  # trigger validation
+        # env_params is validated at parse time; after the overlay its target cmd_args fields hold
+        # concrete scalar draws, so re-validating it here would reject weighted specs. Drop it for
+        # this validation-only pass, which exists to validate the applied action values.
+        validation_args = tdef.model_dump()
+        validation_args.pop("env_params", None)
+        type(tdef)(**validation_args)  # trigger validation
 
         new_tr = copy.deepcopy(self)
         new_tr.test = tdef
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -300,17 +300,24 @@ def _check_installation(
 
 def validate_dse_env_params(test_scenario: TestScenario) -> None:
     """
-    Reject prepped configs that declare env_params on a non-DSE test run.
+    Reject prepped configs that declare env_params but cannot sample them.
 
-    env_params are sampled only during DSE (by CloudAIGymEnv); on a non-DSE run they would be
-    silently ignored. is_dse_job is a property of the fully prepped config, so this is validated
-    here rather than at parse time.
+    env_params are sampled per-trial by CloudAIGymEnv, but only for a DSE run on a *learning*
+    agent. A non-DSE run has no per-trial loop at all, and grid_search exhaustively enumerates the
+    search space - sampling env_params there would just inject noise into a deterministic sweep with
+    no policy to gain robustness. is_dse_job is a property of the fully prepped config, so this is
+    validated here rather than at parse time.
     """
-    offenders = [tr.name for tr in test_scenario.test_runs if tr.test.env_params and not tr.is_dse_job]
+    offenders = [
+        tr.name
+        for tr in test_scenario.test_runs
+        if tr.test.env_params and (not tr.is_dse_job or tr.test.agent == "grid_search")
+    ]
     if offenders:
         raise TestScenarioParsingError(
-            f"Tests {offenders} declare env_params but are not DSE jobs. env_params are sampled only during "
-            "DSE (by CloudAIGymEnv); add a sweep (a list-valued cmd_args/extra_env_vars entry or num_nodes) "
+            f"Tests {offenders} declare env_params but will not sample them. env_params are sampled per-trial "
+            "only by a DSE run on a learning agent (grid_search searches the whole space and ignores them). "
+            "Use a learning agent and add a sweep (a list-valued cmd_args/extra_env_vars entry or num_nodes), "
             "or remove env_params."
         )
 
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -27,6 +27,7 @@ from unittest.mock import Mock
 import toml
 import yaml
 
+from cloudai.configurator.env_params import validate_dse_env_params
 from cloudai.core import (
     BaseInstaller,
     CloudAIGymEnv,
@@ -296,30 +297,6 @@ def _check_installation(
         result = installer.is_installed(installables)
 
     return result
-
-
-def validate_dse_env_params(test_scenario: TestScenario) -> None:
-    """
-    Reject prepped configs that declare env_params but cannot sample them.
-
-    env_params are sampled per-trial by CloudAIGymEnv, but only for a DSE run on a *learning*
-    agent. A non-DSE run has no per-trial loop at all, and grid_search exhaustively enumerates the
-    search space - sampling env_params there would just inject noise into a deterministic sweep with
-    no policy to gain robustness. is_dse_job is a property of the fully prepped config, so this is
-    validated here rather than at parse time.
-    """
-    offenders = [
-        tr.name
-        for tr in test_scenario.test_runs
-        if tr.test.env_params and (not tr.is_dse_job or tr.test.agent == "grid_search")
-    ]
-    if offenders:
-        raise TestScenarioParsingError(
-            f"Tests {offenders} declare env_params but will not sample them. env_params are sampled per-trial "
-            "only by a DSE run on a learning agent (grid_search searches the whole space and ignores them). "
-            "Use a learning agent and add a sweep (a list-valued cmd_args/extra_env_vars entry or num_nodes), "
-            "or remove env_params."
-        )
 
 
 def handle_dry_run_and_run(args: argparse.Namespace) -> int:
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -39,6 +39,7 @@ from cloudai.core import (
     System,
     TestParser,
     TestScenario,
+    TestScenarioParsingError,
 )
 from cloudai.models.scenario import ReportConfig
 from cloudai.models.workload import TestDefinition
@@ -297,11 +298,34 @@ def _check_installation(
     return result
 
 
+def validate_dse_env_params(test_scenario: TestScenario) -> None:
+    """
+    Reject prepped configs that declare env_params on a non-DSE test run.
+
+    env_params are sampled only during DSE (by CloudAIGymEnv); on a non-DSE run they would be
+    silently ignored. is_dse_job is a property of the fully prepped config, so this is validated
+    here rather than at parse time.
+    """
+    offenders = [tr.name for tr in test_scenario.test_runs if tr.test.env_params and not tr.is_dse_job]
+    if offenders:
+        raise TestScenarioParsingError(
+            f"Tests {offenders} declare env_params but are not DSE jobs. env_params are sampled only during "
+            "DSE (by CloudAIGymEnv); add a sweep (a list-valued cmd_args/extra_env_vars entry or num_nodes) "
+            "or remove env_params."
+        )
+
+
 def handle_dry_run_and_run(args: argparse.Namespace) -> int:
     setup_result = _setup_system_and_scenario(args)
     if setup_result is None:
         return 1
     system, test_scenario, tests = setup_result
+
+    try:
+        validate_dse_env_params(test_scenario)
+    except TestScenarioParsingError as e:
+        logging.error(str(e))
+        return 1
 
     if not _handle_single_sbatch(args, system):
         return 1
@@ -491,7 +515,8 @@ def verify_test_scenarios(
             tests = Parser.parse_tests(test_tomls, system)
             hook_tests = Parser.parse_tests(hook_test_tomls, system)
             hooks = Parser.parse_hooks(hook_tomls, system, {t.name: t for t in hook_tests})
-            Parser.parse_test_scenario(scenario_file, system, {t.name: t for t in tests}, hooks)
+            scenario = Parser.parse_test_scenario(scenario_file, system, {t.name: t for t in tests}, hooks)
+            validate_dse_env_params(scenario)
         except Exception:
             nfailed += 1
 
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -135,8 +135,7 @@ def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
         return 1
 
     err = 0
-    # Recoverable failures return a non-zero rc and are accumulated here; an unexpected exception
-    # (a bug) is a hard-fail. We capture it so reports still generate, then re-raise below.
+    # Capture an unexpected error so reports still generate, then re-raise below.
     run_error: Exception | None = None
     try:
         for tr in runner.runner.test_scenario.test_runs:
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -27,7 +27,7 @@ from unittest.mock import Mock
 import toml
 import yaml
 
-from cloudai.configurator.env_params import validate_dse_env_params
+from cloudai.configurator.env_params import validate_domain_randomization_active
 from cloudai.core import (
     BaseInstaller,
     CloudAIGymEnv,
@@ -305,7 +305,7 @@ def handle_dry_run_and_run(args: argparse.Namespace) -> int:
     system, test_scenario, tests = setup_result
 
     try:
-        validate_dse_env_params(test_scenario)
+        validate_domain_randomization_active(test_scenario)
     except TestScenarioParsingError as e:
         logging.error(str(e))
         return 1
@@ -499,7 +499,7 @@ def verify_test_scenarios(
             hook_tests = Parser.parse_tests(hook_test_tomls, system)
             hooks = Parser.parse_hooks(hook_tomls, system, {t.name: t for t in hook_tests})
             scenario = Parser.parse_test_scenario(scenario_file, system, {t.name: t for t in tests}, hooks)
-            validate_dse_env_params(scenario)
+            validate_domain_randomization_active(scenario)
         except Exception:
             nfailed += 1
 
--- a/src/cloudai/configurator/base_agent.py
+++ b/src/cloudai/configurator/base_agent.py
@@ -58,6 +58,11 @@ class BaseAgent(ABC):
     Provides a unified interface and parameter management for action spaces.
     """
 
+    # Opt-in capability: only agents whose search consumes per-trial env_params sampling set this
+    # to True. Defaults False so a config that declares env_params for an agent that would ignore
+    # them is rejected up front instead of silently having no effect.
+    samples_env_params: bool = False
+
     def __init__(self, env: BaseGym, config: BaseAgentConfig):
         """
         Initialize the agent with the environment.
@@ -94,9 +99,8 @@ class BaseAgent(ABC):
 
         Args:
             observation: Latest observation produced by the environment (``env.reset()`` on the
-                first call, then the result of the prior ``env.step()``). Stateless agents such
-                as grid search or Bayesian optimization may ignore this; observation-conditioned
-                agents (RL, contextual bandits) should use it.
+                first call, then the result of the prior ``env.step()``). Stateless agents may
+                ignore this; observation-conditioned agents should use it.
 
         Returns:
             Tuple[int, Dict[str, Any]] | None: The current step index and a dictionary mapping action keys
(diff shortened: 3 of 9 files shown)


rutayan-nv:
@coderabbitai review

rutayan-nv:
@coderabbitai review

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+
+        The first step is written under ``…/<iteration>/1/``, matching how
+        ``handle_dse_job`` numbers steps; this keeps reports and trajectory
+        analysis consistent regardless of whether the env is driven by the
+        DSE loop or by an external training loop wrapping the adapter.
+        """
+        test_run = getattr(self._env, "test_run", None)
+        if test_run is not None:
+            test_run.step = self._step_count + 1
+
```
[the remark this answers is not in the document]
rutayan-nv: Outdated/addressed. The `_sync_underlying_step_counter` / `_step_count` approach this flagged has been removed. `gymnasium_adapter.py` now lives in #930 (gymnasium-adapter) and is a **pure pass-through that never mutates `test_run.step`** — documented as a design invariant and pinned by contract tests — so `CloudAIGymEnv.step()`'s own entry-increment keeps trials 1-based. No off-by-one remains. Resolving.

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+        _, spaces, np = _import_gymnasium()
+
+        self._np = np
+        self._env = env
+        self._step_count = 0
+
+        raw_action_space = env.define_action_space()
+        self._tunable_params: dict[str, list] = {k: v for k, v in raw_action_space.items() if len(v) > 1}
+        self._fixed_params: dict[str, Any] = {k: v[0] for k, v in raw_action_space.items() if len(v) == 1}
+
```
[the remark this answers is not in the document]
rutayan-nv: Outdated. `gymnasium_adapter.py` / `test_gymnasium_adapter.py` were relocated and redesigned in #930 (gymnasium-adapter) during the lazy-import split — the adapter is now a pure pass-through over `test_run.step`. This thread targets the prior design, which no longer exists in this PR's diff; #930 was re-reviewed with no open findings. Resolving.

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+        ``handle_dse_job`` numbers steps; this keeps reports and trajectory
+        analysis consistent regardless of whether the env is driven by the
+        DSE loop or by an external training loop wrapping the adapter.
+        """
+        test_run = getattr(self._env, "test_run", None)
+        if test_run is not None:
+            test_run.step = self._step_count + 1
+
+    def _as_obs_array(self, obs: Any) -> Any:
+        return self._np.asarray(obs, dtype=self._np.float32)
```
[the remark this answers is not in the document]
rutayan-nv: Outdated. `gymnasium_adapter.py` / `test_gymnasium_adapter.py` were relocated and redesigned in #930 (gymnasium-adapter) during the lazy-import split — the adapter is now a pure pass-through over `test_run.step`. This thread targets the prior design, which no longer exists in this PR's diff; #930 was re-reviewed with no open findings. Resolving.

tests/test_gymnasium_adapter.py
```diff
@@ -0,0 +1,234 @@
...
+        return {"param_a": [1, 2, 3], "param_b": [10, 20], "fixed_param": [42]}
+
+
+class _GymWithTestRun(_FakeGym):
+    """Carries a CloudAI-like ``test_run`` so we can verify the step-counter sync."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.test_run = SimpleNamespace(step=0)
+
```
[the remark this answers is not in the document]
rutayan-nv: Outdated. `gymnasium_adapter.py` / `test_gymnasium_adapter.py` were relocated and redesigned in #930 (gymnasium-adapter) during the lazy-import split — the adapter is now a pure pass-through over `test_run.step`. This thread targets the prior design, which no longer exists in this PR's diff; #930 was re-reviewed with no open findings. Resolving.

src/cloudai/configurator/base_agent.py
```diff
@@ -111,3 +112,38 @@ def update_policy(self, _feedback: Dict[str, Any]) -> None:
...
+                    "trial_index": step,
+                    "value": reward,
+                    "observation": observation,
+                    "prev_observation": prev_observation,
+                    "action": action,
+                    "done": done,
+                }
+            )
+            logging.info(f"Step {step}: Observation: {[round(obs, 4) for obs in observation]}, Reward: {reward:.4f}")
+        return 0
```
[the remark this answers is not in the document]
rutayan-nv: Done on #933 (88d35c5d) — `base_agent.py` is owned by that PR; cascaded here via restack. Both `logging.info` f-strings in `BaseAgent.run` are now %-style deferred logging. Resolving.

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,169 @@
...
+samples per trial (categorical, optional weights). It is sibling to
+``cmd_args`` on a ``TestDefinition`` and does not enter the agent's action
+space; the policy learns a robust mapping under that variation.
+
+This module owns the data schema (``EnvParamSpec``), the deterministic
+sampler (``EnvParamsSampler``), the persistence interface
+(``EnvParamsSink`` + ``CsvSink``) and the per-step observer
+(``EnvParamsObserver``). ``CloudAIGymEnv`` consumes these directly so the
+artifacts (``env.csv``) and the cache key align 1:1 with ``trajectory.csv``
+regardless of agent (PPO, BO, GA, MAB) or workload.
```
podkidyshev: please do not mention private implementation in the public repo 🙏

podkidyshev (changes requested):

src/cloudai/configurator/env_params.py
rutayan-nv: Scrubbed all private-implementation references from the docstrings and field descriptions in `env_params.py` and `workload.py`. They now describe the mechanism generically — no algorithm names or internal planning-doc references.

rutayan-nv:
Core owns the randomization and the `TestRun` update; it is not agent-specific.

**Real-world use.** Domain randomization: vary an uncontrolled environment condition per trial so the learned policy generalizes instead of overfitting one fixed setting. Example: the agent tunes `paddle_width` (action) while each trial randomizes `ball_speed` (env_param).

**Injection.** `env_params` is an annotation, not a value store. Candidate values live in `cmd_args`, same as the action space (`cmd_args.ball_speed = [4, 6, 8]`); `[env_params.ball_speed]` reclassifies that list from "search" to "sample." Per trial, `CloudAIGymEnv` samples one value (`EnvParamsObserver`), writes it to `TestRun.current_env_params`, and overlays it onto `cmd_args` via `model_copy(update=…)` before the workload runs.

**Core vs `agent_config`.** The agent never reads or sets `env_params`; it only sees the reward. Sampling and the `TestRun` write happen in core (cloudai's Gym environment). It is environment state, not an agent hyperparameter.

**vs. sweep.** Inverse of a sweep over the same list: a bare list in `cmd_args` is the action space (searched); the same list annotated `env_params` is sampled, not searched (excluded from `param_space`/`is_dse_job`). Hence the hard reject under `grid_search`.

**#900.** `current_env_params` is part of the trajectory cache key, so revisiting an action under a different sample no longer serves a stale reward.

src/cloudai/models/workload.py
```diff
@@ -191,3 +207,44 @@ def validate_agent_config(self) -> Self:
...
+
+        for name, spec in self.env_params.items():
+            if spec.weights is None:
+                continue
+            value = getattr(self.cmd_args, name, None)
+            if not isinstance(value, list) or len(value) < 2:
+                raise ValueError(
+                    f"env_params['{name}'] declares weights but cmd_args.{name} is not a candidate list "
+                    "(need a list of >= 2 values)"
+                )
```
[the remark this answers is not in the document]
rutayan-nv: Addressed in cf7dbf66: validate_env_params now rejects structured (dict/BaseModel) cmd_args targets with a clear error, plus a regression test (test_env_params_structured_target_rejected).

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -61,8 +63,28 @@ def __init__(self, test_run: TestRun, runner: BaseRunner, rewards: RewardOverrid
         self.max_steps = test_run.test.agent_steps
         self.reward_function = Registry().get_reward_function(test_run.test.agent_reward_function)
         self.trajectory: dict[int, list[TrajectoryEntry]] = {}
+        self._env_sink: EnvParamsSink | None = None
+        self.observers: List[StepObserver] = self._build_observers()
```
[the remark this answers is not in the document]
rutayan-nv: Addressed in cf7dbf66: write_trajectory rebinds the env.csv sink to the current iteration's path before each write, keeping env.csv 1:1 aligned with trajectory.csv when the env instance is reused across iterations.

src/cloudai/cli/handlers.py
```diff
@@ -132,64 +134,79 @@ def handle_dse_job(runner: Runner, args: argparse.Namespace) -> int:
...
+            agent_config = agent_class.get_config_class()(**agent_config_data)
+            env = CloudAIGymEnv(
+                test_run=test_run,
+                runner=runner.runner,
+                rewards=agent_config.rewards,
             )
-            err = 1
-            continue
+            if agent_config.start_action == "first":
+                logging.info(f"Using deterministic first sweep for the chosen agent: {env.first_sweep}.")
```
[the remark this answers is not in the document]
rutayan-nv: Skipping: these log lines are not part of this PR's diff (last touched by 228932d4 / 3d47fdef, now in main), and the repo's ruff config does not enable G004 (CI ruff check passes). Better left to a dedicated logging cleanup.

src/cloudai/cli/handlers.py
```diff
@@ -281,12 +298,42 @@ def _check_installation(
...
+    env_params are sampled per-trial by CloudAIGymEnv, but only for a DSE run on a *learning*
+    agent. A non-DSE run has no per-trial loop at all, and grid_search exhaustively enumerates the
+    search space - sampling env_params there would just inject noise into a deterministic sweep with
+    no policy to gain robustness. is_dse_job is a property of the fully prepped config, so this is
+    validated here rather than at parse time.
+    """
+    offenders = [
+        tr.name
+        for tr in test_scenario.test_runs
+        if tr.test.env_params and (not tr.is_dse_job or tr.test.agent == "grid_search")
```
podkidyshev: I suggest two things:

* do not hardcode `grid_search` here but instead disable sampling in the grid search agent somehow (have a flag on the env class or by inheritance somehow). This way we will move this behavior back to agents, not the core code
* make this validation a part of TestRun/TestScenario model (as a pydantic validator). It will fit there naturally rather than in the handlers.py here (it's already huge...)

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -251,8 +315,21 @@ def current_trajectory(self) -> list[TrajectoryEntry]:
...
+        """
+        Return a cached entry only when the full trial identity matches.
+
+        Trial identity is ``(action, env_params)``: env-randomized parameters
+        change the workload's behaviour, so a trial repeating the same action
+        under a different ``env_params`` sample must miss and re-run. Empty
+        env_params on both sides is the back-compat path for workloads that
+        do not declare any ``[env_params.*]`` block.
+        """
+        current_env_params = getattr(self.test_run, "current_env_params", {}) or {}
```
podkidyshev: `current_env_params` is a constant attribute or am I mistaken? I bet we could just access it directly as `self.test_run.current_env_params`

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,199 @@
...
+``weights``), never the values. The policy learns a robust mapping under that
+variation; the knob never enters the agent's action space.
+
+This module owns the annotation schema (``EnvParamSpec``), the deterministic
+sampler (``EnvParamsSampler``), the persistence interface
+(``EnvParamsSink`` + ``CsvSink``) and the per-step observer
+(``EnvParamsObserver``). ``CloudAIGymEnv`` consumes these directly so the
+artifacts (``env.csv``) and the cache key align 1:1 with ``trajectory.csv``
+regardless of agent or workload.
+"""
```
podkidyshev, line 26: I'd vote for shortening most of the comments in this PR. E.g. the 2nd paragraph here simply lists what in the module = no added value

As a tip, I bet AI with caveman skill can produce decent but short notes

Also I believe we should make a dedicated page in the `docs/` AFTER the initiative is fully delivered replacing most of the in-code comments. It will provide a high-level overview on all the RL-features added

what do you think?

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,199 @@
...
+            weights = self._weights.get(name)
+            if weights is not None:
+                out[name] = rng.choices(choices, weights=weights, k=1)[0]
+            else:
+                out[name] = rng.choice(choices)
+        return out
+
+
+@runtime_checkable
+class EnvParamsSink(Protocol):
```
podkidyshev: rather a nitpick: you could simply define this class as abstract + `write` method as abstract an inherit CsvSink from EnvParamsSink

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,199 @@
...
+        for w in self.weights:
+            if not math.isfinite(w) or w < 0:
+                raise ValueError(f"env_params weights must be finite and non-negative; got {w}")
+        total = sum(self.weights)
+        if abs(total - 1.0) > 1e-6:
+            raise ValueError(f"env_params weights must sum to 1.0; got {total}")
+        return self
+
+
+class EnvParamsSampler:
```
podkidyshev: 1. used just once only in this module
2. contains only a single method

1 + 2 = function :D

podkidyshev: essentially this class is a `random.choices` wrapper with a loop over it yet it takes 34 lines with all the comments. kinda an overkill to me

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,199 @@
...
+            candidates[name] = value
+            if spec.weights is not None:
+                weights[name] = spec.weights
+        self._sampler = EnvParamsSampler(candidates, seed=seed, weights=weights)
+
+    def before_step(self, test_run: Any) -> None:
+        test_run.current_env_params = self._sampler.sample(test_run.step)
+
+    def after_step(self, test_run: Any, observation: list, reward: float) -> None:
+        del test_run, observation, reward  # persistence handled by CloudAIGymEnv.write_trajectory
```
podkidyshev: do we need `after_step` at all then if it's just a noop?

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,199 @@
...
+            value = getattr(cmd_args, name, None)
+            if not isinstance(value, list):
+                continue  # scalar cmd_args knob is fixed; the annotation is a no-op
+            candidates[name] = value
+            if spec.weights is not None:
+                weights[name] = spec.weights
+        self._sampler = EnvParamsSampler(candidates, seed=seed, weights=weights)
+
+    def before_step(self, test_run: Any) -> None:
+        test_run.current_env_params = self._sampler.sample(test_run.step)
```
podkidyshev: aside the noop `after_step` - this `before_step` stays as the only meaning of the class = function. overall this class is responsible for applying specific knobs onto a test run to set up an experiment. wouldn't it make sense to move this chunk of code into the existing code that does it - `TestRun.apply_params_set`?

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -61,8 +63,28 @@ def __init__(self, test_run: TestRun, runner: BaseRunner, rewards: RewardOverrid
...
+        an empty mapping yields no observers and zero overhead.
+        """
+        observers: List[StepObserver] = []
+        if self.test_run.test.env_params:
+            seed = int((self.test_run.test.agent_config or {}).get("random_seed", 0))
+            self._env_sink = CsvSink(self._env_csv_path())
+            observers.append(EnvParamsObserver(self.test_run.test.env_params, self.test_run.test.cmd_args, seed))
+        return observers
+
+    def _env_csv_path(self) -> Path:
```
podkidyshev: nit: either make it a property or add `get` suffix so that 🙏

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -61,8 +63,28 @@ def __init__(self, test_run: TestRun, runner: BaseRunner, rewards: RewardOverrid
...
+        """
+        Construct the per-step observers implied by the TestDefinition.
+
+        Workloads opt in to env_params via a TOML ``[env_params.<name>]`` block;
+        an empty mapping yields no observers and zero overhead.
+        """
+        observers: List[StepObserver] = []
+        if self.test_run.test.env_params:
+            seed = int((self.test_run.test.agent_config or {}).get("random_seed", 0))
+            self._env_sink = CsvSink(self._env_csv_path())
```
podkidyshev: `self._env_sink = CsvSink(self._env_csv_path())` - this is duplicated two times in the class

src/cloudai/configurator/env_params.py
podkidyshev: it should also take entire code block in `CloudAIGymEnv.step` that starts with `if self.test_run.current_env_params:` because it is essentially part of applying knobs onto a TestRun per experiment

src/cloudai/models/workload.py
```diff
@@ -136,17 +146,23 @@ def constraint_check(self, tr: TestRun, system: Optional[System]) -> bool:
...
+        def check_dict(d: dict, parent_key: str = "", skip_env_params: bool = False) -> bool:
             if isinstance(d, dict):
                 for key, value in d.items():
                     path = f"{parent_key}.{key}" if parent_key else key
                     if self.is_dse_excluded_arg(path):
                         continue
-                    if isinstance(value, list) or (isinstance(value, dict) and check_dict(value, path)):
+                    # env_params lists are sampled by the env, not searched by the agent, so they
+                    # are not action-space dimensions and must not make a run count as DSE.
+                    if skip_env_params and path.split(".", 1)[0] in self.env_params:
```
podkidyshev: I keep seeing this `path.split(".", 1)[0]` in regards of `env_params`. Does it really need to be dot-separated in all of those places?

moreover, some workloads declare more than just the root cmd_args to contain CLIs. e.g. vllm has `bench_cmd_args` just as well. at the same time certain feature are truly constrained to support only `cmd_args`

What I'm saying, I suggest the format of `env_params` is the full dot-separated path to a given attributed inside TestDefinition, not CmdArgs. I believe it will look cleaner in the toml:

```toml
num_nodes = 2

[cmd_args]
  foo = 1

  [cmd_args.args]
  bar = [1, 2, 3]

[env_params]
"cmd_args.bar" = [0.1, 0.1, 0.8]
```

podkidyshev (changes requested):

podkidyshev:
Overall arch suggestion (summary of some of my comments): **_move parsing and applying of `env_params` closer to where the regular sweeping knobs are processed - TestRun, TestDefinition classes._** So that, we reuse most of the code related to sweeping/knobs in regards of how it works with the TestDefinition and CmdArgs body

src/cloudai/configurator/env_params.py
rutayan-nv: I like the idea. I agree about a dedicated page. We have it for other agents and thus, we should have it for RL as well.

src/cloudai/configurator/env_params.py
rutayan-nv: Let me take a look.

src/cloudai/configurator/env_params.py
rutayan-nv: done - d81001cf

src/cloudai/cli/handlers.py
rutayan-nv: done - 99d7bf22 (capability flag `BaseAgent.samples_env_params` + validator moved out of `handlers.py`; kept it a post-prep function rather than a pydantic validator, since `is_dse_job` only resolves on the fully-prepped config and `TestRun` is a dataclass)

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: done - f25ef4d3

src/cloudai/configurator/env_params.py
rutayan-nv: done - f25ef4d3 (dropped the module-listing paragraph; the dedicated `docs/` RL page stays for after the initiative lands, as discussed)

src/cloudai/configurator/env_params.py
rutayan-nv: done - e1c9769a (collapsed to a single concrete `EnvParamsSink` instead, since there's still only one writer format)

src/cloudai/configurator/env_params.py
rutayan-nv: done - f25ef4d3 (removed `EnvParamsSampler`; folded into `EnvParam.draw` / `EnvParams.sample`)

src/cloudai/configurator/env_params.py
rutayan-nv: done - f25ef4d3 (removed `StepObserver` / `after_step`)

src/cloudai/configurator/env_params.py
rutayan-nv: done - f25ef4d3 (removed `EnvParamsObserver`; knob application now lives in `TestRun.apply_params_set`, including the `current_env_params` block from `step`)

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: done - e1c9769a (now the `env_params_record_path` property)

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: done - e1c9769a (sink built once in `__init__`; no duplication)

src/cloudai/models/workload.py
rutayan-nv: done - f25ef4d3 (removed the redundant `split('.', 1)[0]`; env_params keys are flat `cmd_args` field names. Kept scope to `cmd_args` for now rather than full `TestDefinition` dot-paths — can generalize to `bench_cmd_args` etc. in a follow-up if you'd prefer)

rutayan-nv:
> Overall arch suggestion (summary of some of my comments): **_move parsing and applying of `env_params` closer to where the regular sweeping knobs are processed - TestRun, TestDefinition classes._** So that, we reuse most of the code related to sweeping/knobs in regards of how it works with the TestDefinition and CmdArgs body



done - d81001cf, f25ef4d3.

- Parsing: `env_params` is a `TestDefinition` field; candidates live in `cmd_args` and `validate_env_params` validates against them. `is_dse_job` / `is_env_sampled` traverse `cmd_args` in the same pass as sweep detection.
- Applying: `TestRun.apply_params_set` writes the agent action and the sampled `env_params` through the same `_apply` onto `cmd_args`; `param_space` reuses the same `cmd_args` list-values, excluding env-sampled keys.
- Per-trial sampling stays in the env layer (`CloudAIGymEnv.step`), keeping `apply_params_set` deterministic.

src/cloudai/_core/test_scenario.py
```diff
@@ -179,27 +187,38 @@ def all_combinations(self) -> list[dict[str, Any]]:
...
+        # The agent's chosen action and this trial's sampled env_params are both just concrete
+        # values written onto cmd_args, applied through the one path here. env_params keys name
+        # cmd_args fields; sampling (the RNG) happens in the env before this call, so this method
+        # stays deterministic.
+        for key, value in action.items():
+            _apply(key, value)
+        for key, value in (env_params or {}).items():
+            _apply(key, value)
 
         type(tdef)(**tdef.model_dump())  # trigger validation
```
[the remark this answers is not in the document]
rutayan-nv: done - 5790ab9f

src/cloudai/_core/test_scenario.py
```diff
@@ -184,27 +187,40 @@ def all_combinations(self) -> list[dict[str, Any]]:
...
+        # concrete scalar draws, so re-validating it here would reject weighted specs. Drop it for
+        # this validation-only pass, which exists to validate the applied action values.
+        validation_args = tdef.model_dump()
+        validation_args.pop("env_params", None)
+        type(tdef)(**validation_args)  # trigger validation
 
         new_tr = copy.deepcopy(self)
         new_tr.test = tdef
         if "NUM_NODES" in action:
             new_tr.num_nodes = action["NUM_NODES"]
```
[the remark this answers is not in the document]
rutayan-nv: Confirmed and valid — but deliberately not fixing inline, to keep this PR focused.

The phantom `NUM_NODES` is a pre-existing record/repro-integrity issue: it lands in `cmd_args` via `extra="allow"` but never reaches the generated shell command (strategies read named fields, not the extras bag), and `TestRun.num_nodes` is still set correctly. The clean fix is a small refactor that makes `_apply` the single dispatch point and shares a `NUM_NODES_KEY` constant between `param_space` (producer) and `apply_params_set` (consumer) so the one scope-crossing key routes to `TestRun.num_nodes`.

Tracked separately: NVIDIA/cloudai#943.

src/cloudai/cli/handlers.py
```diff
@@ -303,6 +304,12 @@ def handle_dry_run_and_run(args: argparse.Namespace) -> int:
         return 1
     system, test_scenario, tests = setup_result
 
+    try:
+        validate_dse_env_params(test_scenario)
+    except TestScenarioParsingError as e:
+        logging.error(str(e))
+        return 1
+
```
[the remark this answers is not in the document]
rutayan-nv: Good catch on the symptom, but the inline guard would bake in the wrong coupling: `--single-sbatch` is a scheduling concern and should not gate env / action-space semantics. It is only happenstance that it is currently used for grid unroll; it could later parallelize agent-driven runs (e.g. a genetic algorithm launching multiple jobs at once).

The real fix is a single "is this run agent-driven?" source of truth that resolves env_params in every execution mode, kept orthogonal to `single-sbatch`. Filed to root-cause the design flaw rather than patch it here: NVIDIA/cloudai#944.

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,197 @@
...
+
+    offenders = []
+    for tr in test_scenario.test_runs:
+        if not tr.test.env_params:
+            continue
+
+        agent = agents.get(tr.test.agent)
+        # Unknown agent: defer to the dedicated agent-resolution error rather than masking it here.
+        sampled = tr.is_dse_job and (agent is None or agent.samples_env_params)
+        if not sampled:
```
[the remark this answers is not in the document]
rutayan-nv: Fixed in 8f91664d — but with the opposite resolution to the one suggested.

Rather than silently skipping scalar-only annotations, we now **reject** them at parse time. Annotating a scalar `cmd_args` field is a meaningless label: an `env_params` entry only reclassifies a *list-valued* sweep as env-sampled, and a scalar is already fixed. So `TestDefinition.validate_env_params` now raises with a clear "make `cmd_args.<key>` a list, or remove the annotation" message.

This eliminates the inconsistency you flagged at the source — there is no longer a mode-dependent accept/reject, because such a config never validates in any mode. The post-overlay path already drops `env_params` before re-validating, so concrete per-trial scalar draws are unaffected.

Follow-up: NVIDIA/cloudai#945 to give these parse errors rust-style, line-located diagnostics.

src/cloudai/_core/test_scenario.py
```diff
@@ -184,27 +187,40 @@ def all_combinations(self) -> list[dict[str, Any]]:
 
         return all_combinations
 
-    def apply_params_set(self, action: dict[str, Any]) -> "TestRun":
+    def apply_params_set(self, action: dict[str, Any], env_params: dict[str, Any] | None = None) -> "TestRun":
         tdef = self.test.model_copy(deep=True)
-        for key, value in action.items():
+
+        def _apply(key: str, value: Any) -> None:
```
podkidyshev: i'd suggest reverting the changes with the `_apply` function (local functions are a troublesome thing during debugging) and rather replace the previous root 

```python
for key, value in action.items():
```

with

```python
full_action = action | (env_params or {})
for key, value in full_action.items():
```

it makes the delta of the PR quite shorter + no inner function

src/cloudai/_core/test_scenario.py
```diff
@@ -184,27 +187,40 @@ def all_combinations(self) -> list[dict[str, Any]]:
...
+        # this validation-only pass, which exists to validate the applied action values.
+        validation_args = tdef.model_dump()
+        validation_args.pop("env_params", None)
+        type(tdef)(**validation_args)  # trigger validation
 
         new_tr = copy.deepcopy(self)
         new_tr.test = tdef
         if "NUM_NODES" in action:
             new_tr.num_nodes = action["NUM_NODES"]
+        new_tr.current_env_params = dict(env_params or {})
```
podkidyshev: nit

```suggestion
        new_tr.current_env_params = env_params or {}
```

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -134,6 +145,7 @@ def step(self, action: Any) -> Tuple[list, float, bool, dict]:
                     action=action,
                     reward=cached_result.reward,
                     observation=cached_result.observation,
+                    env_params=dict(self.test_run.current_env_params),
```
podkidyshev: nit

and I see this (to my understanding) redundant `dict(...)` call across entire PR. is there really a need for it?

```suggestion
                    env_params=self.test_run.current_env_params,
```

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -162,6 +174,9 @@ def step(self, action: Any) -> Tuple[list, float, bool, dict]:
             self.test_run.step = new_tr.step
             self.test_run.output_path = new_tr.output_path
 
+        # The test_run rebuild above drops the sample; restore it so the entry, cache key, and env.csv match.
+        self.test_run.current_env_params = new_tr.current_env_params
```
podkidyshev: I see the same line in `src/cloudai/_core/test_scenario.py:223`. If it's indeed a duplicate - I would vote for removing this one and keeping one in scenario

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -243,17 +266,36 @@ def write_trajectory(self, entry: TrajectoryEntry):
...
+        env_params on both sides is the back-compat path for workloads that
+        do not declare any ``[env_params.*]`` block.
+        """
+        current_env_params = self.test_run.current_env_params
         for entry in self.current_trajectory:
-            if self._values_match_exact(entry.action, action):
+            if not self._values_match_exact(entry.action, action):
+                continue
+            if self._values_match_exact(entry.env_params, current_env_params):
                 return entry
```
podkidyshev: nit again, but I'd like to push a bit for readibility. we sacrifice a few ms for validation but I believe it's just mush easier to read

```suggestion
            action_match = self._values_match_exact(entry.action, action)
            env_params_match = self._values_match_exact(entry.env_params, current_env_params)
            if action_match and env_params_match:
                return entry
```

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,198 @@
...
+    otherwise ``from_test`` returns ``None`` and the env carries no env-params state at all.
+    Owns the per-parameter :class:`EnvParam` draws (resolved from ``cmd_args``, the single source
+    of truth) and the seed, and draws one value per parameter per trial.
+    """
+
+    params: Dict[str, EnvParam]
+    seed: int
+
+    @classmethod
+    def from_test(cls, test: Any) -> Optional["EnvParams"]:
```
podkidyshev: is it really `Any`?

src/cloudai/models/workload.py
podkidyshev: Yes, I would do a follow-up later to support full paths `cmd_args.foo.bar` or `anything.abc.def` meaning any nested attribute starting from TestDefinition. So that all parsing feature act consistently 

just to be clear - don't need to be this PR but highly appreciated not to be prolonged. I can take it myself as well

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,198 @@
...
+        Draw this trial's value for each parameter.
+
+        Determinism: the same ``(seed, name, trial)`` yields the same draw across processes.
+        Independence: each parameter's RNG is seeded ``f"{seed}:{name}:{trial}"`` so adding or
+        removing one parameter never perturbs the others' draw sequences.
+        """
+        return {name: param.draw(random.Random(f"{self.seed}:{name}:{trial}")) for name, param in self.params.items()}
+
+
+class EnvParamsSink:
```
podkidyshev: redundant class definition: no state, no polymorphism

src/cloudai/configurator/env_params.py
```diff
@@ -0,0 +1,198 @@
...
+        new_file = not path.exists()
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with path.open("a", newline="") as f:
+            writer = csv.writer(f)
+            if new_file:
+                writer.writerow(("step", "env"))
+            writer.writerow([step, sample])
+
+
+def validate_dse_env_params(test_scenario: "TestScenario") -> None:
```
podkidyshev: I wonder if you could just merge this validation with `TestDefinition.validate_env_params` (eliminating this one at all and calling it explicitly from `handlers.py`)

podkidyshev (changes requested):

rutayan-nv:
### Follow-up gap: single-candidate (`[1]`) env_params still slip through

This PR rejects a **scalar** env_params target ("nothing to sample"), but a **single-element list** target (`cmd_args.osl = [1]` + `[env_params.osl]`) currently passes: `validate_env_params` only checks "is a non-empty list", not "≥ 2 candidates".

Semantically `[1]` is "sample from {1}" = always 1 = a constant draw — exactly the same no-op this PR is meant to kill, just spelled as a list. So it's the same class of bug leaking through.

Consistent fix: require an env_params target to be a candidate list of **≥ 2** values (a single candidate is a no-op, like a scalar). This also lines up with the parse-time canonicalization proposed in #946 (collapse non-env `[x]` → `x`): once that exists, a `[1]` env_param naturally reduces to the already-rejected scalar case.

Filing as a follow-up rather than expanding this PR's scope.

src/cloudai/_core/test_scenario.py
rutayan-nv: Done — `a067e357`. Reverted the `_apply` local function; `apply_params_set` now runs a single `full_action = action | (env_params or {})` loop, exactly as suggested.

src/cloudai/_core/test_scenario.py
rutayan-nv: Resolved, but differently — `a067e357`. I dropped the `TestRun.current_env_params` field entirely rather than assigning it. It stored transient per-trial state on a long-lived object, and the runner's fallback path could rebind `test_run` and silently lose the sample. The per-trial draw is now a local in `step()`, threaded explicitly into the cache lookup and `TrajectoryEntry` (mirroring `action`), so this assignment no longer exists.

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: Done — `a067e357`. Dropped the redundant `dict(...)` wrappers across the PR; the sample is already a plain dict passed by value.

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: Done — `a067e357`. Both copies are gone: removing `current_env_params` eliminated the gym-side restore (the duplicate you flagged) and the assignment in `apply_params_set`. No duplication left.

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: Done — `a067e357`. Split into `action_match` / `env_params_match` locals for readability, as suggested.

src/cloudai/configurator/env_params.py
rutayan-nv: Done — `a067e357`. Typed it `from_test(test: "TestDefinition")` — it was never really `Any`.

src/cloudai/models/workload.py
rutayan-nv: Tracked in #950 — generalize `env_params` keys to full `TestDefinition` dot-paths (`cmd_args.foo.bar`, `bench_cmd_args.*`, etc.) so all parsing features stay consistent. Deliberately out of scope for this PR per your note, but captured so it won't be forgotten.

src/cloudai/configurator/env_params.py
rutayan-nv: Done — `a067e357`. Collapsed `EnvParamsSink` into a module-level `write_env_params(path, step, sample)` function — no state, no polymorphism, so the class earned nothing.

src/cloudai/configurator/env_params.py
rutayan-nv: Done — `a067e357`, though not as a literal merge into `TestDefinition.validate_env_params`.

That validator is a pydantic field validator, so it fires at parse time. But whether `env_params` will actually be sampled depends on `is_dse_job` (incl. a `num_nodes` sweep) and the resolved agent's capability — both of which only exist on the fully-prepped `TestRun`, after parse. A parse-time validator structurally can't see them.

So I pushed the *logic* onto the models as two predicates and kept the function a thin caller:
- `TestDefinition.is_domain_randomization_enabled` — declared? (`bool(env_params)`)
- `TestRun.is_domain_randomization_active` — will it actually sample? (enabled ∧ DSE ∧ agent opts in)

The validator (renamed `validate_dse_env_params` → `validate_domain_randomization_active`) now just flags `enabled and not active`.

**On the name** (in case "domain randomization" is unfamiliar): `env_params` perturb the *environment* of each trial — a fresh random draw per step — so the agent is optimized across a *distribution* of conditions instead of one fixed setup. That is the established RL / sim-to-real term **domain randomization**. The old `dse_env_params` named the mechanism (DSE + env_params) and conflated two separate things: DSE is the search loop; domain randomization is the per-trial randomness layered on top. The predicate split keeps them distinct, and the validator now reads as the exact invariant it enforces.

rutayan-nv:
@coderabbitai review

rutayan-nv:
@coderabbitai review

src/cloudai/models/workload.py
```diff
@@ -191,3 +214,64 @@ def validate_agent_config(self) -> Self:
...
+            )
+        if not value:
+            raise ValueError(
+                f"env_params['{name}'] references an empty candidate list in cmd_args.{name}; "
+                "provide at least one candidate (the sampler would otherwise fail on an empty draw)"
+            )
+        if spec.weights is None:
+            return
+        if len(value) < 2:
+            raise ValueError(f"env_params['{name}'] declares weights but cmd_args.{name} needs >= 2 candidate values")
```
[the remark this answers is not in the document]
rutayan-nv: Good catch — fixed in `74621352`. A single-element list has the same degenerate semantics as the scalar case rejected in `8f91664d`, so I moved the `len(value) < 2` guard ahead of the `weights is None` short-circuit. It now rejects singleton candidate lists for every env-sampled field, not just weighted ones, with a dedicated message ("a single-element list is a fixed value, not domain randomization"). Added `test_env_params_singleton_candidate_list_rejected` to cover it.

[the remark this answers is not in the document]
rutayan-nv: done - 74621352

rutayan-nv:
@coderabbitai review

podkidyshev (approved):

PR #914, opened 2026-06-05
add DSE support for AI Dynamo + LMCache aiperf workload

saivishal1999:
Fix AIDynamoReportGenerationStrategy to correctly parse aiperf CSV output when running with non-standard metric names or multi-section CSV files.  
   
  Changes:                                                                                                                                           
  - Add _AnyMetricSet (a list subclass where __contains__ always returns True) so the strategy accepts any metric name reported by aiperf without
  needing an explicit allowlist                                                                                                                      
  - Pass on_bad_lines="skip" to pd.read_csv to gracefully handle malformed or multi-section lines in the aiperf output CSV
                                                                                                                                                     
  Why: aiperf reports metrics under varying names depending on workload configuration, and its CSV output can contain non-data header/separator lines
   that cause pandas to raise on strict parsing. These two changes make the report generation robust to both issues

### DIFF
--- a/src/cloudai/workloads/ai_dynamo/report_generation_strategy.py
+++ b/src/cloudai/workloads/ai_dynamo/report_generation_strategy.py
@@ -18,17 +18,28 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
+from typing import ClassVar
 
 from cloudai.core import METRIC_ERROR, MetricValue, ReportGenerationStrategy
 from cloudai.util.lazy_imports import lazy
 from cloudai.workloads.ai_dynamo.ai_dynamo import AIDynamoTestDefinition, parse_aiperf_accuracy
 
 
+class _AnyMetricSet(list):
+    """List subclass that reports containment for any item; used to mark a reporter as handling all metrics."""
+
+    def __contains__(self, item: object) -> bool:
+        return True
+
+
 class AIDynamoReportGenerationStrategy(ReportGenerationStrategy):
     """Strategy for generating reports from AI Dynamo run directories."""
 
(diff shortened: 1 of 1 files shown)


srivatsankrishnan (dismissed):

podkidyshev:
@saivishal1999 please adjust PR description. I see it mentions things I don't see in this PR. The main issue though: it's a public repo so please avoid mentions of private NV resources :D like clusters names and so on

anyway, the main point: this new reward function seems tied to a specific production DSE objective rather than to generic CloudAI behavior. Shouldn't we move it into the private repo?

This public repo currently only ships grid_search, and the production agents/configs that need this objective live in the private one, should this reward be defined/registered there instead? If we do keep something in public CloudAI, I’d suggest making it technology-neutral, e.g. ai_dynamo_ttft_throughput_log_scale

fyi @srivatsankrishnan

podkidyshev (approved):

srivatsankrishnan:
> @saivishal1999 please adjust PR description. I see it mentions things I don't see in this PR. The main issue though: it's a public repo so please avoid mentions of private NV resources :D like clusters names and so on
> 
> anyway, the main point: this new reward function seems tied to a specific production DSE objective rather than to generic CloudAI behavior. Shouldn't we move it into the private repo?
> 
> This public repo currently only ships grid_search, and the production agents/configs that need this objective live in the private one, should this reward be defined/registered there instead? If we do keep something in public CloudAI, I’d suggest making it technology-neutral, e.g. ai_dynamo_ttft_throughput_log_scale
> 
> fyi @srivatsankrishnan

This makes sense to me. @saivishal1999  I think we should keep these reward functions private.

PR #915, opened 2026-06-07
Map results dir into container at /cloudai_run_results for m-bridge t…

dshirron3:
## Summary
In order to allow direction of nccl debug file to cloudai results directory in case of megatron bridge template, need to add mapping of results dir inside the container (as done in nemorun and megatronrun templates)

## Test Plan
I verified this fix to work with adding the below to test toml file:
"NCCL_DEBUG_FILE" = "/cloudai_run_results/nccl_debug_file.%h.%p.txt"


dshirron3:
closing pull request as there is already an existing alternative.
/nemo_run is mapped to the run's experiment directory so a solution can be:
NCCL_DEBUG_FILE = "/nemo_run/nccl_debug_file.%h.%p.txt"

PR #919, opened 2026-06-09
Add machine-readable scenario summary report

### LINKED ISSUE #917: Add a stable machine-readable summary artifact for generated scenario reports
Hi CloudAI maintainers,

While building a small closed-loop experiment wrapper around CloudAI, I noticed that generated report artifacts seem to be intentionally flexible and workload-specific. That makes sense for rich workload reports, but it also makes automation harder because there does not appear to be one stable machine-readable artifact that external tools can rely on after a run.

From what I saw:

- `cloudai generate-report` writes reports under the scenario/result directory.
- The README documents a general HTML report as `<scenario>.html`.
- DSE scenarios can produce `<scenario>-dse-report.html` and a best config TOML.
- Workloads may produce their own artifacts, for example SGLang uses `sglang-bench.jsonl`, AI Dynamo uses CSV report files, and other workloads have their own filenames/fields.
- Custom reporters may also generate additional artifacts.

This is useful for humans, but external automation has to know workload-specific filenames and metric field names, or fall back to scraping logs/HTML.

## Why this matters

Closed-loop tools, CI jobs, dashboards, and experiment trackers need a predictable way to answer a few basic questions after a CloudAI run:

- Did the scenario/test run complete successfully?
- Where is the result directory?
- Which artifacts were generated?
- Which config files were used?
- What metrics were produced?
- Which workload/test run produced each metric?

Without a stable machine-readable summary, integrations have to duplicate CloudAI-specific parsing logic for each workload.

## Suggested improvement

Add a small JSON artifact, for example `cloudai-summary.json`, emitted during report generation. It could live next to the existing scenario reports and include fields like:

```json
{
  "scenario": "example-scenario",
  "status": "completed",
  "result_dir": "...",
  "reports": [
    {
      "name": "status",
      "path": "example-scenario.html",
      "format": "html"
    }
  ],
  "test_runs": [
    {
      "name": "sglang-bench",
      "status": "completed",
      "output_path": "...",
      "artifacts": [
        {
          "path": "sglang-bench.jsonl",
          "format": "jsonl"
        }
      ],
      "metrics": {
        "request_throughput": 42.5,
        "mean_ttft_ms": 18.2
      }
    }
  ]
}
```

This would not need to replace the existing HTML reports or workload-specific artifacts. It would just provide a stable automation entry point for tools that want to consume CloudAI results programmatically.

## Possible implementation direction

One lightweight path might be to add a scenario-level summary reporter that gathers known report paths, test run output paths, statuses, and any metrics already exposed by each workload's `ReportGenerationStrategy` / metric reporter. Even a minimal first version with scenario metadata, status, output paths, and artifact paths would be useful; normalized metrics could be expanded later.

## Context

I ran into this while trying to ingest CloudAI SGLang results from an external closed-loop config tuning tool. The SGLang JSONL output was parseable, but the integration had to know the exact filename and fields instead of reading a single CloudAI-level summary artifact.

shreyaskommuri:
## Summary
- Adds a default `summary` scenario reporter that writes `cloudai-summary.json` next to the existing scenario reports.
- Includes scenario name/status, system metadata, result directory, scenario-level report artifacts, per-test-run status, output paths, artifacts, and metrics already exposed through configured metric reporters.
- Registers the reporter by default and keeps `tarball` as the final scenario reporter.

Fixes NVIDIA/cloudai#917.

## Test Plan
Environment:
- macOS
- Python 3.10.20 via `uv run`

Commands and results:

```console
$ uv run ruff check src/cloudai/reporter.py src/cloudai/registration.py src/cloudai/_core/registry.py src/cloudai/core.py tests/test_reporter.py tests/test_init.py
All checks passed!
```

```console
$ uv run ruff format --check src/cloudai/reporter.py src/cloudai/registration.py src/cloudai/_core/registry.py src/cloudai/core.py tests/test_reporter.py tests/test_init.py
6 files already formatted
```

```console
$ uv run pytest tests/test_reporter.py tests/test_init.py tests/test_registry.py
60 passed in 0.43s
```

```console
$ uv run pytest
1573 passed, 4 skipped, 463 deselected in 5.14s
```

## Additional Notes
- This keeps existing HTML, tarball, comparison, and workload-specific reports unchanged.
- The summary intentionally uses metrics CloudAI already exposes through each test run's configured metric reporter, so this PR adds a stable automation artifact without introducing new workload-specific parsing logic.

### DIFF
--- a/src/cloudai/reporter.py
+++ b/src/cloudai/reporter.py
@@ -223,15 +223,10 @@ class SummaryReporter(Reporter):
         logging.info("Generated scenario summary at %s", report_path)
 
     def build_summary(self) -> dict[str, Any]:
-        test_runs = [self._test_run_summary(tr) for tr in self.trs]
+        test_runs = self._test_runs_summary()
         return {
-            "schema_version": "1.0",
             "scenario": self.test_scenario.name,
             "status": self._scenario_status(test_runs),
-            "system": {
-                "name": self.system.name,
-                "scheduler": self.system.scheduler,
-            },
             "result_dir": self._relative_path(self.results_root),
             "reports": self._scenario_artifacts(),
             "test_runs": test_runs,
@@ -244,20 +239,46 @@ class SummaryReporter(Reporter):
             return "completed"
         return "failed"
 
+    def _test_runs_summary(self) -> list[dict[str, Any]]:
+        loaded_by_name: dict[str, list[TestRun]] = {}
+        for tr in self.trs:
+            loaded_by_name.setdefault(tr.name, []).append(tr)
+
+        summary: list[dict[str, Any]] = []
+        for test_run in self.test_scenario.test_runs:
+            loaded_runs = loaded_by_name.get(test_run.name, [])
+            if test_run.is_dse_job:
+                summary.append(self._sweep_test_run_summary(test_run, loaded_runs))
+            else:
+                summary.extend(self._test_run_summary(tr) for tr in loaded_runs)
+
+        return summary
+
+    def _sweep_test_run_summary(self, tr: TestRun, sweeps: list[TestRun]) -> dict[str, Any]:
+        sweep_summaries = [self._test_run_summary(sweep) for sweep in sweeps]
+        summary = {
+            "name": tr.name,
+            "status": self._scenario_status(sweep_summaries),
+            "output_path": self._relative_path(self.results_root / tr.name),
+            "artifacts": self._artifacts_excluding(
+                self.results_root / tr.name, [sweep.output_path for sweep in sweeps]
+            ),
+            "metrics": {},
+            "sweeps": sweep_summaries,
+        }
+        return summary
+
     def _test_run_summary(self, tr: TestRun) -> dict[str, Any]:
         status = tr.test.was_run_successful(tr)
         summary = {
-            "name": tr.name,
-            "case": case_name(tr),
-            "description": tr.test.description,
-            "iteration": tr.current_iteration,
-            "step": tr.step,
+            "name": case_name(tr),
             "status": "completed" if status.is_successful else "failed",
-            "error_message": status.error_message,
             "output_path": self._relative_path(tr.output_path),
             "artifacts": self._artifacts(tr.output_path),
             "metrics": self._metrics(tr),
         }
+        if status.error_message:
+            summary["error_message"] = status.error_message
         return summary
 
     def _metrics(self, tr: TestRun) -> dict[str, float]:
@@ -285,6 +306,23 @@ class SummaryReporter(Reporter):
             return []
 
         return [self._artifact(path) for path in sorted(root.rglob("*")) if path.is_file()]
+
+    def _artifacts_excluding(self, root: Path, excluded_roots: list[Path]) -> list[dict[str, str]]:
+        if not root.is_dir():
+            return []
+
+        return [
+            self._artifact(path)
+            for path in sorted(root.rglob("*"))
+            if path.is_file() and not any(self._is_relative_to(path, excluded_root) for excluded_root in excluded_roots)
+        ]
+
+    def _is_relative_to(self, path: Path, root: Path) -> bool:
+        try:
+            path.relative_to(root)
+        except ValueError:
+            return False
+        return True
 
     def _artifact(self, path: Path) -> dict[str, str]:
         return {
--- a/src/cloudai/reporter.py
+++ b/src/cloudai/reporter.py
@@ -15,11 +15,12 @@
 # limitations under the License.
 
 import contextlib
+import json
 import logging
 import tarfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional
+from typing import Any, Optional
 
 import jinja2
 import toml
@@ -31,7 +32,7 @@ from cloudai.report_generator.dse_report import build_dse_summaries
 from cloudai.report_generator.util import load_system_metadata
 from cloudai.util.lazy_imports import lazy
 
-from .core import CommandGenStrategy, Reporter, TestRun, case_name
+from .core import METRIC_ERROR, CommandGenStrategy, Reporter, TestRun, case_name
 from .models.scenario import TestRunDetails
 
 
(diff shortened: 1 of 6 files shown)


tests/test_reporter.py
```diff
@@ -339,11 +340,57 @@ def test_metadata_for_single_sbatch(self, slurm_system: SlurmSystem, slurm_metad
...
+    (output_path / "cloudai_nccl_test_csv_report.csv").write_text("size,bw\n1,2\n")
+
+    scenario = TestScenario(name="test_scenario", test_runs=[benchmark_tr])
+    reporter = SummaryReporter(slurm_system, scenario, slurm_system.output_path, ReportConfig())
+    reporter.generate()
+
+    summary_path = slurm_system.output_path / SummaryReporter.SUMMARY_FILE_NAME
+    summary = json.loads(summary_path.read_text())
+
+    assert summary["schema_version"] == "1.0"
```
podkidyshev: I'd rather do a full object assert like

```python
assert summary == {
    "schema_version": "1.0",
    ...
}

File: src/cloudai/reporter.py [hunk 2]
```diff
@@ -207,6 +208,97 @@ def report_best_dse_config(self):
...
+        with report_path.open("w") as f:
+            json.dump(self.build_summary(), f, indent=2)
+            f.write("\n")
+
+        logging.info("Generated scenario summary at %s", report_path)
+
+    def build_summary(self) -> dict[str, Any]:
+        test_runs = [self._test_run_summary(tr) for tr in self.trs]
+        return {
+            "schema_version": "1.0",
```
podkidyshev: I think this property is a bit too much atm, let's keep the set of fields to minimum. Only those that bring value

src/cloudai/reporter.py
```diff
@@ -207,6 +208,97 @@ def report_best_dse_config(self):
...
+
+        return metrics
+
+    def _scenario_artifacts(self) -> list[dict[str, str]]:
+        if not self.results_root.is_dir():
+            return []
+
+        return [
+            self._artifact(path)
+            for path in sorted(self.results_root.iterdir())
```
podkidyshev, line 300: this won't work gracefully when we run sweeping. in the sweeping case, the folders structure will be

```
/scenario foler/
               /case 1/
                      /sweep 1/
                              /some-report.html
                      /sweep 2/
                              /some-report.html
```

podkidyshev: sweeping is actually quite important in our work. I believe this json summary report should bring insights if those were run

podkidyshev (changes requested):

shreyaskommuri:
Addressed the requested changes and updated the PR branch from latest main.

What changed:
- Replaced the summary reporter regression with a full-object JSON assert.
- Trimmed the first summary contract to the minimum useful automation fields: scenario, status, result_dir, reports, and test_runs.
- Grouped DSE/sweep outputs under their parent test run via a sweeps list, including case-level artifacts such as trajectory.csv and per-sweep nested report artifacts.

Validation on the updated PR head dfa28da3:
- uv run ruff check src/cloudai/reporter.py tests/test_reporter.py tests/test_init.py src/cloudai/registration.py src/cloudai/_core/registry.py src/cloudai/core.py -> passed
- uv run ruff format --check src/cloudai/reporter.py tests/test_reporter.py tests/test_init.py src/cloudai/registration.py src/cloudai/_core/registry.py src/cloudai/core.py -> passed
- uv run pytest tests/test_reporter.py tests/test_init.py tests/test_registry.py -q -> 60 passed
- uv run pytest -q -> 1599 passed, 4 skipped, 476 deselected

src/cloudai/reporter.py
```diff
@@ -207,6 +208,135 @@ def report_best_dse_config(self):
...
+    """Generate a machine-readable scenario summary for automation."""
+
+    SUMMARY_FILE_NAME = "cloudai-summary.json"
+
+    def generate(self) -> None:
+        self.load_test_runs()
+        report_path = self.results_root / self.SUMMARY_FILE_NAME
+        with report_path.open("w") as f:
+            json.dump(self.build_summary(), f, indent=2)
+            f.write("\n")
```
podkidyshev, line 221: a bit of nitpicking but let's make this code a bit more modern

```python
summary = self.build_summary()
report_path.write_text(json.dumps(summary, indent=2))
```

src/cloudai/reporter.py
```diff
@@ -207,6 +208,135 @@ def report_best_dse_config(self):
...
+
+        logging.info("Generated scenario summary at %s", report_path)
+
+    def build_summary(self) -> dict[str, Any]:
+        test_runs = self._test_runs_summary()
+        return {
+            "scenario": self.test_scenario.name,
+            "status": self._scenario_status(test_runs),
+            "result_dir": self._relative_path(self.results_root),
```
podkidyshev, line 230: why did you decide to make it relative? looking at your unit test I see

```python
assert summary == {
        "scenario": "test_scenario",
        "status": "completed",
        "result_dir": ".",
        ....
}
```

What's the value in `"result_dir": "."`? I would go for absolute value here

shreyaskommuri:
Rebased onto latest main to resolve merge conflicts (this had drifted significantly and was showing as `CONFLICTING`). Two small conflicts, both cosmetic collisions with unrelated upstream additions, not logic conflicts:
- `core.py`: `__all__` list — kept both `StructuredObservationProducer` (new upstream export) and `SummaryReporter` (this PR), alphabetized.
- `registration.py`: import block — kept both the new `TrainingReportGenerationStrategy` import and `SummaryReporter` added to the `reporter` import.

Re-ran `python -m pytest -q` (1756 passed, 18 skipped), `ruff check`, and `ruff format --check` — all clean. Also re-verified end-to-end: ran the bundled sleep scenario dry-run and confirmed `cloudai-summary.json` is still generated with the expected shape (`scenario`, `status`, `result_dir`, `reports`, `test_runs[].metrics`).

podkidyshev (changes requested):

src/cloudai/reporter.py
```diff
@@ -207,6 +208,135 @@ def report_best_dse_config(self):
...
+            "sweeps": sweep_summaries,
+        }
+        return summary
+
+    def _test_run_summary(self, tr: TestRun) -> dict[str, Any]:
+        status = tr.test.was_run_successful(tr)
+        summary = {
+            "name": case_name(tr),
+            "status": "completed" if status.is_successful else "failed",
+            "output_path": self._relative_path(tr.output_path),
```
podkidyshev, line 276: speaking about all the parths in the report:

* I would say the overall scenario path (result_dir) should be absolute
* **all the other** paths (per test run) should be relative to that result_dir path

src/cloudai/reporter.py
```diff
@@ -207,6 +208,135 @@ def report_best_dse_config(self):
                 toml.dump(trd.test_definition.model_dump(), f)
 
 
+class SummaryReporter(Reporter):
+    """Generate a machine-readable scenario summary for automation."""
+
+    SUMMARY_FILE_NAME = "cloudai-summary.json"
+
+    def generate(self) -> None:
+        self.load_test_runs()
```
podkidyshev, line 217: please research how this function works. apparently your implementation completely ignores the `iterations` feature (run one case multiple times). some people do fine it useful (when a workload is not consistent thus one runs it multiple times and calculates avg/max/min for metrics). 

there are a few existing issues however:

1. iterations aren't supported with sweeping 
2. `load_test_runs` iterates wrongly over dse vs iters (issue number 1. though makes this bug not reproducible in real runs)
3. `load_test_runs` doesn't group test runs per iteration
4. there's no mechanism to select one output metric per set of iterations per test-case/per sweep


all in all, this reporter you work on is something we truly need + we have an internal initiative to implement one. however it would take way more to make it well and the second we release it - we will have to maintain backwards compatibility for long

with a heavy heart, maybe it would be better to park this PR for now... yes, you could make the first step but I'll likely will stay defensive and unsure because of our internal plans for the near future and all the ground work needed to be done

podkidyshev (changes requested):

shreyaskommuri:
Before deciding whether to park this, I want to put a real root-cause writeup here, since I think the iterations-plus-sweeping gap is more specific and more consequential than it looked.

## Root cause

DSE-configured test runs never advance past `current_iteration = 0`, regardless of what `iterations` is set to in the scenario file. Two independent places confirm this:

1. `CloudAIGymEnv.reset()` sets `self.test_run.current_iteration = 0` once, and nothing in `cloudai_gym.py` ever increments it afterward. Only `step` advances, via `increment_step()`.
2. The normal runner's iteration-increment logic, in `base_runner.py` `on_job_completion`, is gated by `if completed_job.test_run.step <= 0`. Every real DSE step has `step >= 1` (since `increment_step()` runs before the job executes), so this code path is structurally skipped for every DSE-driven job run.

So this is not a reporter bug. It is an execution-layer gap in `handle_dse_job`/`CloudAIGymEnv`. It also means this silently affects `DSEReporter` today, not just this PR's `SummaryReporter`, since anyone running a DSE job with `iterations > 1` is only ever getting iteration 0's data, with no error or warning.

## Design question I don't want to guess on

Fixing this needs a decision on what `iterations` should mean for a DSE job, and I see two reasonable options:

- **A: repeat the whole sweep N times.** Wrap the per-test-run `agent.run()` call in `handle_dse_job` with an outer loop over `iterations`, incrementing `current_iteration` each pass. Directory layout becomes `<test>/<iteration>/<step>/`, which is what `load_test_runs` already expects, so this is the smaller change.
- **B: repeat each point N times.** For every sweep point, run it `iterations` times and aggregate before moving to the next point. This changes the trial/step semantics inside `CloudAIGymEnv` itself, bigger change, but might be closer to what "iterations" means elsewhere in CloudAI (variance reduction on a single config, not replaying an entire sweep).

I don't have enough context on how `iterations` is used elsewhere in DSE-adjacent work to pick between these confidently.

## Separate, downstream problem: metric aggregation

Once iterations produce multiple samples per sweep point, something needs to pick one representative value per point for reporting, mean, median, best, etc. Nothing does this today, this is the "no mechanism to select one output metric per set of iterations per test-case/per sweep" you flagged. This is a reporting-layer decision that depends on option A vs B above.

## Scope and backward compatibility

Given the fix touches `handle_dse_job`, which every DSE job goes through, this is real core-execution surface, not something scoped to the summary reporter. I want to flag that plainly rather than undersell it. Given your team's internal initiative and the backwards-compat concern you raised, I think the right move is: I hold off on writing any code, and you tell me if A, B, or neither is worth pursuing given what your internal plans already cover. If the answer is "we're already handling this internally," I'll close this PR and this stays as a documented gap for whoever picks it up next.

PR #930, opened 2026-06-16
feat(configurator): add GymnasiumAdapter for CloudAI envs

rutayan-nv:
## Issue
- RL agents (PPO/DQN) and external training loops need a `gymnasium.Env`-shaped view of a CloudAI `BaseGym`; there is no upstream adapter, and a flat `[0.0]` observation gives adapters the wrong `Box` shape.

## Fix
- Add `GymnasiumAdapter` (configurator): `spaces.Dict` of `Discrete` (list) + `Box` (`ContinuousSpace`) actions with fixed params injected per step; flat-`Box` or structured `spaces.Dict` (per-leaf `ObsLeafDescriptor`) observations; `dtype="int"` continuous actions quantized at `decode_action`. Pure pass-through over `test_run.step` (never mutated) so contextual-bandit `reset()`-per-trial keeps a monotonic trial index. `gymnasium` is lazy-imported behind a new `[rl]` extra; `define_observation_space()` now sizes by agent metrics. Exported via `cloudai.core`.

## Testing
- `tests/test_gymnasium_adapter_contract.py`: caller-contract tests for step-monotonicity (within/across episodes), observation pass-through, continuous quantization/clamping, and the structured-obs gate. ruff + pyright + vulture + import-linter clean; 108 related tests pass.

**Stack:** #901 ← #927 (ContinuousSpace) ← #928 (ObsLeafDescriptor) ← **this**. Final cloudai-side PR of the gymnasium-adapter upstreaming; consumes both primitives.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,9 @@ classifiers = [
     "import-linter~=2.10",
     "pytest-deadfixtures~=3.1",
     "taplo~=0.9.3",
+    "gymnasium~=1.2",
   ]
+  rl = ["gymnasium~=1.2"]
   docs = [
     "sphinx~=8.1",
     "nvidia-sphinx-theme~=0.0.8",
--- a/src/cloudai/configurator/cloudai_gym.py
+++ b/src/cloudai/configurator/cloudai_gym.py
@@ -26,7 +26,7 @@ from cloudai.util.lazy_imports import lazy
 
 from .base_agent import RewardOverrides
 from .base_gym import BaseGym
-from .env_params import EnvParams, write_env_params
+from .env_params import EnvParams, ObsLeafDescriptor, write_env_params
 
 
 @dataclasses.dataclass(frozen=True)
@@ -77,8 +77,9 @@ class CloudAIGymEnv(BaseGym):
         Index of the next trial ``step`` will run.
 
         ``step`` increments ``test_run.step`` before it samples/runs, so at rest the counter
-        holds the last-run trial and the next one is ``+ 1``. ``step`` advances into it. The
-        ``+ 1`` offset is defined only here so callers cannot re-encode it inconsistently.
+        holds the last-run trial and the next one is ``+ 1``. ``reset`` peeks this to report the
+        regime the first ``step`` will apply; ``step`` advances into it. The ``+ 1`` offset is
+        defined only here so the two paths cannot disagree.
         """
         return self.test_run.step + 1
 
@@ -119,9 +120,11 @@ class CloudAIGymEnv(BaseGym):
         if seed is not None:
             lazy.np.random.seed(seed)
         self.test_run.current_iteration = 0
-        observation = [0.0]
-        info = {}
-        return observation, info
+
+        info: dict[str, Any] = {}
+        if self.params is not None:
+            info["env_params"] = self.params.sample(self.upcoming_trial)
+        return self.define_observation_space(), info
 
     def step(self, action: Any) -> Tuple[list, float, bool, dict]:
         """
@@ -141,9 +144,11 @@ class CloudAIGymEnv(BaseGym):
         self.test_run.increment_step()
         # RNG lives in the env: sample here, then apply action + sample so the run and cache key see them.
         sampled_env_params = self.params.sample(trial) if self.params else {}
+        info: dict[str, Any] = {"env_params": sampled_env_params} if self.params is not None else {}
         self.test_run = self.test_run.apply_params_set(action, env_params=sampled_env_params)
 
         cached_result = self.get_cached_trajectory_result(action, sampled_env_params)
+
         if cached_result is not None:
             logging.info(
                 "Retrieved cached result from trajectory with reward %s (from step %s). Skipping execution.",
@@ -159,11 +164,12 @@ class CloudAIGymEnv(BaseGym):
                     env_params=sampled_env_params,
                 )
             )
-            return cached_result.observation, cached_result.reward, False, {}
+
+            return cached_result.observation, cached_result.reward, False, info
 
         if not self.test_run.test.constraint_check(self.test_run, self.runner.system):
             logging.info("Constraint check failed. Skipping step.")
-            return [-1.0], self.rewards.constraint_failure, True, {}
+            return [-1.0], self.rewards.constraint_failure, True, info
 
         new_tr = copy.deepcopy(self.test_run)
         new_tr.output_path = self.runner.get_job_output_path(new_tr)
@@ -185,20 +191,20 @@ class CloudAIGymEnv(BaseGym):
             self.test_run.step = new_tr.step
             self.test_run.output_path = new_tr.output_path
 
-        observation = self.get_observation(action)
-        reward = self.compute_reward(observation)
+        metrics = self.get_observation(action)
+        reward = self.compute_reward(metrics)
 
         self.write_trajectory(
             TrajectoryEntry(
                 step=self.test_run.step,
                 action=action,
                 reward=reward,
-                observation=observation,
+                observation=metrics,
                 env_params=sampled_env_params,
             )
         )
 
-        return observation, reward, False, {}
+        return metrics, reward, False, info
 
     def render(self, mode: str = "human"):
         """
@@ -252,6 +258,24 @@ class CloudAIGymEnv(BaseGym):
                 v = self.rewards.metric_failure
             observation.append(v)
         return observation
+
+    def structured_observation_descriptors(self) -> Optional[Dict[str, ObsLeafDescriptor]]:
+        """
+        Per-leaf descriptors for the env_param regime, or ``None`` when none are declared.
+
+        The flat observation (metrics) is unchanged; these describe the extra env_param leaves
+        the ``GymnasiumAdapter`` merges with the metrics leaf to build its ``spaces.Dict``.
+        """
+        return self.params.observation_descriptors() if self.params is not None else None
+
+    def encode_env_params(self, env_params: dict[str, Any]) -> Dict[str, Any]:
+        """
+        Encode a queried regime (the env_params behind an observation) into named leaves.
+
+        ``env_params`` is a ``{name: drawn value}`` regime (the ``info["env_params"]`` that
+        ``reset``/``step`` report). The adapter pairs the result with the flat metrics observation.
+        """
+        return self.params.encode(env_params) if self.params is not None else {}
 
     def write_trajectory(self, entry: TrajectoryEntry):
         """
--- a/src/cloudai/configurator/cloudai_gym.py
+++ b/src/cloudai/configurator/cloudai_gym.py
@@ -96,9 +96,10 @@ class CloudAIGymEnv(BaseGym):
         Define the observation space for the environment.
 
         Returns:
-            list: The observation space.
+            list: One float slot per agent metric (at least one), giving the correct shape
+            for adapters that derive ``gymnasium.spaces.Box`` from this output.
         """
-        return [0.0]
+        return [0.0] * max(len(self.test_run.test.agent_metrics), 1)
 
     def reset(
         self,
@@ -120,7 +121,6 @@ class CloudAIGymEnv(BaseGym):
         if seed is not None:
             lazy.np.random.seed(seed)
         self.test_run.current_iteration = 0
-
         info: dict[str, Any] = {}
         if self.params is not None:
             info["env_params"] = self.params.sample(self.upcoming_trial)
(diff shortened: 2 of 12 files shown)


rutayan-nv:
@coderabbitai review

rutayan-nv:
@coderabbitai review

rutayan-nv:
### Design note: why `GymnasiumAdapter` inherits `gymnasium.Env`, and why `unwrapped` became `cloudai_env`

Flagging this as a deliberate decision (it was an open design question) so it reads as intentional rather than incidental.

**Why inherit `gymnasium.Env` (vs. duck-typing, vs. `gymnasium.Wrapper`)**
- RL libraries (RLlib, SB3) and gymnasium's own `check_env` do `isinstance(env, gymnasium.Env)` and rely on base-`Env` behavior. A duck-typed object that merely returns the 5-tuple can fail those checks.
- `gymnasium.Wrapper` is the natural "I wrap another env" base, but it **requires the wrapped object to itself be a `gymnasium.Env`**. Our backend is a CloudAI `BaseGym` (a foreign type), so `Wrapper` doesn't fit. `Env` is the correct base: the adapter *is* the bottom-of-chain env that bridges a non-gymnasium backend.
- Optional dep is preserved: `Env` is imported under `try/except` with an `object` fallback (`TYPE_CHECKING` for pyright), and `spaces`/`numpy` stay lazy — installs without the `[rl]` extra still import fine; instantiation is what raises `ImportError`.

**Why rename `unwrapped` → `cloudai_env`**
- Once we inherit `Env`, `unwrapped` has a fixed contract: return the base `gymnasium.Env` (on a bare env, `self`). Our old `unwrapped` returned the inner `BaseGym` — the opposite meaning and an incompatible return type. That (a) pyright rejects as an invalid property override, and (b) breaks gymnasium tooling that calls `env.unwrapped` expecting a `gymnasium.Env` (e.g. `check_env`'s `env.unwrapped is not env`, RLlib reaching for `.np_random` / `.spec` / `.render_mode`).
- So we keep gymnasium's inherited `unwrapped` (returns `self`) and expose the CloudAI backend under an honestly-named accessor, `cloudai_env`.

**Knock-on effects in this PR** (so they don't look arbitrary): `metadata` drops `ClassVar` to match `Env`'s instance-attribute shape (`# noqa: RUF012`), and the contract test casts `action_space` to `spaces.Dict` because inheriting `Env` widens its static type to the base `Space`.

**Blast radius:** no callers inside `cloudai/` used `adapter.unwrapped`; `cloudai_env` is the only accessor. Downstream consumers should use `adapter.cloudai_env`.

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -76,9 +98,10 @@ def define_observation_space(self) -> list:
         Define the observation space for the environment.
 
         Returns:
-            list: The observation space.
+            list: One float slot per agent metric (at least one), giving the correct shape
+            for adapters that derive ``gymnasium.spaces.Box`` from this output.
         """
-        return [0.0]
+        return [0.0] * max(len(self.test_run.test.agent_metrics), 1)
```
podkidyshev, line 102: is it safe for existing agents?

src/cloudai/core.py
```diff
@@ -16,6 +16,7 @@
 
 """Core CloudAI base classes and interfaces."""
 
+from ._core.action_space import ContinuousSpace
```
podkidyshev: I was told ContinuousSpace is parked yet I see related code here, please don't forget about it 🙏

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,291 @@
...
+contextual-bandit rollout onto ``step=1`` because RLlib calls ``reset()`` per
+trial. This adapter never mutates ``test_run.step``; contract tests pin that
+property.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Optional, cast
+
+from cloudai._core.action_space import ContinuousSpace
```
podkidyshev: I assume this line is gonna be removed for now, but please import from `cloudai.core...`

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,291 @@
...
+    @staticmethod
+    def _structured_obs_descriptors(env: BaseGym) -> Optional[dict[str, Any]]:
+        """
+        Return the env's per-leaf obs descriptors, or ``None`` for the flat-Box path.
+
+        The env owns the opt-in decision via ``structured_observation_descriptors``
+        (returns ``None`` unless an observed name is a declared env_param). Envs
+        without that hook keep the legacy flat-Box path.
+        """
+        getter = getattr(env, "structured_observation_descriptors", None)
```
podkidyshev: I wasn't able to see `structured_observation_descriptors` in the BaseGym (nor in the CloudAIGymEnv) classes...

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,291 @@
...
+    def _descriptor_to_space(self, descriptor: Any) -> Any:
+        """Map a framework-agnostic ``ObsLeafDescriptor`` to a gymnasium subspace."""
+        if descriptor.kind == "discrete":
+            return self._spaces.Discrete(descriptor.n)
+        return self._spaces.Box(low=-self._np.inf, high=self._np.inf, shape=(descriptor.dim,), dtype=self._np.float32)
+
+    @property
+    def cloudai_env(self) -> BaseGym:
+        """Return the wrapped CloudAI :class:`BaseGym` (gymnasium's ``unwrapped`` returns ``self``)."""
+        return self._env
```
podkidyshev: what's the point of this property? I see no added value...

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Structured path: the flat raw obs (which feeds the env's reward and
+        ``trajectory.csv``) is encoded per-leaf by the env and returned as a
+        ``dict`` keyed to match ``observation_space`` (a ``spaces.Dict``).
+        Flat path: a single ``float32`` ``Box`` array (legacy behaviour).
+        """
+        descriptors = self._obs_descriptors
+        if descriptors is None:
+            return self._np.asarray(obs, dtype=self._np.float32)
+        env = cast(StructuredObservation, self._env)
```
podkidyshev: wait, in the `__init__` I see `self._env = env` where `env` comes from arguments and is a `BaseGym`. how does it end up being `StructuredObservation` here?

src/cloudai/configurator/cloudai_gym.py
rutayan-nv: Safe — and it's actually a latent correctness fix. The real observation has always been the length-`agent_metrics` metrics vector; `define_observation_space()` previously returned length-1 regardless of metric count, so the declared `observation_space` shape didn't match what the env emits.

Tracing every consumer:
- **RL (PPO/DQN)** wrap the adapter in `_ContextualDSEEnv` / `_FlatContextualDSEEnv`, which define their own 4-dim search-progress observation space and take only `adapter.action_space` — they never read the env observation.
- **BO / GA / MAB** accept `observation` only for `BaseAgent` ABC compatibility and ignore it.
- **nvopt_llmbo** reads the trajectory observation and indexes it by `metric_index`, i.e. it already expects the per-metric vector.

So no agent expects length-1; this aligns the declared shape with reality.

src/cloudai/core.py
rutayan-nv: Done — `ContinuousSpace` and all continuous-action code have been removed from the branch (the `cloudai._core.action_space` import, `_continuous_params`, the `Box` action handling, and the continuous encode/decode paths). Deferred as agreed; nothing continuous remains.

src/cloudai/configurator/gymnasium_adapter.py
rutayan-nv: The private `cloudai._core.action_space` import is gone with `ContinuousSpace`. For the remaining `BaseGym` / `StructuredObservationProducer`, I switched to relative sibling imports (`from .base_gym import ...`, `from .env_params import ...`) to match the `configurator` package convention — every other module there imports its siblings relatively; the absolute paths here were the outliers.

(`BaseGym` isn't re-exported by `cloudai.core`, and the adapter is imported by `cloudai.core`, so relative also avoids an import cycle.)

src/cloudai/configurator/gymnasium_adapter.py
rutayan-nv: Correct at the time — this branch predated the producer. It's now rebased onto the structured-obs producer, so `CloudAIGymEnv` implements `structured_observation_descriptors` and `encode_env_params`. The adapter now narrows via the `StructuredObservationProducer` protocol (`isinstance`) rather than `hasattr`.

src/cloudai/configurator/gymnasium_adapter.py
rutayan-nv: Removed. It was added speculatively alongside the `gymnasium.Env` inheritance switch (since `.unwrapped` returns `self`), but it has no in-repo consumer — every adapter holder also holds the raw env. If a consumer later needs a specific env capability through the adapter, we'll add a narrow forwarding method then rather than exposing the whole adaptee.

src/cloudai/configurator/gymnasium_adapter.py
rutayan-nv: Good catch — the `cast` hid where the safety came from (a `hasattr` guard in `__init__`). Replaced it: the protocol is `@runtime_checkable` and renamed `StructuredObservationProducer` (it *produces* structured observations, it isn't one), and the adapter narrows with `isinstance(self._env, StructuredObservationProducer)` at both sites — runtime-checked and type-narrowing, no `cast`. No inheritance is added to `CloudAIGymEnv` (it is not an is-a `StructuredObservationProducer`).

rutayan-nv:
@coderabbitai full review

pyproject.toml
```diff
@@ -57,7 +57,9 @@ classifiers = [
     "import-linter~=2.10",
     "pytest-deadfixtures~=3.1",
     "taplo~=0.9.3",
+    "gymnasium~=1.2",
   ]
+  rl = ["gymnasium~=1.2"]
```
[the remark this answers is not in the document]
rutayan-nv: Skipping this one. Standard `pyproject.toml` has no variable/anchor mechanism, so the only single-source alternative is a self-referencing extra (`dev = [..., "cloudai[rl]"]`). That couples `dev` to the package name and adds indirection for a single, well-known pin. Given this is a trivial/low-value nit, we prefer to keep the duplicated `gymnasium~=1.2` pin as-is.

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        metrics = self._np.asarray(obs, dtype=self._np.float32)
+        descriptors = self._obs_descriptors
+        env = self._env
+        if descriptors is None or not isinstance(env, StructuredObservationProducer):
+            return metrics
+        encoded = env.encode_env_params(info["env_params"])
+        self._assert_keys(encoded.keys(), set(descriptors), "encoded env_params")
+        context_leaves = {name: self._leaf_to_value(descriptors[name], encoded[name]) for name in descriptors}
+        return {"observation": metrics, "context": context_leaves}
```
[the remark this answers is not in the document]
rutayan-nv: Skipping this one after verifying it against the current code. The structured branch in `_as_obs` is only reached when the env declared observation descriptors (`self._obs_descriptors is not None`). In `CloudAIGymEnv` that is gated on the same `self.params is not None` that also populates `info["env_params"]` (both in `reset` and `step`). So "descriptors present" ⟺ "`env_params` present" is an invariant, and the branch is unreachable for a correctly-coupled producer.

For the legitimate no-`env_params` RL case (no `[env_params.*]` in the TOML), `structured_observation_descriptors()` returns `None`, so `_as_obs` returns the flat metrics Box before ever indexing `info["env_params"]` — no error. Adding a `ValueError` here would only guard a state that cannot occur (defensive dead code); if a future producer ever breaks the coupling, the plain `KeyError` is a sufficiently loud programming-error signal.

src/cloudai/configurator/gymnasium_adapter.py
```diff
@@ -0,0 +1,251 @@
...
+        # Two classes of params from cloudai's param_space:
+        #   list with len > 1   -> discrete tunable, mapped to gym.Discrete.
+        #   list with len == 1  -> fixed (collapsed); injected on every step so
+        #                          agents never see them.
+        self._discrete_params: dict[str, list] = {
+            k: v for k, v in raw_action_space.items() if isinstance(v, list) and len(v) > 1
+        }
+        self._fixed_params: dict[str, Any] = {
+            k: v[0] for k, v in raw_action_space.items() if isinstance(v, list) and len(v) == 1
+        }
```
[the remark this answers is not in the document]
rutayan-nv: done - 5663ca6c

podkidyshev (approved):

PR #921, opened 2026-06-10
Fix standalone sleep dry-run command generation

shreyaskommuri:
## Summary
- Fixes #920.
- Implements `store_test_run()` for `SleepStandaloneCommandGenStrategy` so the standalone sleep dry-run command generator is concrete and can be instantiated.
- Persists `test-run.toml` with `TestRunDetails`, matching the command generator contract used by other standalone workloads.
- Adds a regression test covering the generated sleep command and persisted test-run details.

## Test Plan
- Environment: local macOS checkout of CloudAI, branch `issue/sleep-standalone-store-test-run`.
- `uv run pytest tests/workloads/sleep/test_command_gen_strategy_standalone.py -q`
  - Result: `1 passed in 0.01s`
- `uv run pytest tests/workloads/sleep -q`
  - Result: `4 passed in 0.19s`
- `uv run pytest tests/test_init.py tests/workloads/sleep/test_command_gen_strategy_standalone.py tests/workloads/sleep/test_command_gen_strategy_slurm.py -q`
  - Result: `12 passed in 0.02s`
- `uv run ruff check src/cloudai/workloads/sleep/standalone_command_gen_strategy.py tests/workloads/sleep/test_command_gen_strategy_standalone.py`
  - Result: `All checks passed!`
- `uv run pre-commit run --all-files`
  - Result: all hooks passed, including `pyright`, `ruff check`, `ruff format`, `vulture`, `import-linter`, and `taplo`.

I also ran a standalone sleep dry-run smoke check with `--output-dir /tmp/...`. It got past the previous abstract-class crash, executed the sleep commands, and wrote `test-run.toml` files for the sleep test runs. The local harness interrupted before full CLI completion, so I am not claiming full dry-run completion from that smoke run.

## Additional Notes
- This issue was found while validating CloudAI through CloudAI Autotune: https://github.com/shreyaskommuri/CloudAI-Autotune
- AI was used for context and guidance while investigating and drafting this PR.

Issue #920, opened 2026-06-10
Standalone sleep dry-run crashes because command generator is abstract


shreyaskommuri:
# Standalone sleep dry-run crashes because command generator is abstract

## Describe the Bug

Running a standalone `sleep` scenario through `cloudai dry-run` crashes with a
Python traceback when CloudAI tries to instantiate
`SleepStandaloneCommandGenStrategy`.

The failure is:

```text
TypeError: Can't instantiate abstract class SleepStandaloneCommandGenStrategy with abstract method store_test_run
```

This came up while validating CloudAI through a small wrapper project,
[CloudAI Autotune](https://github.com/shreyaskommuri/CloudAI-Autotune).
Autotune can call the current CloudAI CLI shape, and the failure appears to
happen inside CloudAI after the standalone runner starts the sleep test.

Likely cause: `CommandGenStrategy` requires `store_test_run()` as an abstract
method, but `SleepStandaloneCommandGenStrategy` only implements
`gen_exec_command()`. Other standalone command generators, such as
`AiconfiguratorStandaloneCommandGenStrategy` and
`DynamoMockerStandaloneCommandGenStrategy`, do implement `store_test_run()`.

## Steps to Reproduce

Version/context where observed:

- Local CloudAI checkout based on `origin/main`.
- Also reproduced from a local branch based on the same checkout:
  `issue/machine-readable-summary`.
- CloudAI was invoked from CloudAI Autotune as the execution layer.

Run a standalone sleep dry-run, for example using CloudAI's documented sleep
scenario pattern:

```bash
cloudai dry-run \
  --system-config <standalone-system-config> \
  --tests-dir <tests-dir> \
  --test-scenario <sleep-scenario>
```

The dry-run reaches the standalone runner and starts `Tests.sleep1`, then
crashes:

```text
[INFO] System Name: standalone_system
[INFO] Scheduler: standalone
[INFO] Test Scenario Name: sleep-scenario
[INFO] Initializing Runner [DRY-RUN] mode
[INFO] Creating StandaloneRunner
[INFO] Starting test: Tests.sleep1
[INFO] Running test: Tests.sleep1
Traceback (most recent call last):
  ...
  File "cloudai/_core/base_runner.py", line 369, in get_cmd_gen_strategy
    return strategy_cls(system, test_run)
TypeError: Can't instantiate abstract class SleepStandaloneCommandGenStrategy with abstract method store_test_run
```

The same abstract-method mismatch is visible in the source:

[...]

PR #931, opened 2026-06-16
Guard standalone termination against zero job id

### LINKED ISSUE #929: Standalone sleep dry-run can kill parent wrapper when job id is 0
## Describe the Bug

While testing CloudAI as the benchmark execution layer for a small wrapper project, CloudAI Autotune, I found a standalone sleep `dry-run` path that can terminate the parent wrapper process group.

The dry-run reached the common sleep scenario and then logged:

```text
[INFO] Scheduling termination of job 0 after 0 seconds.
```

The CloudAI debug log for the same run shows the termination command:

```text
Executing termination command for job 0: kill -9 0
```

Because `kill -9 0` targets the current process group, the parent wrapper process was killed too. From the wrapper's perspective, the command exited with code `137` before it could print a normal pass/fail summary or parse any resulting artifacts.

The likely area is standalone job termination. In my local checkout, `src/cloudai/systems/standalone/standalone_system.py` builds:

```python
cmd = f"kill -9 {job.id}"
```

If `job.id` is `0`, that becomes `kill -9 0`.

## Steps to Reproduce

Observed using a local CloudAI checkout at commit `dfa28da3` with the common standalone sleep scenario.

I triggered it through a wrapper command that shells out to CloudAI dry-run:

```bash
python -m autotune.cli smoke-cloudai \
  /Users/shreyas/cloudai/conf/common/test_scenario/sleep.toml \
  --cloudai-bin /Users/shreyas/cloudai/.venv/bin/cloudai \
  --system-config /Users/shreyas/cloudai/conf/common/system/standalone_system.toml \
  --tests-dir /Users/shreyas/cloudai/conf/common/test \
  --runs-dir runs \
  --timeout-sec 60
```

Relevant stdout captured by the wrapper:

```text
[INFO] System Name: standalone_system
[INFO] Scheduler: standalone
[INFO] Test Scenario Name: sleep-scenario
[INFO] Creating StandaloneRunner
[INFO] Starting test: Tests.sleep1
[INFO] Executing command for test Tests.sleep1: sleep 10
[INFO] Starting test: Tests.sleep20
[INFO] Executing command for test Tests.sleep20: sleep 20
[INFO] Starting test: Tests.sleep5
[INFO] Executing command for test Tests.sleep5: sleep 5
[INFO] Job completed: Tests.sleep1 (iteration 1 of 1)
[INFO] Starting test: Tests.sleep5_2
[INFO] Executing command for test Tests.sleep5_2: sleep 5
[INFO] Scheduling termination of job 0 after 0 seconds.
```

The parent wrapper then exited with code `137`.

## Expected Behavior

A standalone CloudAI dry-run should return control to the parent process cleanly, even when a dependency-triggered termination is scheduled.

For wrapper tools, CI jobs, and smoke tests, I would expect one of these outcomes:

- CloudAI completes dry-run successfully and exits `0`; or
- CloudAI reports a scoped job termination/failure and exits non-zero without killing the caller.

The termination command should target only the child process/job CloudAI owns, not process group `0`.

## Suggested Fix Direction

- Guard standalone `kill` against invalid or process-group-sensitive ids such as `0`.
- Track the actual subprocess PID for standalone jobs and terminate that PID/process tree explicitly.
- Add a regression test for dry-running `conf/common/test_scenario/sleep.toml` under a parent wrapper/subprocess and assert the parent process survives.

## Screenshots

Not applicable. The relevant evidence is the captured stdout/debug-log text above.

shreyaskommuri:
## Summary

- Guard standalone job termination so `job.id == 0` does not shell out to `kill -9 0`.
- Log a warning and skip termination when the standalone job ID is the dry-run sentinel rather than a launched process.
- Add regression coverage for integer and string zero job IDs.

## Context

Fixes NVIDIA/cloudai#929.

While using CloudAI as the execution layer for CloudAI Autotune, a wrapper smoke test hit the common standalone sleep dry-run path. CloudAI scheduled dependency termination for job `0`, which produced `kill -9 0`. On POSIX shells that targets the current process group, so the parent wrapper was killed and exited with code `137` before it could report a normal result.

In standalone dry-run mode, `StandaloneRunner._submit_test()` uses `job_id = 0` because no subprocess is launched. That sentinel should not be passed to the shell as a real kill target.

## Test Plan

- `python -m pytest tests/systems/standalone/test_system.py -q`
  - `5 passed`
- `python -m pytest -q`
  - `1574 passed, 4 skipped, 463 deselected`
- `uv run pre-commit run --all-files`
  - all hooks passed

I also attempted the live standalone sleep dry-run on a fresh `origin/main` branch, but that path is currently blocked earlier by the already-filed abstract-method issue NVIDIA/cloudai#920 / PR NVIDIA/cloudai#921. This PR keeps the scope to the zero-job-id termination fix from NVIDIA/cloudai#929.

### DIFF
--- a/src/cloudai/systems/standalone/standalone_system.py
+++ b/src/cloudai/systems/standalone/standalone_system.py
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
-# Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -78,6 +78,22 @@ class StandaloneSystem(System):
         Args:
             job (BaseJob): The job to be terminated.
         """
-        cmd = f"kill -9 {job.id}"
-        logging.debug(f"Executing termination command for job {job.id}: {cmd}")
+        try:
+            pid = int(str(job.id).strip())
+        except ValueError:
+            logging.warning(
+                "Skipping termination for standalone job %s because it is not a valid process ID.",
+                job.id,
+            )
+            return
+
+        if pid <= 0:
+            logging.warning(
+                "Skipping termination for standalone job %s because it does not reference a launched process.",
+                job.id,
+            )
+            return
+
+        cmd = f"kill -9 {pid}"
+        logging.debug(f"Executing termination command for job {pid}: {cmd}")
         self.cmd_shell.execute(cmd)
--- a/tests/systems/standalone/test_system.py
+++ b/tests/systems/standalone/test_system.py
@@ -108,3 +108,39 @@ def test_kill_job(mock_execute, standalone_system, standalone_job):
     kill_command = f"kill -9 {standalone_job.id}"
 
     mock_execute.assert_called_once_with(kill_command)
+
+
+@pytest.mark.parametrize("job_id", [0, "0", "00", "+0", "-0", -1, "-1", "not-a-pid"])
+@patch("cloudai.util.CommandShell.execute")
+def test_kill_job_skips_invalid_pid(mock_execute, standalone_system, mock_test, job_id):
+    """
+    Test that standalone dry-run sentinel IDs and invalid PIDs are not killed.
+
+    Args:
+        mock_execute (MagicMock): Mocked CommandShell execute method.
+        standalone_system (StandaloneSystem): Instance of the system under test.
+        mock_test (Test): The mock test instance associated with the job.
+        job_id (int | str): Job ID that must not be killed.
+    """
+    job = StandaloneJob(mock_test, id=job_id)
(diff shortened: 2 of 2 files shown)


shreyaskommuri:
@coderabbitai review

shreyaskommuri:
Addressed CodeRabbit feedback by parsing standalone job IDs as integers before constructing the kill command. The guard now skips non-numeric IDs and all non-positive PID variants such as `00`, `+0`, `-0`, and `-1`, and uses the normalized positive PID for valid kills. Added regression coverage for those cases plus numeric normalization. Re-ran `python -m pytest tests/systems/standalone/test_system.py -q`, full `python -m pytest -q`, and `uv run pre-commit run --all-files` successfully.

shreyaskommuri:
Rebased onto latest main (was 186 commits behind) to keep this mergeable. No conflicts. Re-ran `python -m pytest -q` (1764 passed, 18 skipped), `ruff check`, and `ruff format --check` on the changed files — all clean. Re-verified the fix against the bundled sleep scenario: dry-run now exits 0 with all jobs `PASSED` where it previously died with `SIGKILL` from `kill -9 0`.

podkidyshev (approved):

PR #933, opened 2026-06-16
[CLI/Core] Polymorphic agent dispatch + TestRun owns trial counter

rutayan-nv:
## Issue
- `TestRun.step` had multiple writers; RLlib's frequent `reset()` collapsed every trial onto `step=1`, overwriting `trajectory.csv`/`env.csv` rows.
- The DSE loop lacked a uniform agent entry point and didn't distinguish recoverable failures from bugs.

## Fix
- `BaseAgent.run()` is the single polymorphic entry point; `handle_dse_job` is just `err |= agent.run()` — agents implement their own training loop in `run()`.
- `TestRun.increment_step()` is the sole mutator, called only by `CloudAIGymEnv.step()` (advances before `output_path`/trajectory are computed).
- Hybrid failure contract: a non-zero rc accumulates and the sweep continues; an unexpected exception hard-fails (reports still generated, error written to `<scenario_root>/dse_failure.txt`, then re-raised).

## Testing
`pytest tests` green; ruff + pyright + vulture clean.

> Replaces #893 (branch renamed `custom-training-loop-dispatch` → `agent-run-polymorphism`; history rebuilt into 2 clean commits, no add-then-remove churn). Stack base for #900 → #901 → #927 → #928 → #930 → #932.

### DIFF
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -110,6 +110,11 @@ class TestRun:
         """
         return self.current_iteration + 1 < self.iterations
 
+    def increment_step(self) -> int:
+        """Advance the trial counter and return the new value."""
+        self.step += 1
+        return self.step
+
     @property
     def metric_reporter(self) -> Optional[Type[ReportGenerationStrategy]]:
         if not self.reports:
--- a/src/cloudai/cli/handlers.py
+++ b/src/cloudai/cli/handlers.py
@@ -18,6 +18,7 @@ import argparse
 import copy
 import logging
 import signal
+import traceback
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Callable, List, Optional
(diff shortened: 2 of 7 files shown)


rutayan-nv:
@coderabbitai review

rutayan-nv:
@podkidyshev — this PR supersedes the closed #893 (head branch was renamed, which auto-closed it). Carrying over your two review threads from there since GitHub stranded them on the closed PR:

**1. [`handlers.py:121` — move out of handlers / mirror test structure](https://github.com/NVIDIA/cloudai/pull/893#discussion_r3267205188)**
Resolved by the `agent.run()` polymorphism refactor: the custom-loop dispatch, the `CustomTrainingLoopAgent` Protocol, and the TypeGuard are gone from `handlers.py`. `handle_dse_job` now collapses to `err |= agent.run()` and agents that drive their own loop just override `run()`. Since no agent-loop logic remains in `handlers.py`, the only tests left are `handle_dse_job` delegation tests, which stay in `tests/test_handlers.py` (per your "not related to tests against `handle_dse_job`" note). If you'd still like a `tests/configurator/test_base_agent.py` split, say the word.

**2. [`handlers.py:215` — early `return err` on err > 0?](https://github.com/NVIDIA/cloudai/pull/893#discussion_r3267248062)** — still your call.
Current behavior is a hybrid: a *graceful* non-zero rc from `agent.run()` is accumulated (`err |= ...`) so the remaining independent `test_run`s still execute and produce results/reports, and the handler returns non-zero if any failed. An *unexpected exception* (a bug) is a hard-fail — it's caught only long enough to generate reports, then re-raised. Happy to switch to fail-fast for the whole scenario if you prefer that semantics.

rutayan-nv:
@podkidyshev — re: item 2 above (early `return err` on err > 0):

This is resolved — handled as a hybrid policy in `handle_dse_job`:
- Recoverable failures (non-zero rc from `agent.run()`) accumulate via `err |= agent.run()` and the sweep continues; the accumulated `err` is returned. No early-return.
- An unexpected exception is a hard-fail: it's captured, reports are generated (`generate_reports(error=...)`, which writes `dse_failure.txt` with the traceback), then the exception is re-raised.

Lives on #933 (agent-run-polymorphism), inherited up the stack; routing + the mixed-job guard are on #934. Covered by `test_handle_dse_job_accumulates_nonzero_rc_and_continues`, `..._propagates_agent_run_exception`, `..._hard_fail_aborts_remaining_runs`, and `..._documents_failure_in_reports_before_raising`. No further decision needed.

podkidyshev (approved):

PR #938, opened 2026-06-21
fix(aiconfig): unwrap single-value list dims in standalone command-gen

rutayan-nv:
## Summary

The Aiconfigurator standalone command-gen rendered single-value parallelism dims — declared as one-element lists in TOML (e.g. `p_pp = [1]`) — as the literal string `[1]`. `simple_predictor.py` parses those args as `int` and rejected `"[1]"`, so every predictor invocation failed and the workload emitted empty reports (all `-1` observations).

This adds a `_scalar()` helper that unwraps one-element lists to their scalar and raises on multi-element lists (an unresolved sweep leaking into command generation). Applied to the disagg, agg, and isl/osl args.

Independent of the RL/Gym enablement stack — a standalone correctness fix for the Aiconfigurator workload. It was surfaced while running aiconfigurator disagg RL end-to-end: the run only produced real metrics after this fix.

## Test plan

- [x] `test_gen_exec_command_unwraps_single_value_list_dims` — `p_pp=[1]` renders as `--p-pp 1`, never `[1]`
- [x] `test_gen_exec_command_rejects_unresolved_sweep` — a multi-element list raises
- [x] existing standalone command-gen tests stay green
- [x] e2e: aiconfigurator disagg RL (ppo + dqn) now produces real metrics in `trajectory.csv` (was all `-1` before)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,9 @@ requires-python = ">=3.10"
     "import-linter~=2.10",
     "pytest-deadfixtures~=3.1",
     "taplo~=0.9.3",
+    "gymnasium~=1.2",
   ]
+  rl = ["gymnasium~=1.2"]
   docs = [
     "sphinx~=8.1",
     "nvidia-sphinx-theme~=0.0.8",
--- a/src/cloudai/_core/test_scenario.py
+++ b/src/cloudai/_core/test_scenario.py
@@ -203,10 +203,7 @@ class TestRun:
             else:
                 setattr(obj, attrs[-1], value)
 
-        # The agent's chosen action and this trial's sampled env_params are both just concrete
-        # values written onto cmd_args, applied through the one path here. env_params keys name
-        # cmd_args fields; sampling (the RNG) happens in the env before this call, so this method
-        # stays deterministic.
+        # RNG runs in the env before this call; applying only concrete values keeps this deterministic.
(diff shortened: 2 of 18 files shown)


rutayan-nv:
Rebased onto #930 (`rpatro/gymnasium-adapter`) so this now sits on top of the env_params/gym stack: `#901 -> #928 -> #930 -> #938`. Base stays `main` (the parent branch lives on the fork, so GitHub can't use it as a PR base), so the diff here is cumulative and will shrink to just the aiconfig command-gen fix once #930 lands. Review the top commit `b0731e7e` for this PR's actual change. This pairing (#930 + #938) is what cloudaix #612 pins to for the aiconfigurator RL release. @podkidyshev

rutayan-nv:
Closing in favor of the central fix.

This PR is a **local band-aid** in the aiconfigurator standalone command-gen (unwraps single-value list dims). The **real, durable fix** — a parse-time canonicalization invariant (`[x]` → `x`, so fixed = scalar / sweep = list of ≥2) plus a single central command-gen boundary guard that fails loud on any unresolved sweep — is tracked in #946 and supersedes this per-workload patch.

Full implementation spec is in the **"Prompt for AI agent"** dropdown at the top of:
https://github.com/NVIDIA/cloudai/issues/946

PR #939, opened 2026-06-22
test(aiconfig): add aiconfigurator SDK contract canary

rutayan-nv:
## What
A lightweight CI canary that verifies `aiconfig`'s `runtime/predictor.py` stays compatible with the **pinned** `aiconfigurator` SDK.

## Why
`predictor.py` is our adapter onto `aiconfigurator.sdk.*`; the version is pinned by the aiconfig workload's `PythonEnvironment` (`aiconfigurator~=0.5.0`). If a pin bump renames or drops a SDK symbol the adapter uses, the adapter breaks at run time and DSE silently stops producing metrics — and nothing catches it today (the existing aiconfig tests mock or never touch the SDK).

## How
- `ci_only` test. Fetches **only** the pinned source via `uv pip install --no-deps` (seconds, no torch), then **AST-audits** it — no imports, so the torch-importing SDK modules still parse.
- Asserts the exact surface `predictor.py` consumes still exists: `ModelConfig` / `RuntimeConfig` kwargs, the five `*QuantMode` enums + members, `get_database` / `get_model` / `get_backend`, and `InferenceSession.run_static`.
- The pin is read from the workload (single source of truth); expected symbols are kept in lockstep with `predictor.py`.

## Scope
Catches input-side drift (removed/renamed symbols). Return-shape changes (e.g. a renamed summary-df column) need a real prediction run and are out of scope.

## Test
`pytest -m ci_only` → 7 passed (~0.2s). Deselected from the hermetic suite; `tests/*` coverage gate unaffected.

---
Review: @podkidyshev

### DIFF
--- a/tests/workloads/aiconfig/test_aiconfigurator_sdk_contract.py
+++ b/tests/workloads/aiconfig/test_aiconfigurator_sdk_contract.py
@@ -1,0 +1,217 @@
+# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


podkidyshev:
@rutayan-nv I'm not quite sure about this PR... It crosses the line of a unit test by accessing outside world (fetching the lib). Additionally, one of the PythonEnvironment features is to use a custom python version if needed (used for one of the private workloads) which this unit test doesn't do, it modifies current venv. Essentially it's an integration test now and we don't have any dedicated CI environment for it yet.

What I would do is make the python version + aiconfigurator version configurable in the test toml so that users can control it. What do you think?

PR #947, opened 2026-06-29
Unified Training Output for training workloads (M-Bridge, MegatronRun, NemoRun)

blugassi:
## Summary
Produces a unified `training_report.json` across the three training workloads
(`nemo_run`, `megatron_run`, `megatron_bridge`) so their config and per-step metrics
become directly comparable: default-on logging makes every run emit what the report
needs, and a parser normalizes each framework's artifacts into a single schema.

This PR consists of **4 commits**; I recommend reviewing them one by one:

1. **`feat(nemo_run): dump resolved training config`** — `ConfigDumpCallback` writes
   NeMo's resolved config to `nemo_config.json`, attached at the `cloudai_nemorun`
   entrypoint so any `--factory` emits it.
2. **`feat(nemo_run): enable TFLOPS_per_GPU metrics`** — `FLOPsMeasurementCallback`
   puts `TFLOPS_per_GPU` in the TB stream for any `--factory`; fails soft (skips, never
   crashes) on unsupported models.
3. **`feat(training_output): default-on logging flags`** — turns on the config dump
   (`ConfigContainer.yaml`) and TensorBoard scalars for Megatron-Bridge/Run; updates the
   reference `.sbatch` fixtures.
4. **`feat(training_output): training report parser`** — adds the parser under
   `src/cloudai/reports/training/`, mapping TB scalars → `TrainingStep` and config artifacts
   → `TrainingConfig` via per-workload `str -> str` mappings, deriving CloudAI fields
   (`world_size`, `data_parallel_size`, `model_name`, `num_nodes`). Wired in via
   `registration.py`.

## Test Plan
**Environment:** NVIDIA GB200, CloudAI in a local `.venv` (Python 3.10).

- Ran `cloudai generate-report` on completed `nemo_run`, `megatron_run`, and
  `megatron_bridge` runs; all three produce a valid `training_report.json` with 100 steps,
  a populated `config`, declining loss, and memory reported in bytes (Megatron-Bridge GB
  values scaled), with `data_parallel_size == world // (tp·pp·cp)` for each.
- `pytest tests/reports/test_training_parser.py` → 10 passed.
- pre-commit (ruff, pyright, vulture, import-linter, taplo) passes.


src/cloudai/report_generator/training/parser.py
```diff
@@ -0,0 +1,215 @@
...
+        gpus_per_node = (
+            getattr(system, "gpus_per_node", None)
+            or getattr(system, "ntasks_per_node", None)
+            or self.DEFAULT_GPUS_PER_NODE
+        )
+        config.num_nodes = tr.nnodes
+        config.world_size = config.num_nodes * gpus_per_node
+        config.model_name = self.get_model_name(tr)
+        config.data_parallel_size = config.world_size // self._parallel_product(config)
+        return config
```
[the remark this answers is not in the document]
blugassi: Skipping this one — it's not valid for Megatron-Core/NeMo semantics. world_size = TP·PP·CP·DP; expert parallelism subdivides the data-parallel dimension (EP·expert_DP = DP·CP), so it's not an independent factor in world_size and EP ≤ DP always holds. Dividing by EP undercounts DP — e.g. the nemo_run MoE run (world=32, tp=4, ep=8) would report data_parallel_size=1 with expert_parallel_size=8, i.e. EP > DP, which is impossible. Current world_size/(tp·pp·cp) is correct (EP=8 cleanly divides DP=8).

src/cloudai/report_generator/training/report_generation_strategy.py
```diff
@@ -0,0 +1,57 @@
...
+        "MegatronBridge": MegatronBridgeParser,
+    }
+
+    def can_handle_directory(self) -> bool:
+        parser_cls = self.PARSERS.get(self.test_run.test.test_template_name)
+        return parser_cls is not None and parser_cls().can_parse(self.test_run)
+
+    def generate_report(self) -> None:
+        parser_cls = self.PARSERS[self.test_run.test.test_template_name]
+        training_results = parser_cls().parse(self.test_run, self.system)
```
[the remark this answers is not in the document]
blugassi: Thanks — this is already handled one layer down. can_handle_directory() delegates to TrainingParser.can_parse(), which now gates on the config artifact in addition to the TB dir:


def can_parse(self, tr: TestRun) -> bool:
    tb_dir = self.get_tb_dir(tr)
    if not (tb_dir.is_dir() and any(tb_dir.iterdir())):
        return False
    config_path = self.get_config_path(tr)
    return config_path is not None and config_path.is_file()
Each parser points get_config_path() at its required artifact — nemo_config.json (NeMo), ConfigContainer.yaml (M-Bridge), and a TB event file for Megatron (whose config lives in the TB text stream). So a run whose fail-soft config dump produced no file returns can_handle_directory() == False and parse()/get_config() is never reached — no crash. I kept the check in can_parse() rather than wrapping get_config() in can_handle_directory() so the gating lives with the parser that knows its own artifacts. Resolving as addressed.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -170,6 +173,62 @@ def timing_callback() -> run.Config[TimingCallback]:
...
+        try:
+            # Single failure domain: if any of the three roots raises (module without
+            # .config, non-Megatron strategy, datamodule None), the whole file is skipped.
+            cfg = {
+                "model": dataclasses.asdict(pl_module.config),
+                "parallelism": dataclasses.asdict(trainer.strategy.parallelism),
+                "data": vars(trainer.datamodule),
+            }
+            Path("/cloudai_run_results/nemo_config.json").write_text(json.dumps(cfg, indent=2, default=str))
+        except Exception as e:  # reporting must never break training
```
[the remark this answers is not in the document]
blugassi: Intentionally keeping this atomic — I'd prefer not to write partial configs.

The file feeds the unified training-report parser, which maps all three sections (model, parallelism, data) into a single TrainingConfig. A partial file would still pass can_parse() (the file exists) but then _build_config would resolve the missing section's fields to None — e.g. a missing parallelism section makes tensor_parallel_size None, which now raises in _compute_data_parallel_size. So a partial dump produces a half-valid/again-failing report instead of a clean outcome.

The all-or-nothing contract is deliberate: either we emit a complete, trustworthy nemo_config.json, or we emit nothing and can_parse() cleanly skips the run (no crash, logged). The three roots also share one failure domain in practice — a module without .config, a non-Megatron strategy, or a None datamodule are fundamental setup problems, not cases where two sections are meaningfully usable on their own. Splitting them would trade a clean skip for a partially-populated, misleading report. Keeping as-is.

src/cloudai/workloads/nemo_run/cloudai_nemorun.py
```diff
@@ -170,6 +173,62 @@ def timing_callback() -> run.Config[TimingCallback]:
...
+    def setup(self, trainer, pl_module, stage):
+        if any(isinstance(cb, FLOPsMeasurementCallback) for cb in trainer.callbacks):
+            return
+        try:
+            flops_cb = FLOPsMeasurementCallback(
+                model_config=pl_module.config,
+                data_config=trainer.datamodule,
+                model_name=os.getenv("CLOUDAI_NEMO_RECIPE", ""),
+            )
+            flops_cb.eval_model_flops()  # raises if the recipe maps to no FLOPs formula
```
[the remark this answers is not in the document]
blugassi: Skipping this one — no translation is needed. NeMo's eval_model_flops() already resolves the family by substring match: it does model_matches = [m for m in _model_flops_map if m in self.model] over {gpt3, llama2, llama3, nemotron, mixtral, bert}. CloudAI's factory names are family-prefixed, so they resolve as-is — gpt3_175b → gpt3, llama3_70b → llama3, llama31_405b → llama3, nemotron4_15b → nemotron. Confirmed empirically: our gpt3_175b run emits TFLOPS_per_GPU (≈4.7 → 591 across steps) in the generated report.

For families NeMo has no formula for (e.g. deepseek/kimi/qwen), there's nothing to map to, so a name translation can't produce TFLOPS anyway — and that path is handled by the existing try/except, which skips FLOPs without breaking training.

The llama_3b example doesn't match any current factory name produced by NeMoRunSlurmCommandGenStrategy; the real names carry the family prefix. Also, since tflops_per_gpu is now an optional field in the report schema, a missing FLOPs value no longer drops the step — the run still produces a full report with tflops_per_gpu: null. So there's no correctness loss to address here.

src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -246,6 +246,16 @@ def _list_or_comma_str(self, val: str | list[str] | None) -> Optional[str]:
...
 
+    def _add_extra_cmd_args(self, extra_cmd_args: dict[str, str]) -> list[str]:
+        """Hydra overrides: defaults merged with the user's extra_cmd_args, which take precedence."""
+        overrides = {
+            "logger.log_timers_to_tensorboard": "true",
+            "logger.log_throughput_to_tensorboard": "true",
+            "logger.log_memory_to_tensorboard": "true",
+        }
+        overrides.update(extra_cmd_args)
+        return [f"{key}={value}" if value else key for key, value in overrides.items()]
```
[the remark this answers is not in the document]
blugassi: Fixed in the latest push. Each override token is now shlex.quote-d so values with spaces or shell metacharacters survive the later " ".join(parts) + shell as a single argv token:

overrides.update(extra_cmd_args)
return [shlex.quote(f"{key}={value}" if value else key) for key, value in overrides.items()]
On the falsey-value point: I kept the f"{k}={v}" if v else k form deliberately — it matches the existing WorkloadModel.extra_args_str convention (empty value = intentional bare flag), and extra_cmd_args is typed dict[str, str], so 0/False can't occur (only ""). shlex.quote is a no-op for already-safe tokens, so the default logger.*=true overrides are unchanged and no .sbatch ref-data shifted. Resolving.

src/cloudai/reports/training/mappings.py
```diff
@@ -0,0 +1,127 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Per-workload field mappings for the training report parsers.
+
+Source of truth: doc/unified-training-output-design.md sections 5.2 (TrainingStep ->
+TensorBoard tags) and 6.1.1 (TrainingConfig -> framework artifact field names).
```
podkidyshev: is this .md file supposed to be a part of this PR? 

* if yes - please add it,
* if no - please remove this line (I assume it references some internal doc, right?)

src/cloudai/report_generator/training/__init__.py
podkidyshev, line 1: let's move the whole `src/cloudai/reports/training` into `/src/cloudai/report_generator/training`

`report_generator` is not the best module name yet it already exists and contains all reporting code. otherwise there are both `src/cloudai/reports` and `src/cloudai/report_generator`

(please don't forget to move tests either. atm there's no tests/report_generator level - let's bring one)

src/cloudai/workloads/nemo_run/nemo_run.py
```diff
@@ -85,6 +85,7 @@ class Trainer(BaseModel):
     strategy: TrainerStrategy = Field(default_factory=TrainerStrategy)
     plugins: Optional[Plugin] = None
     callbacks: Optional[Union[str, list[str]]] = None
+    log_every_n_steps: int = 1
```
podkidyshev, line 88: this brings no compute overhead, right?

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
```diff
@@ -39,6 +39,7 @@ def _set_additional_env_vars(self, tdef: NeMoRunTestDefinition):
         """Set environment variables based on NeMoRunTestDefinition."""
         self.final_env_vars["CLOUDAI_NEMO_TASK"] = tdef.cmd_args.task
         self.final_env_vars["CLOUDAI_NEMO_RECIPE"] = tdef.cmd_args.recipe_name
+        self.final_env_vars.setdefault("NEMO_LOG_MEMORY_USAGE", "1")
```
podkidyshev, line 42: no compute overhead here either?

src/cloudai/reports/training/parser.py
```diff
@@ -0,0 +1,242 @@
...
+    def _build_config(self, raw: dict, tr: TestRun, system: System) -> TrainingConfig:
+        """Map the raw config via CONFIG_MAPPING, then fill the CloudAI-computed fields."""
+        field_values: dict[str, Any] = {
+            field: self._get_from_dict(raw, path) for field, path in self.CONFIG_MAPPING.items()
+        }
+        config = TrainingConfig(**field_values)
+        gpus_per_node = (
+            getattr(system, "gpus_per_node", None)
+            or getattr(system, "ntasks_per_node", None)
+            or self.DEFAULT_GPUS_PER_NODE
```
podkidyshev: nit: I'm not sure it's safe to assume default=8. I would rather abort (raise exception) if gpus_per_node is not accessible

podkidyshev (changes requested):

src/cloudai/reports/training/parser.py
blugassi: Done.

src/cloudai/report_generator/training/__init__.py
blugassi: Done.

src/cloudai/reports/training/mappings.py
blugassi: Done.

src/cloudai/workloads/nemo_run/nemo_run.py
blugassi: Agent said: runs in the post-step logging hook, not forward/backward, so it only changes how often already-computed metrics are flushed (no extra compute or sync).  

I'll run with log_every_n_steps 1 vs high and get back with numbers.

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
blugassi: Agent said: only reads the CUDA memory counters (max_memory_reserved/allocated) in the post-step logging hook, not during forward/backward (O(1), no extra kernels or sync). 

I'll run with it on vs off and get back with numbers.

podkidyshev (approved):

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
podkidyshev: once ready - just write here that you confirm no overhead. I resolve and merge

src/cloudai/workloads/nemo_run/nemo_run.py
podkidyshev: once ready - just write here that you confirm no overhead. I resolve and merge

src/cloudai/workloads/nemo_run/slurm_command_gen_strategy.py
blugassi: Confirmed no overhead:
| Workload | Runtime overhead | Storage delta | Verdict |
|---|---|---|---|
| NeMoRun | 0% (1.047 s/step both; identical 7:42 job elapsed) | +56 KB TB logs | negligible |
| MegatronRun | +0.7% (280 → 282 s training; noise) | +8 KB TB logs (+28 KB report) | negligible |
| MegatronBridge | +0.3% (0.958 → 0.961 s/step) | +80 KB TB logs (+28 KB report) | negligible |

src/cloudai/workloads/nemo_run/nemo_run.py
blugassi: Confirmed no overhead:
| Workload | Runtime overhead | Storage delta | Verdict |
|---|---|---|---|
| NeMoRun | 0% (1.047 s/step both; identical 7:42 job elapsed) | +56 KB TB logs | negligible |
| MegatronRun | +0.7% (280 → 282 s training; noise) | +8 KB TB logs (+28 KB report) | negligible |
| MegatronBridge | +0.3% (0.958 → 0.961 s/step) | +80 KB TB logs (+28 KB report) | negligible |

PR #957, opened 2026-07-02
Add pre-test hook support for MegatronBridge workload

saivishal1999:
MegatronBridge's gen_exec_command uses setup_experiment.py to generate its own sbatch, bypassing CloudAI's normal pre_test injection. This adds support by submitting the pre-hook as a separate sbatch first, then passing dependency=afterok:<job_id> to setup_experiment.py via --additional_slurm_params so Slurm gates training on pre-hook success.

- _gen_pre_hook_sbatch: generates standalone sbatch for pre-hook tests, injecting main test's nnodes so srun allocates all training nodes
- _collect_additional_slurm_params: extracted from _build_launcher_parts so the base params can be built in Python and dependency appended at bash runtime after the pre-hook job ID is known
- _wrap_launcher_for_job_id_and_quiet_output: submits pre-hook sbatch, captures job ID, builds --additional_slurm_params with dependency


podkidyshev:
@saivishal1999 please resolve coderabbit comments

src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -73,13 +73,76 @@ def gen_exec_command(self) -> str:
...
+            f"#SBATCH --output={pre_hook_output.absolute() / 'stdout.txt'}",
+            f"#SBATCH --error={pre_hook_output.absolute() / 'stderr.txt'}",
+            f"#SBATCH --partition={self.system.default_partition}",
+        ]
+        if self.system.account:
+            sbatch_lines.append(f"#SBATCH --account={self.system.account}")
+        if node_list:
+            sbatch_lines.append(f"#SBATCH --nodelist={','.join(node_list)}")
+        elif num_nodes:
+            sbatch_lines.append(f"#SBATCH --nodes={num_nodes}")
```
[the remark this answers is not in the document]
podkidyshev: this is a broader issue rather than only num nodes. it also doesn't set distribution, gres/gpu-per-node and some other potential directives. what I suggest as a fix, is to update the base class `SlurmCommandGenStrategy` with extract directives generation into a sub-method. So that you could explicitly call it here to write directives into your buffer

podkidyshev (changes requested):

src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -73,13 +73,135 @@ def gen_exec_command(self) -> str:
...
+        if self.system.account:
+            sbatch_lines.append(f"#SBATCH --account={self.system.account}")
+        self._append_resource_directives(sbatch_lines, self.test_run.time_limit)
+        if self.test_run.extra_srun_args:
+            for param in self._parse_srun_args_as_slurm_params(self.test_run.extra_srun_args):
+                key, _, val = param.partition("=")
+                sbatch_lines.append(f"#SBATCH --{key}={val}" if val else f"#SBATCH --{key}")
+        sbatch_lines.append("")
+
+        assert self.test_run.pre_test is not None
```
podkidyshev, line 142: do not use asserts in production code 🙏 

https://google.github.io/styleguide/pyguide.html#244-decision

src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -73,13 +73,135 @@ def gen_exec_command(self) -> str:
...
+            )
+
+            success_var = f"SUCCESS_{idx}"
+            success_vars.append(success_var)
+
+            min_busbw = getattr(tr.test.cmd_args, "min_busbw", None)
+            if min_busbw is not None:
+                check_cmd = (
+                    f"awk '/Avg bus bandwidth/ {{ if ($NF+0 >= {min_busbw}) found=1 }}"
+                    f" END {{ exit !found }}' {node_out}/stdout_$_node.txt 2>/dev/null"
```
podkidyshev, line 166: this code effectively limits megatron-bridge to support only this pre-test hook. as it my work for your use case - it's not a GA level of code I'm afraid

src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -181,6 +313,54 @@ def _wrap_launcher_for_job_id_and_quiet_output(self, launcher_cmd: str, launcher
...
+            if capture_nodelist:
+                nodelist_lines = [
+                    "# Wait for pre-hook to reach RUNNING; exit on terminal state or timeout (1 h)",
+                    "PRE_HOOK_NODES=''",
+                    "_NODELIST_WAIT=0",
+                    "_NODELIST_TIMEOUT=3600",
+                    "while true; do",
+                    '    _state=$(squeue -j "$PRE_HOOK_JOB_ID" -h -o "%T" 2>/dev/null || true)',
+                    '    if [ "$_state" = "RUNNING" ]; then',
+                    '        PRE_HOOK_NODES=$(squeue -j "$PRE_HOOK_JOB_ID" -h -o "%N" 2>/dev/null | head -1 || true)',
```
podkidyshev, line 329: this implementation is somewhat fragile:

I'd avoid deriving the pre-hook nodelist from transient squeue state. The wrapper polls every 10s for the pre-hook job to enter RUNNING and exits if the job has already completed. A short successful pre-hook can therefore be reported as failure. Since the pre-hook sbatch itself has access to $SLURM_JOB_NODELIST, it should write that value to a known file immediately after allocation, and the wrapper should wait for/read that file before launching MBridge

src/cloudai/workloads/nccl_test/nccl.py
```diff
@@ -103,6 +103,7 @@ class NCCLCmdArgs(CmdArgs):
     stepfactor: Optional[Union[int, list[int]]] = None
     use_deepep_matrix: bool = False
     alltoallv_matrix_container_path: str = "/tmp/traffic_matrix.txt"
+    min_busbw: Optional[float] = None
```
podkidyshev, line 106: `min_busbw` is accepted by `NCCLCmdArgs`, but the regular NCCL workload success path does not enforce it. If this is intended to be a generic NCCL threshold, `NCCLTestDefinition.was_run_successful()` or `gen_srun_success_check()` should honor it. If it is only for the Megatron-Bridge pre-hook, consider keeping it out of the generic NCCL workload contract or documenting the limited scope.

podkidyshev (approved):

PR #969, opened 2026-07-14
Formalize DSE Trajectory

alexmanle:
## Summary

Introduces a pandas-native `Trajectory` for storing, analyzing, caching, and persisting DSE trial data.

Each trajectory is maintained internally as a single flat `DataFrame`. Persistence is split between:

- `trajectory.csv` for the released core format: `step`, `action`, `reward`, and `observation`
- `metadata.csv` for optional extension data such as environment parameters and logging metrics

This removes the previous entry/component, identity metadata, and writer abstractions while preserving compatibility with existing CloudAI reporting.

## Features

Initialize a trajectory with an iteration directory:

```python
trajectory = Trajectory(iteration_dir=iteration_dir)
```

An existing DataFrame can optionally provide an in-memory warm start:

```python
trajectory = Trajectory(
    iteration_dir=iteration_dir,
    dataframe=existing_dataframe,
)
```

Append a completed trial using named domains:

```python
row = trajectory.append(
    step=1,
    action={"batch_size": 32},
    reward=0.95,
    observation={"throughput": 120.0},
    env_params={"network_speed": 100},
)
```

Internally, nested mappings are recursively flattened into pandas-friendly columns:

```text
step
action.batch_size
reward
observation.throughput
env_params.network_speed
```

Core trial data is persisted to `trajectory.csv` using the released CloudAI layout:

```text
step,action,reward,observation
```

Extension data is persisted separately to a step-aligned `metadata.csv`:

```text
step,env_params.network_speed
```

`metadata.csv` is created only when extension data is present.

Search for cached trials using any subset of stored domains:

```python
cached = trajectory.find(
    action={"batch_size": 32},
    env_params={"network_speed": 100},
)
```

Both `append()` and `find()` return copied pandas `Series` objects. The complete in-memory trajectory is available as a copied DataFrame:

```python
df = trajectory.dataframe
best = df.loc[df["reward"].idxmax()]
```

CloudAIGym keeps observations as metric-name mappings internally and passes them directly to the trajectory. Observations are converted to ordered lists only at the Gym interface and persistence boundaries.

DSE reporting continues to read the released serialized `trajectory.csv` format.

## Test Plan

Validated locally with:

```bash
uv run --locked --python 3.14 --extra dev python -m pytest
uv run --locked --python 3.14 --extra dev python -m pytest --dead-fixtures
uv run --locked pre-commit run --all-files
git diff --check
```

Also tested with CloudAI internal agents and real CloudAI DSE workflows.


src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def get(self, component_type: type[ComponentT]) -> ComponentT | None:
+        """Return the component with exactly the requested type, if present."""
+        for component in self.components:
+            if type(component) is component_type:
+                return cast(ComponentT, component)
+        return None
+
+
+class TrajectoryWriter(Protocol):
```
podkidyshev: is there a reason to have a Protocol? I would really prefer a simple direct inheritance with the base class being abstract

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,435 @@
...
+                existing_fields = tuple(next(csv.reader(file), ()))
+
+        if self._fields is None:
+            if existing_fields and existing_fields != fields:
+                raise ValueError(f"trajectory file fields do not match: expected {fields}, got {existing_fields}")
+            self._fields = fields
+        elif fields != self._fields:
+            raise ValueError(f"trajectory record fields changed: expected {self._fields}, got {fields}")
+        elif existing_fields and existing_fields != self._fields:
+            raise ValueError(f"trajectory file fields do not match: expected {self._fields}, got {existing_fields}")
```
podkidyshev: I find this block somewhat complicated to read... also the condition

`existing_fields and existing_fields != self._fields`

is repeated twice.

Maybe we don't need these checks at all? I can't see how any of the checks fail when the class is used correctly (meaning that it can't be config issue)

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,435 @@
...
+            len(self),
+            str(file_type),
+            component_names,
+        )
+
+    @staticmethod
+    def _create_writer(
+        iteration_dir: Path | Callable[[], Path] | None,
+        file_type: Literal["csv", "jsonl"],
+    ) -> TrajectoryWriter | None:
```
podkidyshev: what's the purpose of the Trajectory instance without a writer?

podkidyshev:
Despite the implementation is technically correct, I find it too complicated for the goal it achieves (apparently the PR summary misses the business goal!)

As far as I could understand, the goal is to make the trajectory a better container:

1. store any data (per step)
2. derive insights from the data

My alternative suggestion for 2. would be to use the standard thing in data analysis - `pandas` with its dataframes. As long as we provide something that could be fed to `pd.read_csv` or `pd.read_json` - we solve the point 2.

For the point 1. - our container should be just columnar flat data. So for any additional data that we'd like to populate into trajectory, we could use format `<domain>.<data point key>`, e.g. `env_params.some_param`

what do you think?

podkidyshev (changes requested):

alexmanle:
> Despite the implementation is technically correct, I find it too complicated for the goal it achieves (apparently the PR summary misses the business goal!)
> 
> As far as I could understand, the goal is to make the trajectory a better container:
> 
> 1. store any data (per step)
> 2. derive insights from the data
> 
> My alternative suggestion for 2. would be to use the standard thing in data analysis - `pandas` with its dataframes. As long as we provide something that could be fed to `pd.read_csv` or `pd.read_json` - we solve the point 2.
> 
> For the point 1. - our container should be just columnar flat data. So for any additional data that we'd like to populate into trajectory, we could use format `<domain>.<data point key>`, e.g. `env_params.some_param`
> 
> what do you think?

@podkidyshev I agree actually that this solution is to much for the problem. I think as long as it's easy to append new data and search elements (for caching) your proposed alternative is great with me!

src/cloudai/configurator/cloudai_gym.py
```diff
@@ -277,83 +257,21 @@ def encode_env_params(self, env_params: dict[str, Any]) -> Dict[str, Any]:
...
-            return True
-
-        else:
-            return left == right
+        path = self.trajectory.output_path
+        if path is None:
+            raise RuntimeError("trajectory persistence is not configured")
+        return path
+
+    def get_cached_trajectory_result(self, action: Any, env_params: dict[str, Any]) -> pd.Series | None:
```
podkidyshev:
```suggestion
    def get_cached_trajectory_result(self, action: Any, env_params: dict[str, Any]) -> "pd.Series" | None:
```

as the pandas is imported conditionally for type checking only. that should fix the test

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,180 @@
...
+class Trajectory:
+    """An ordered DataFrame of DSE steps persisted to ``trajectory.csv``."""
+
+    file_name = "trajectory.csv"
+
+    def __init__(
+        self,
+        dataframe: pd.DataFrame | None = None,
+        *,
+        iteration_dir: Path | Callable[[], Path] | None = None,
```
podkidyshev: can this really be `None`? and does it need the `Callable` functionality?

src/cloudai/systems/slurm/single_sbatch_runner.py
```diff
@@ -205,27 +208,29 @@ def handle_dse(self):
...
                 next_tr.step = idx
                 next_tr.output_path = self.get_job_output_path(next_tr)
 
-                rewards = None
-                agent_class = registry.get_agent(next_tr.test.agent)
-                agent_config_data = next_tr.test.agent_config or {}
-                agent_config = agent_class.get_config_class()(**agent_config_data)
-                rewards = agent_config.rewards
+                if not next_tr.test.constraint_check(next_tr, self.system):
+                    continue
```
podkidyshev, line 230: nice catch! (and overall updating single-sbatch to support env params)

podkidyshev:
@alexmanle this look far better to me! I really like the new trajectory file look

please make sure to thoroughly test these changes, including single-sbatch and reporter 🙏  potentially also let other team members be aware of it.

alexmanle:
> @alexmanle this look far better to me! I really like the new trajectory file look
> 
> please make sure to thoroughly test these changes, including single-sbatch and reporter 🙏 potentially also let other team members be aware of it.

@podkidyshev The tests succeeded on Lyris cluster.
See the accompanying [report](https://drive.google.com/file/d/1ZrliJk2mX11hzC_YbtoDRZKQPJ9wxD_W/view?usp=sharing) for more details.

alexmanle:
FYI
@srivatsankrishnan @rutayan-nv @saivishal1999

Keeping you in the loop on core change.

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,237 @@
...
+        raise ValueError(f"trajectory values produce duplicate column: {key}")
+    record[key] = deepcopy(value)
+
+
+def _copy_dataframe(dataframe: pd.DataFrame) -> pd.DataFrame:
+    """Copy a DataFrame and recursively snapshot object cells."""
+    copied = dataframe.copy(deep=True).astype(object)
+    for row_index in range(dataframe.shape[0]):
+        for column_index in range(dataframe.shape[1]):
+            copied.iat[row_index, column_index] = deepcopy(dataframe.iat[row_index, column_index])
```
podkidyshev, line 216: what possible mutable objects are to be supported in the trajectory?

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,237 @@
...
+        previous_step: int | None = None
+        for step in self._dataframe["step"]:
+            self._validate_step(step)
+            if previous_step is not None and step <= previous_step:
+                raise ValueError(f"trajectory steps must increase: last step is {previous_step}, got {step}")
+            previous_step = int(step)
+
+    @staticmethod
+    def _validate_step(step: object) -> None:
+        if isinstance(step, bool) or not isinstance(step, Integral) or step < 1:
```
podkidyshev:
```suggestion
        if not isinstance(step, int) or step < 1:
```

do we really need the bool and Integral checks?

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,237 @@
...
+    if write_header:
+        return True
+    with path.open(newline="") as file:
+        existing_fields = tuple(next(csv.reader(file), ()))
+    if existing_fields != fields:
+        raise ValueError(f"trajectory file fields do not match: expected {fields}, got {existing_fields}")
+    return False
+
+
+def _flatten_value(record: dict[str, object], key: str, value: object) -> None:
```
podkidyshev: I would rather keep the function only responsible for flattening value into resulting dict[str, any].

`record[key] = deepcopy(value)` is an object mutation inside a recursive function = thus making it unnecessary more complicated

src/cloudai/configurator/trajectory.py
```diff
@@ -0,0 +1,237 @@
...
+            if previous_step is not None and step <= previous_step:
+                raise ValueError(f"trajectory steps must increase: last step is {previous_step}, got {step}")
+            previous_step = int(step)
+
+    @staticmethod
+    def _validate_step(step: object) -> None:
+        if isinstance(step, bool) or not isinstance(step, Integral) or step < 1:
+            raise ValueError(f"trajectory step must be a positive integer; got {step}")
+
+    def _persist(self, row: pd.Series, *, action: object, reward: object, observation: object) -> None:
```
podkidyshev: I find it troublesome to follow the flow of this function...

How about organizing the persistence a bit differently:

* in `append` build two series: `row_trajectory` and `row_metadata`
* have a new function `append_csv_row` that could work with both. It would take `row: Series` and `path: Path`, using the `_validate_csv_header`

src/cloudai/configurator/trajectory.py
alexmanle: > what possible mutable objects are to be supported in the trajectory?

Mainly lists, such as logging or metadata values containing multiple samples. Mappings are flattened, but list-valued leaves remain mutable, so we copy them to prevent later changes from affecting stored data.

podkidyshev (approved):

Issue #980, opened 2026-07-27
Sleep workload never checks whether the command actually succeeded; always reports PASSED


shreyaskommuri:
# Sleep workload never checks whether the command actually succeeded; always reports PASSED

## Describe the Bug

`SleepTestDefinition` never overrides `was_run_successful()`, so it inherits<br>the base `TestDefinition` default, which unconditionally reports success:

```python
# src/cloudai/models/workload.py
def was_run_successful(self, tr: TestRun) -> JobStatusResult:
    return JobStatusResult(is_successful=True)
```

Every other workload I checked overrides this with a real check —<br>`fio.py`, `vllm.py`, `nccl.py`, `nixl_ep.py`, `dynamo_mocker.py`,<br>`nemo_run.py`, and `sglang.py` all implement `was_run_successful()`.<br>`src/cloudai/workloads/sleep/sleep.py` does not. As a result, a Sleep test<br>is reported `PASSED` regardless of whether the underlying command actually<br>ran, produced output, or exited non-zero.

This came up while validating CloudAI through a small wrapper project,<br>[CloudAI Autotune](<https://github.com/shreyaskommuri/CloudAI-Autotune>).<br>Autotune calls the current CloudAI CLI as its execution layer; I was running<br>a real (non-dry-run) `cloudai run` through Autotune to check whether a<br>genuinely-failing command would be surfaced correctly, and it wasn't.

Because `was_run_successful` lives on the workload's `TestDefinition`<br>rather than on any system-specific class, this isn't a standalone-only<br>issue — it affects Sleep identically on every system it supports (Slurm,<br>Kubernetes, Standalone, per the support matrix in the README).

## Steps to Reproduce

Version/context where observed: local CloudAI checkout based on<br>`origin/main` (commit `28c317c9`).

Create a Sleep test config with a command that will genuinely fail (an<br>invalid `seconds` value produces `sleep -1`, which fails immediately with<br>`sleep: illegal option -- 1` on this system):

```toml
# test/sleep_bad.toml
name = "sleep_bad"
description = "sleep test"
test_template_name = "Sleep"

[cmd_args]
seconds = -1
```

```toml
# test_scenario/sleep_bad_scenario.toml
name = "sleep-bad-scenario"

[[Tests]]
id = "Tests.sleepbad"
test_name = "sleep_bad"
time_limit = "00:01:00"
```

Run it for real (not dry-run) through Autotune, which just shells out to the<br>`cloudai` CLI:

```bash
[...]

PR #981, opened 2026-07-27
Make Sleep's was_run_successful check the real exit code

shreyaskommuri:
## Summary
- `SleepTestDefinition` never overrode `was_run_successful()`, so it fell back to the base `TestDefinition` default, which unconditionally returns `is_successful=True` — no check of exit code, output, or anything else
- Every other workload I checked (`fio.py`, `vllm.py`, `nccl.py`, `nixl_ep.py`, `dynamo_mocker.py`, `nemo_run.py`, `sglang.py`) implements a real `was_run_successful()`. Sleep did not
- `SleepStandaloneCommandGenStrategy.gen_exec_command()` now redirects stdout/stderr to files and captures the real exit code, following the same pattern `FioStandaloneCommandGenStrategy` already uses for stdout/stderr redirection
- `SleepTestDefinition.was_run_successful()` reads that exit code file. If it's absent (Slurm/LSF/Kubernetes command-gen strategies don't capture one yet), it falls back to the previous always-true behavior — additive, not a behavior change for those systems
- Scoped to standalone: I don't have Slurm/LSF/Kubernetes systems available to verify a fix there, so those still have the same underlying gap in principle; the issue documents it generally for whoever picks that up

Separately noticed, **out of scope for this PR**: even with `was_run_successful` now correctly returning `FAILED`, `handle_dry_run_and_run`'s non-DSE path (`src/cloudai/cli/handlers.py`) unconditionally returns `0` regardless of job outcome, so the overall `cloudai run` process exit code still doesn't reflect a failed test. That's a broader gap affecting every workload's exit code, not Sleep-specific, and not addressed here.

## How this was found
Found while validating CloudAI through a small wrapper project, [CloudAI Autotune](https://github.com/shreyaskommuri/CloudAI-Autotune). I was running a real (non-dry-run) `cloudai run` through Autotune to check whether a genuinely-failing command would be surfaced correctly — it wasn't. Reproduced directly against `cloudai run` too (not just through the wrapper). Filed as #980.

Issue: #980

## Test Plan
- `uv run pytest -q` — 1767 passed, 18 skipped
- `uv run ruff check` / `uv run ruff format --check` on the changed files — clean
- Manually re-ran the exact broken repro from the issue against this fix: a standalone Sleep test with `seconds = -1` (which fails with `sleep: illegal option -- 1`, exit code 1). Before this fix: results table showed `PASSED`. After: results table correctly shows `FAILED` with `sleep exited with code 1. Check .../stderr.txt for details.`

AI was used for context and guidance while investigating, implementing, and testing this fix.

### DIFF
--- a/src/cloudai/workloads/sleep/sleep.py
+++ b/src/cloudai/workloads/sleep/sleep.py
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
-# Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,8 +15,10 @@
 # limitations under the License.
 
 
-from cloudai.core import Installable
+from cloudai.core import Installable, JobStatusResult, TestRun
 from cloudai.models.workload import CmdArgs, TestDefinition
+
+EXIT_CODE_FILE_NAME = "exit_code.txt"
 
 
 class SleepCmdArgs(CmdArgs):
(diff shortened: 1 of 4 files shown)


shreyaskommuri:
Addressed both current CodeRabbit comments:
- `sleep.py`: `was_run_successful()` now catches `OSError`/`UnicodeDecodeError` from reading the exit-code file and returns a clean `FAILED` status instead of letting the exception propagate.
- `standalone_command_gen_strategy.py`: made `gen_exec_command`'s `store` parameter keyword-only (ruff FBT001/FBT002), matching the only call site which already used keyword form.

(The earlier Copilot comment about `store_test_run`'s signature was against my first commit — already fixed in the follow-up commit before Copilot's review ran, per the commit SHA it's anchored to.)

Re-ran `pytest -q` (1769 passed), `pyright` (0 errors), `ruff check`/`ruff format --check` — all clean.

PR #982, opened 2026-07-28
Add aggregate mode support to AIDynamo workload

saivishal1999:
In aggregate mode, a single vLLM worker handles both prefill and decode on the same node(s). This is opt-in and keeps the existing disaggregated
  behavior as the default.

  Changes:
  - `ai_dynamo.py`: add `mode` to `AIDynamoArgs` with supported values `aggregate` and `disaggregated`; keep `disaggregated` as the default;
  normalize legacy spelling `disaggregate` to `disaggregated`; restrict aggregate mode to vLLM.
  - `ai_dynamo.py`: skip decode-worker model population and split-worker constraint checks only when `mode = "aggregate"`.
  - `slurm_command_gen_strategy.py`: preserve legacy/disaggregated command generation by omitting `--dynamo-mode` for default configs; omit
  `--decode-*` args only when `mode = "aggregate"`.
  - `ai_dynamo.sh`: initialize role defaults centrally; support zero decode nodes in aggregate mode; use the prefill node as `frontend-node` when no
  decode node exists.
  - Tests: add regression coverage that default/disaggregated configs still emit legacy prefill/decode args and do not emit `--dynamo-mode`; add
  aggregate-mode coverage for decode-arg omission and node-count behavior.

  Validation:
  - Existing/disaggregated configs remain backward compatible: default mode is unchanged and generated command lines still include prefill/decode
  worker args.
  - Tested on Lyris GB200:
    - aggregate: 1 node, prefill 1/1, no decode worker
    - disaggregated: 2 nodes, prefill 1/1, decode 1/1
  - Tested on Daria:
    - aggregate mode dse ran successfully

### DIFF
--- a/src/cloudai/workloads/ai_dynamo/ai_dynamo.py
+++ b/src/cloudai/workloads/ai_dynamo/ai_dynamo.py
@@ -161,6 +161,7 @@ class AIDynamoArgs(BaseModel):
 
     model: str = "Qwen/Qwen3-0.6B"
     backend: Literal["vllm", "sglang", "sglang_dsr1"] = "vllm"
+    mode: Literal["aggregate", "disaggregate"] = "disaggregate"
     endpoint: str = Field(default="v1/chat/completions")
     connector: Optional[str | list[str]] = None
 
@@ -235,12 +236,14 @@ class AIDynamoArgs(BaseModel):
         """Populate prefill/decode args."""
         if self.backend.lower() == "vllm":
             self.prefill_worker.args.model = self.model
-            self.decode_worker.args.model = self.model
+            if self.mode == "disaggregate":
+                self.decode_worker.args.model = self.model
         elif self.backend.lower() in ["sglang", "sglang_dsr1"]:
             self.prefill_worker.args.model_path = self.model
-            self.decode_worker.args.model_path = self.model
             self.prefill_worker.args.served_model_name = self.model
-            self.decode_worker.args.served_model_name = self.model
+            if self.mode == "disaggregate":
+                self.decode_worker.args.model_path = self.model
+                self.decode_worker.args.served_model_name = self.model
         else:
             raise ValueError(f"Invalid backend: {self.backend}")
 
@@ -577,6 +580,10 @@ class AIDynamoTestDefinition(TestDefinition):
         return JobStatusResult(workloads_successful and accuracy_successful)
 
     def constraint_check(self, tr: TestRun, system: Optional[System]) -> bool:
+        if tr.test.cmd_args.dynamo.mode == "aggregate":
+            logging.info("constraint_check skipped: aggregate mode has no prefill/decode split")
+            return True
+
         prefill_worker = tr.test.cmd_args.dynamo.prefill_worker
         decode_worker = tr.test.cmd_args.dynamo.decode_worker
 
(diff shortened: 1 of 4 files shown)


src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -162,9 +162,17 @@ class AIDynamoArgs(BaseModel):
 
     model: str = "Qwen/Qwen3-0.6B"
     backend: Literal["vllm", "sglang", "sglang_dsr1"] = "vllm"
+    mode: Literal["aggregate", "disaggregated"] = "disaggregated"
     endpoint: str = Field(default="v1/chat/completions")
     connector: Optional[str | list[str]] = None
 
+    @field_validator("mode", mode="before")
+    @classmethod
+    def normalize_mode(cls, v: str) -> str:
```
podkidyshev, line 171: this is too much of a syntax sugar

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -162,9 +162,17 @@ class AIDynamoArgs(BaseModel):
 
     model: str = "Qwen/Qwen3-0.6B"
     backend: Literal["vllm", "sglang", "sglang_dsr1"] = "vllm"
+    mode: Literal["aggregate", "disaggregated"] = "disaggregated"
```
podkidyshev, line 165: why not just `agg` / `disagg`? this will also eliminate the potential confusion between `-e` and `-ed` suffix (apparently your typing has `aggregatE` but `disaggregatED`)

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -234,14 +242,21 @@ def validate_connector(cls, v: str | list[str] | None) -> str | list[str] | None
     @model_validator(mode="after")
     def populate_prefill_decode_args(self) -> "AIDynamoArgs":
         """Populate prefill/decode args."""
+        if self.mode == "aggregate" and self.backend.lower() != "vllm":
+            raise ValueError("AI Dynamo aggregate mode is currently supported only for the vLLM backend")
```
podkidyshev, line 246: let's omit this validation. if one day aggregate mode will be supported by sglang/trt-llm I don't want CloudAI code to be the blocker

podkidyshev:
I'm curious about the overall need for this PR as it is possible to run agg inference for both vLLM and SGLang with the given scenario (reference `conf/experimental/ai_dynamo/test` as tests-dir)

<details>
<summary>Agg vLLM + SGLang scenario TOML</summary>

```toml
name = "dynamo-aggregate-multinode"
job_status_check = false

[[Tests]]
id = "test.aggregate.vllm.multinode"
test_name = "vLLM"
time_limit = "00:10:00"

  [Tests.cmd_args]

    # Legacy aggregate topology: disable the dedicated prefill workers and
    # run combined prefill/decode workers from the decode worker configuration.
    [Tests.cmd_args.dynamo.prefill_worker]
    num-nodes = 0

    [Tests.cmd_args.dynamo.decode_worker]
    num-nodes = 2

[[Tests]]
id = "test.aggregate.sglang.multinode"
test_name = "sglang"
time_limit = "00:10:00"

  [Tests.cmd_args]

    # Legacy aggregate topology: disable the dedicated prefill workers and
    # run combined prefill/decode workers from the decode worker configuration.
    [Tests.cmd_args.dynamo.prefill_worker]
    num-nodes = 0

    [Tests.cmd_args.dynamo.decode_worker]
    num-nodes = 2

      [Tests.cmd_args.dynamo.decode_worker.args]
      # Override the decode-only value inherited from test/sglang.toml.
      disaggregation-mode = "null"
      tensor-parallel-size = 8
```

</details>

podkidyshev (changes requested):

PR #990, opened 2026-08-06
Add opt-in compute-node local Hugging Face model staging

juntaowww:
## Motivations
Loading the large model from the shared NFS can be slow, so for faster loading, we add an opt-in option that allow user to copy the model from NFS to local storage of the compute nodes (the model would still be downloaded to the shared NFS first), and in the subsequent run on the same set of nodes, the model can be directly loaded from local storage.

## Summary
Add optional, per-test Hugging Face model staging for Slurm workloads.
- Copy required models from the shared `hf_home_path` to compute-node local storage before workload containers start.
- Reuse complete local caches and repair incomplete copies using per-model locking and manifest validation.
- Mount the local cache for vLLM, SGLang, and AI Dynamo workloads while preserving existing behavior when `hf_local_home_path` is unset.
- Support regular and single-sbatch execution, including independently configured pre-test and post-test workloads.
- Add documentation and a Dynamo/vLLM Qwen3-0.6B example scenario.

## Test Plan
Tests + manual run.

## Additional Notes
- The feature is opt-in through the test-level `hf_local_home_path` field.
- The shared Hugging Face cache remains the installation source and must be readable from compute nodes.
- The local path must be absolute, writable, different from `hf_home_path`, and consistent across allocated nodes.
- Compute nodes must provide the `flock` utility.
- Existing workflows remain unchanged when local staging is not configured.


conf/common/system/example_slurm_cluster.toml
```diff
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
-# Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
podkidyshev, line 2: you could try to squash the commit when a change occurred in this file to get rid of this redundant copyright update

podkidyshev:
@juntaowww could you please elaborate on the end goal (problem of yours) you solve with this change?

tests/workloads/vllm/test_command_gen_strategy_slurm.py
```diff
@@ -75,6 +77,127 @@ def test_container_mounts(vllm_cmd_gen_strategy: VllmSlurmCommandGenStrategy) ->
...
+    source_ref.write_text("revision-1")
+
+    stage_script = _local_hf_stage_script(vllm_cmd_gen_strategy, shared_hf_home, local_hf_home)
+    subprocess.run(["bash", "-lc", stage_script], check=True, capture_output=True, text=True)
+
+    local_ref = local_hf_home / "hub" / model_cache / "refs" / "main"
+    assert local_ref.read_text() == "revision-1"
+
+    source_ref.write_text("revision-2")
+    second_run = subprocess.run(["bash", "-lc", stage_script], check=True, capture_output=True, text=True)
```
podkidyshev: I'm not sure we could accept such test here... it's rather integration test than unit. e.g. it requires `flock` to be installed on the OS to run it

let's keep it for now until we finalize the rest of the PR review

juntaowww:
> @juntaowww could you please elaborate on the end goal (problem of yours) you solve with this change?

Loading the large model from the shared NFS can be slow, so add an option that can copy the model from NFS to local storage of the compute nodes (the model would still be downloaded to the shared NFS first), and in the subsequent run on the same nodes, the model can be directly loaded from local storage.

tests/workloads/vllm/test_command_gen_strategy_slurm.py
juntaowww: yes, I am fixing this, the `subprocess` is really a bad practice here..

src/cloudai/systems/slurm/slurm_command_gen_strategy.py
```diff
@@ -105,6 +106,160 @@ def container_mounts(self) -> list[str]:
...
+        self,
+        models: Iterable[HFModel],
+        local_hf_home_path: Path,
+        node_spec: tuple[int, list[str]] | None = None,
+    ) -> str:
+        """Generate a host-side Slurm step that stages HF models on every compute node."""
+        models_to_stage = sorted(set(models), key=lambda model: model.model_name)
+        configured_shared_hf_home = shlex.quote(str(self.system.hf_home_path.absolute()))
+        configured_local_hf_home = shlex.quote(str(local_hf_home_path.absolute()))
+        model_cache_dirs = " ".join(shlex.quote(model.cache_dir_name) for model in models_to_stage)
```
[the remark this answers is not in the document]
juntaowww: invalid, that is property no method

src/cloudai/systems/slurm/slurm_command_gen_strategy.py
```diff
@@ -105,6 +106,160 @@ def container_mounts(self) -> list[str]:
...
+            f"--error={output_path}/hf-stage-node-%N-stderr.txt",
+            "bash",
+            "-lc",
+            shlex.quote(stage_script),
+        ]
+        return " \\\n  ".join(command_parts) + " || exit 1"
+
+    def gen_hf_model_staging_commands(
+        self,
+        test_runs: Iterable[TestRun],
```
podkidyshev, line 249: does it really need to have this arg? the only usage of this method (outside tests) I see in the `_write_sbatch_script` below (519-520 lines)

src/cloudai/systems/slurm/slurm_command_gen_strategy.py
```diff
@@ -105,6 +106,160 @@ def container_mounts(self) -> list[str]:
...
+            "--mpi=none",
+            f"--nodes={num_nodes}",
+            *([] if not node_list else [f"--nodelist={','.join(node_list)}"]),
+            f"--ntasks={num_nodes}",
+            "--ntasks-per-node=1",
+            f"--output={output_path}/hf-stage-node-%N-stdout.txt",
+            f"--error={output_path}/hf-stage-node-%N-stderr.txt",
+            "bash",
+            "-lc",
+            shlex.quote(stage_script),
```
podkidyshev, line 243: this is quite an inline script... I suggest:

* moving the bash script into the `HFModel` class
* calling its generation from here, saving into some file in the test run output folder (which will be mounted)  as shell file and running it with srun here

src/cloudai/systems/slurm/slurm_command_gen_strategy.py
```diff
@@ -105,6 +106,160 @@ def container_mounts(self) -> list[str]:
...
+
+        num_nodes, node_list = node_spec or self.get_cached_nodes_spec()
+        output_path = self.test_run.output_path.absolute()
+        command_parts = [
+            "srun",
+            "--export=ALL",
+            "--mpi=none",
+            f"--nodes={num_nodes}",
+            *([] if not node_list else [f"--nodelist={','.join(node_list)}"]),
+            f"--ntasks={num_nodes}",
```
podkidyshev, line 237: all 3 could be replaced with

```
--ntasks="$SLURM_JOB_NUM_NODES"
```

(keeping in mind the `--ntasks-per-node=1` below)

podkidyshev (changes requested):

src/cloudai/systems/slurm/slurm_command_gen_strategy.py
podkidyshev: I was wrong, I see the changes in the single-sbatch-runner

PR #991, opened 2026-08-06
[Reporting] SOL for NCCL/NIXL in comparison report

podkidyshev:
## Summary
Flexible SOL configuration + using it in comparison reports for NCCL and NIXL

## Test Plan
- Automated CI
- Manual runs 

### Trade-off

The complexity of comparison reporters has grown significantly. I attempted many solutions that both produce clear code and stay under the sane limit of code changes. Unfortunately, **we need a complete rebuild of data structures**. 

I'll create an internal ticket for this once we pass this PR. The PR change is a P1 requirement for another team thus it cannot take forever to iterate on it...

### Example: NIXL

image

image

### Example: NCCL

image

image


src/cloudai/metrics.py
```diff
@@ -0,0 +1,254 @@
...
+        return self.target.value if self.target is not None else None
+
+    @property
+    def attainment(self) -> float | None:
+        """Return the measured performance relative to SOL."""
+        if self.target is None:
+            return None
+        if self.observation.metric.direction is OptimizationDirection.MAXIMIZE:
+            return self.observation.value / self.target.value
+        return self.target.value / self.observation.value
```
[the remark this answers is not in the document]
podkidyshev: Fixed in 692c4d26. Zero-valued MINIMIZE observations now produce unavailable attainment instead of raising during report generation.

src/cloudai/report_generator/comparison_report_metric.py
```diff
@@ -0,0 +1,140 @@
...
+                    result[f"{data_column} SOL"] = None
+                    result[f"{data_column} % SOL"] = None
+                    continue
+                points = frame
+                for dimension, value in zip(series, series_key, strict=True):
+                    points = points[points[dimension] == value]
+                points = points.groupby(x_dimension, as_index=False).first().set_index(x_dimension)
+                result[data_column] = result[x_column].map(points["measured"])
+                result[f"{data_column} SOL"] = result[x_column].map(points["sol"])
+                result[f"{data_column} % SOL"] = result[x_column].map(points["attainment"])
```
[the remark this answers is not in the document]
podkidyshev: Not changed after review. These two column names intentionally follow a transparent local suffix convention. Calling protected ComparisonReport helpers from the metric producer would add cross-class coupling without reducing meaningful complexity, and a rename would already require coordinated producer/consumer changes.

src/cloudai/report_generator/comparison_report_metric.py
```diff
@@ -0,0 +1,140 @@
...
+            for series_key, data_column in zip(series_keys, data_columns, strict=True):
+                if frame.empty:
+                    result[data_column] = None
+                    result[f"{data_column} SOL"] = None
+                    result[f"{data_column} % SOL"] = None
+                    continue
+                points = frame
+                for dimension, value in zip(series, series_key, strict=True):
+                    points = points[points[dimension] == value]
+                points = points.groupby(x_dimension, as_index=False).first().set_index(x_dimension)
```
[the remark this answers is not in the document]
podkidyshev: Fixed in 692c4d26. NIXL comparison sections now include batch_size as a series dimension, so batch sweeps are preserved rather than collapsed by groupby().first().

src/cloudai/report_generator/comparison_report.py
```diff
@@ -420,8 +458,28 @@ def _build_line_datasets_v2(self, section: ComparisonSection) -> tuple[list[str]
...
+                    sol_data = [sol_by_label.get(label) for label in labels]
+                datasets.insert(
+                    (metric_idx + 1) * len(section.group.items) + inserted,
+                    {
+                        "label": f"{data_column} · SOL" if include_metric else "SOL",
+                        "data": sol_data,
+                        "is_sol": True,
+                    },
+                )
+                inserted += 1
```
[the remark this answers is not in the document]
podkidyshev: Partially adopted in 692c4d26. The redundant insert-index arithmetic is replaced with append. Numeric x revalidation was intentionally omitted: the only metric comparison reporters in this PR use size_bytes, which MetricObservation validates as an integer before section construction.

src/cloudai/report_generator/comparison_report.py
```diff
@@ -598,11 +698,38 @@ def create_chart(
...
         p.legend.location = "top_left"
         p.legend.click_policy = "hide"
 
-        y_max = max(df[col].max() for df in dfs for col in data_columns if not df.empty)
-        y_min = min(df[col].min() for df in dfs for col in data_columns if not df.empty)
+        measured_max = max(df[col].max() for df in dfs for col in data_columns if not df.empty)
+        measured_min = min(df[col].min() for df in dfs for col in data_columns if not df.empty)
+        y_max = max([measured_max, *sol_values])
+        y_min = min([measured_min, *sol_values])
         p.y_range = lazy.bokeh_models.Range1d(start=y_min * -1 * y_max * 0.01, end=y_max * 1.1)
```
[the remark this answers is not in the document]
podkidyshev: Fixed in 692c4d26. Legacy chart bounds now collect only finite numeric measured values, include SOL values, and return without assigning an invalid Range1d when no finite bounds exist.

src/cloudai/workloads/nccl_test/nccl.py
```diff
@@ -188,3 +189,40 @@ def was_run_successful(self, tr: TestRun) -> JobStatusResult:
...
+                            dimensions,
+                        ),
+                        cloudai.metrics.MetricObservation(
+                            cloudai.metrics.BANDWIDTH,
+                            float(row[bandwidth_idx]),
+                            {**dimensions, "bandwidth_basis": "bus"},
+                        ),
+                    ]
+                )
+        return observations
```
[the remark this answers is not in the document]
podkidyshev: Fixed in 692c4d26. NCCL rows now convert all required numeric fields before creating observations and skip malformed rows while retaining valid benchmark points.

src/cloudai/workloads/nixl_bench/nixl_bench.py
```diff
@@ -55,3 +59,34 @@ def was_run_successful(self, tr: TestRun) -> JobStatusResult:
...
+        for row in df.itertuples(index=False):
+            row = cast(Any, row)
+            dimensions: cloudai.metrics.MetricDimensions = {
+                "operation": str(getattr(self.cmd_args, "op_type", "default")).lower(),
+                "size_bytes": int(row.block_size),
+                "batch_size": int(row.batch_size),
+                "backend": str(getattr(self.cmd_args, "backend", "default")).lower(),
+                "source_memory": str(getattr(self.cmd_args, "initiator_seg_type", "default")).lower(),
+                "target_memory": str(getattr(self.cmd_args, "target_seg_type", "default")).lower(),
+            }
```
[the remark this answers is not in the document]
podkidyshev: Partially adopted in 692c4d26. Explicit None values now map to default instead of the string none. Direct typed access was intentionally not added because these NIXL CLI options are extensible Pydantic extras, not declared NIXLBenchCmdArgs fields; getattr is the correct contract here.

tests/workloads/nixl_bench/test_report.py
```diff
@@ -71,6 +74,37 @@ def test_nixl_bench_report_parsing(tmp_path: Path, sample: str, exp_latency: lis
...
+
+    legacy_html = tmp_path.joinpath("nixl_comparison.html").read_text()
+    assert "SOL" in legacy_html
+    assert "100.0" in legacy_html
+    assert "58.0%" in legacy_html
+
+    html = tmp_path.joinpath("nixl_comparison_v2.html").read_text()
+    assert "Bandwidth" in html
+    assert "100.0" in html
+    assert "58.0%" in html
```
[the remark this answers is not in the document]
podkidyshev: Fixed in 692c4d26 without adding another test. The existing HTML-generation test now compares two runs with different SOL values, verifies no shared SOL chart is emitted, confirms latency omits SOL columns, checks the rendered legacy SOL header, and preserves expected target/attainment values.

src/cloudai/report_generator/comparison_report.py
```diff
@@ -367,6 +367,33 @@ def _numeric_value(value: Any) -> float | None:
...
+        points = []
+        for df in dfs:
+            for data_column in data_columns:
+                sol_column = cls._sol_column(data_column)
+                if x_column in df.columns and sol_column in df.columns and df[sol_column].notna().any():
+                    points.append(df[[x_column, sol_column]].rename(columns={sol_column: "sol"}))
+        if not points:
+            return []
+        curve = lazy.pd.concat(points).dropna(subset=["sol"]).drop_duplicates(x_column).sort_values(x_column)
+        return list(curve[[x_column, "sol"]].itertuples(index=False, name=None))
```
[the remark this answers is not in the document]
podkidyshev: this is the intended flow. One sol curve is chosen and then assessment is run only against this curve (thus having only one possible candidate per observed metric + x dimension)

src/cloudai/metrics.py
```diff
@@ -0,0 +1,256 @@
...
+    MAXIMIZE = "maximize"
+    MINIMIZE = "minimize"
+
+
+@dataclass(frozen=True)
+class DimensionDefinition:
+    """A typed semantic dimension shared by metric observations and SOL selectors."""
+
+    key: str
+    label: str
```
jj10306, line 45: what's the difference b/t key and label here? can you give an example where these would differ?

src/cloudai/metrics.py
```diff
@@ -0,0 +1,256 @@
...
+            available = ", ".join(sorted(cls._metrics))
+            raise ValueError(f"Unknown SOL metric '{key}'. Available metrics: {available}") from exc
+
+    @classmethod
+    def dimension(cls, key: str) -> DimensionDefinition:
+        try:
+            return cls._dimensions[key]
+        except KeyError as exc:
+            available = ", ".join(sorted(cls._dimensions))
+            raise ValueError(f"Unknown metric dimension '{key}'. Available dimensions: {available}") from exc
```
jj10306, line 83: what are your thoughts on renaming the `metric`/`dimension` methods to `get_metric|dimension` to more clearly indicate these are "getters" and then add `register_metric|dimension` / `unregister_metric|dimension`?
 This would provide better encapsulation and avoid manually modifying the "private" class vars the way is currently done below

src/cloudai/metrics.py
```diff
@@ -0,0 +1,256 @@
...
+MetricCatalog._dimensions = {
+    dimension.key: dimension
+    for dimension in (
+        SIZE_BYTES,
+        BATCH_SIZE,
+        OPERATION,
+        PLACEMENT,
+        BANDWIDTH_BASIS,
+        BACKEND,
+        SOURCE_MEMORY,
```
jj10306: my understanding is these metric + dimensions are specific to nccl/nixl tests, but are being defined at a global (not workload-specific level) - how would this work if another workload (not nccl/nixl) were to run and want SOL comparison? 

Basically i'm wondering if the metric registry creation should be made in some workload specific location? 

let me know if this makes sense

src/cloudai/metrics.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+MetricSOLConfig: TypeAlias = dict[str, list[SOLTarget]]
+
+
+def _selectors_overlap(first: MetricDimensions, second: MetricDimensions) -> bool:
+    return all(first[key] == second[key] for key in first.keys() & second.keys())
+
+
+def parse_sol_spec(value: dict[str, Any] | None) -> MetricSOLConfig:
```
jj10306, line 157: do we have unittests covering this new parsing logic?

src/cloudai/metrics.py
```diff
@@ -0,0 +1,256 @@
...
+                if len(target.match) == len(other.match) and _selectors_overlap(target.match, other.match):
+                    raise ValueError(
+                        f"Ambiguous SOL targets for metric '{metric_key}': {target.match} and {other.match} "
+                        "can match the same observation with equal specificity"
+                    )
+        parsed[metric_key] = targets
+    return parsed
+
+
+def merge_sol_configs(*configs: MetricSOLConfig | None) -> MetricSOLConfig:
```
jj10306, line 181: do we have unittests covering this new merging logic?

src/cloudai/report_generator/comparison_report.py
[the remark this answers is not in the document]
jj10306: @podkidyshev I have the same question as coderrabbit here:
Looking at the example NCCL config you included in doc/reporting.rst:
```
[[sol.bandwidth]]
value = 250.0
match = { operation = "all_reduce", placement = "out_of_place", bandwidth_basis = "bus" }

[[sol.bandwidth]]
value = 300.0
match = { operation = "all_reduce", placement = "in_place", bandwidth_basis = "bus" }
```

My understanding:
`parse_sol_spec` enforces uniqueness per full dimension set and these differ on `placement`, but placement isn't the x dimension, so this `all_reduce` run has two candidates at every `size_bytes` and they become two series. `drop_duplicates(x_column)` would keep whichever the loop reaches first and that would be used as the SOL for both in-place and out_of_place.

tl;dr the chart would show either OOP against IP SOL or IP against OOP SOL.

Please let me know if I'm misunderstanding - if you can show the actual results from this config run that would be awesome!

jj10306 (changes requested):

src/cloudai/metrics.py
podkidyshev: no, these metrics are meant to be universal and only differentiate by dimensions provided by the workload's `TestDefinition.metric_observations`

src/cloudai/metrics.py
podkidyshev: this is a common separation. some examples of the metric definitions (1st arg key, 2nd label)

```python
SIZE_BYTES = DimensionDefinition("size_bytes", "Size", Annotated[int, Field(strict=True, ge=0)])
BATCH_SIZE = DimensionDefinition("batch_size", "Batch size", Annotated[int, Field(strict=True, gt=0)])
OPERATION = DimensionDefinition("operation", "Operation", Annotated[str, Field(strict=True, min_length=1)])
```

it's I bet possible to use only `label` and construct key by keeping only letters + lowercase + chars into underscores but not sure if it's really a clean solution

or do you have something else in mind?

src/cloudai/metrics.py
podkidyshev: good catch, added

src/cloudai/metrics.py
podkidyshev: good catch, added

src/cloudai/metrics.py
podkidyshev: addressed, good point

podkidyshev: I intentionally kept it simple since I'm not 100% sure about the metrics feature. We anyway need some foundation for this so I'm trying this small step, let's see how it goes

src/cloudai/metrics.py
podkidyshev: later it would deserve its own place in the global Registry object where we track workloads, reporters, reward functions, agents, etc. but for now I decided to keep it here as an experiment

src/cloudai/metrics.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @classmethod
+    def register_metrics(cls, *metrics: MetricDefinition):
+        for metric in metrics:
+            cls._metrics[metric.key] = metric
+
+    @classmethod
+    def register_dimensions(cls, *dimensions: DimensionDefinition):
+        for dimension in dimensions:
+            cls._dimensions[dimension.key] = dimension
```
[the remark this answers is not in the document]
podkidyshev: not worth it

tests/test_metrics.py
```diff
@@ -0,0 +1,75 @@
...
+        }
+    )
+    test = cloudai.metrics.parse_sol_spec({"bandwidth": [{"value": 120, "match": {"operation": "write"}}]})
+
+    merged = cloudai.metrics.merge_sol_configs(None, system, test)
+
+    assert merged == {
+        "bandwidth": test["bandwidth"],
+        "latency": system["latency"],
+    }
```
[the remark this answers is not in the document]
podkidyshev: end-to-end nccl/nixl tests cover it

src/cloudai/metrics.py
podkidyshev: one day I'll add codecov finally...

src/cloudai/metrics.py
jj10306: > or do you have something else in mind?

nope, I think keeping them separate makes sense for clarity!

jj10306 (approved):
lgtm! please clarify https://github.com/NVIDIA/cloudai/pull/991/changes/BASE..013ebc6d9698875d669b19becbfd7976d414186a#r3753277571 before merging

Issue #985, opened 2026-07-29
Support relative-path test references and lazy-load only referenced test/hook tomls


shreyaskommuri:
### Problem

`Parser.parse()` in `src/cloudai/parser.py` globs and fully parses every `.toml` in `--tests-dir` and every hook toml, even when a scenario only references a few of them. Filtering to the tests actually used happens after everything is already loaded and validated.

This is hard to fix in isolation. A scenario references a test by its internal `name` field via `test_name`, not by file path. The only way to know which file has which name is to open and parse it. There is no name to file index, so lazy loading has nothing to work from without a change to how tests are referenced.

### Proposal

Two small, sequential PRs.

**PR 1: relative path test references.** Add an optional path based reference on `TestRunModel`, alongside the existing `test_name` lookup. Fully backward compatible. Existing scenarios keep working unchanged. A `[[Tests]]` entry could point directly at a toml relative to the scenario file instead of relying on directory wide name resolution.

**PR 2: lazy load referenced tomls.** Depends on PR 1. When every `[[Tests]]` entry in a scenario uses a path reference, skip the eager glob and parse of the whole `--tests-dir` and hooks dir, and load only the files actually referenced. If a scenario mixes path and name references, fall back to today's eager behavior for the name based ones. Full mixed mode laziness can be a later follow up.

### Open questions

What should a relative path resolve against. The scenario file's own directory seems most intuitive, flagging it rather than assuming.

Today, duplicate `name` values are rejected across the whole `--tests-dir`. Once loading is scoped to only the referenced files, does that check still need to hold directory wide, or just among the files actually loaded for a run.

### Out of scope

Not touching the `X | list[X]` sweeping annotation issue or the wider parsing refactor. Those need their own scoping discussion.

shreyaskommuri:
Thanks for merging #986. Wanted to put a concrete sketch here for PR 1 so there is something specific to react to rather than just prose.

Today, a `[[Tests]]` entry finds a test by name lookup:

```toml
[[Tests]]
id = "Tests.main"
test_name = "my_test"
```

Proposed addition, a new optional `path` field on `TestRunModel`, resolved relative to the scenario file's own directory:

```toml
[[Tests]]
id = "Tests.main"
path = "../tests/my_test.toml"
```

[...]

PR #994, opened 2026-08-10
Support path-based test references in scenario files

shreyaskommuri:
## Summary
- Adds an optional `path` field on `TestRunModel`, an alternative to `test_name` for referencing a test in a scenario file, resolved relative to the scenario file's own directory (confirmed shape in #985).
- `test_name` and `path` are mutually exclusive, same rules `test_name` already has with `test_template_name`.
- `TestScenarioParser._prepare_tdef` gets a new branch: when `path` is set, it loads that toml file directly instead of doing a name lookup against `test_mapping`, then merges scenario-level overrides the same way the `test_name` branch already does.
- Raises a clear `TestScenarioParsingError` if the resolved path does not exist.
- This is PR1 of the two-PR plan discussed in #985. PR2 (skip eager glob-and-parse of `--tests-dir`/hooks when a scenario's tests are all path-referenced) depends on this one and is not included here, keeping this PR small and reviewable on its own.
- Fully backward compatible: `test_name` and fully-inline (`test_template_name` + `name` + `description`) scenarios are unaffected.

## Test Plan
7 new tests in `tests/test_test_scenario.py` (`TestPathReference`), covering: `path`/`test_name` mutual exclusion, `path`/`test_template_name` mutual exclusion, relative-path resolution against the scenario file's directory, scenario-level override merging over the referenced file, a missing-file error, and a full scenario TOML parsed end to end with a `path` reference.

```
$ uv run pytest tests/test_test_scenario.py -q -k "TestPathReference"
.......
7 passed, 74 deselected in 0.04s

$ uv run pytest -q -m "not ci_only"
1848 passed, 5 skipped, 511 deselected in 4.47s

$ uv run pre-commit run --all-files
check for added large files..............................................Passed
check python ast.........................................................Passed
check for merge conflicts.................................................Passed
check toml................................................................Passed
check yaml.................................................................Passed
debug statements (python)..................................................Passed
fix end of files...........................................................Passed
mixed line ending..........................................................Passed
trim trailing whitespace...................................................Passed
pyright.....................................................................Passed
ruff check...................................................................Passed
ruff format...................................................................Passed
vulture.......................................................................Passed
import-linter..................................................................Passed
taplo...........................................................................Passed
```

Also updated the wording of two pre-existing validation error messages (in `models/scenario.py`) that referenced only `test_name`, so they stay accurate now that `path` is a second way to satisfy the same requirement. Updated the two existing tests in `tests/test_test_scenario.py` that asserted on the old wording.

## Additional Notes
Design confirmed in #985 before writing any code, per `CONTRIBUTING.md`'s "communicate with the main developers before starting work." Not touching the eager-loading behavior itself, sweeping, or anything unrelated, this PR is purely additive.

### DIFF
--- a/src/cloudai/models/scenario.py
+++ b/src/cloudai/models/scenario.py
@@ -155,7 +155,7 @@ class TestRunModel(BaseModel):
                     f"Possible values are: {', '.join(registry.test_definitions_map.keys())}"
                 )
         else:
-            if self.test_template_name:
+            if self.test_template_name is not None:
                 raise ValueError("'test_template_name' must not be set if 'test_name' or 'path' is set.")
 
         return self
--- a/src/cloudai/models/scenario.py
+++ b/src/cloudai/models/scenario.py
@@ -154,9 +154,8 @@ class TestRunModel(BaseModel):
                     f"Test type '{self.test_template_name}' not found in the test definitions. "
                     f"Possible values are: {', '.join(registry.test_definitions_map.keys())}"
                 )
-        else:
-            if self.test_template_name is not None:
-                raise ValueError("'test_template_name' must not be set if 'test_name' or 'path' is set.")
+        elif self.test_template_name is not None:
+            raise ValueError("'test_template_name' must not be set if 'test_name' or 'path' is set.")
 
         return self
 
--- a/src/cloudai/models/scenario.py
+++ b/src/cloudai/models/scenario.py
@@ -71,6 +71,13 @@ class TestRunModel(BaseModel):
 
     id: str = Field(min_length=1)
     test_name: Optional[str] = None
+    path: Optional[str] = Field(
+        default=None,
+        description=(
+            "Path to a test TOML file, resolved relative to this scenario file's own directory. "
+            "Alternative to 'test_name': references a test by file location instead of by name."
+        ),
+    )
     num_nodes: int | list[int] | None = None
     nodes: list[str] = Field(default_factory=list)
     exclude_nodes: list[str] = Field(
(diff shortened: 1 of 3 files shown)


src/cloudai/models/scenario.py
```diff
@@ -71,6 +71,13 @@ class TestRunModel(BaseModel):
 
     id: str = Field(min_length=1)
     test_name: Optional[str] = None
+    path: Optional[str] = Field(
+        default=None,
+        description=(
+            "Path to a test TOML file, resolved relative to this scenario file's own directory. "
+            "Alternative to 'test_name': references a test by file location instead of by name."
+        ),
+    )
```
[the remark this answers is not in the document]
podkidyshev: @shreyaskommuri please take a look at this one, if it makes sense to be resolved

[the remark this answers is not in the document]
shreyaskommuri: Yeah, made sense, good catch. Pushed a fix: `Field(min_length=1)` on `path` rejects an empty string at validation time instead of letting it silently resolve to the scenario file's own directory. Added a regression test too.

shreyaskommuri:
@podkidyshev the empty-path issue from CodeRabbit is resolved, min_length=1 pushed and confirmed rejecting "" while still allowing None/real paths (see reply on the review thread above), plus a regression test. All CI green on the latest commit. Let me know if there's anything else you'd like changed.

src/cloudai/test_scenario_parser.py
```diff
@@ -228,12 +228,25 @@ def _prepare_tdef(self, test_info: TestRunModel) -> TestDefinition:
             tc_defined = test_info.tdef_model_dump(by_alias=True)
             merged_data = deep_merge(test_defined, tc_defined)
             test = tp.load_test_definition(merged_data)
+        elif test_info.path:  # test referenced by file path, relative to this scenario file's directory
```
podkidyshev: please remove comment

src/cloudai/models/scenario.py
```diff
@@ -71,6 +71,14 @@ class TestRunModel(BaseModel):
 
     id: str = Field(min_length=1)
     test_name: Optional[str] = None
+    path: Optional[str] = Field(
```
podkidyshev, line 74: when `path` does not exist I get no error message in the console, no stack trace, app just fails immediately (both run and dry-run)

podkidyshev (changes requested):

src/cloudai/models/scenario.py
podkidyshev: `src/cloudai/test_scenario_parser.py` -> line 234 - I think this is the place where one needs to manually log exception and only then raise

src/cloudai/test_scenario_parser.py
shreyaskommuri: Removed.

src/cloudai/models/scenario.py
shreyaskommuri: Good catch, confirmed: Parser.parse() catches TestScenarioParsingError and exits without printing the exception itself, it relies on the raiser having already called logging.error(). Every other raise of this exception in the file follows that pattern, this branch was the one exception. Fixed and verified with a real dry-run against a missing path, now prints a clear [ERROR] message instead of failing silently.

podkidyshev (approved):

PR #996, opened 2026-08-11
[NIXL] ASIO support

podkidyshev:
## Summary
- Add ASIO support into NIXL
- ETCD isn't run for case when it's not needed (`etcd_endpoints` is set to an empty string)
  - This is not a backwards compatibility violation but rather bug fix. It may result in minor decreases of bench execution time, but other metrics should be the same

## Test Plan
- Automated CI
- Manual runs with the new NIXL scenario

## Additional Notes
-

### DIFF
--- a/src/cloudai/workloads/common/nixl.py
+++ b/src/cloudai/workloads/common/nixl.py
@@ -151,8 +151,8 @@ class NIXLBaseTestDefinition(TestDefinition, Generic[NIXLCmdArgsT]):
         return True
 
     @property
-    def supports_asio(self) -> bool:
-        """Return whether this workload supports NIXLBench's ASIO runtime."""
+    def uses_asio(self) -> bool:
+        """Return whether this workload uses NIXLBench's ASIO runtime."""
         return False
 
     @property
@@ -269,7 +269,7 @@ class NIXLCmdGenBase(EtcdCmdGenMixin):
             env_vars["NIXL_ETCD_NAMESPACE"] = "/nixl/kvbench/$(uuidgen)"
             env_vars["NIXL_ETCD_ENDPOINTS"] = '"$SLURM_JOB_MASTER_NODE:2379"'
         env_vars["SLURM_JOB_MASTER_NODE"] = "$(scontrol show hostname $SLURM_JOB_NODELIST | head -n 1)"
-        if tdef.supports_asio:
+        if tdef.uses_asio:
             env_vars["NIXL_ASIO_ADDRESS"] = (
                 "$(getent ahostsv4 \"$SLURM_JOB_MASTER_NODE\" | awk 'NR == 1 {print $1; exit}')"
             )
--- a/src/cloudai/workloads/common/nixl.py
+++ b/src/cloudai/workloads/common/nixl.py
@@ -147,7 +147,7 @@ class NIXLBaseTestDefinition(TestDefinition, Generic[NIXLCmdArgsT]):
         return self._etcd_image
 
     @property
-    def uses_managed_etcd(self) -> bool:
+    def uses_etcd(self) -> bool:
         """Return whether this workload needs CloudAI to manage an ETCD server."""
         return True
 
@@ -159,7 +159,7 @@ class NIXLBaseTestDefinition(TestDefinition, Generic[NIXLCmdArgsT]):
     @property
     def installables(self) -> list[Installable]:
         installables = [self.docker_image, *self.git_repos]
-        if self.uses_managed_etcd and self.etcd_image:
+        if self.uses_etcd and self.etcd_image:
             installables.append(self.etcd_image)
         return installables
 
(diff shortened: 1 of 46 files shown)


tests/workloads/nixl_bench/test_command_gen_strategy_slurm.py
```diff
@@ -179,31 +197,6 @@ def test_device_list_validation(
             assert cmd_args.total_buffer_size == expected_total_buffer_size
 
 
-def test_gen_etcd_srun_command(nixl_bench_tr: TestRun, slurm_system: SlurmSystem):
```
podkidyshev, line 182: a few tests have been removed as they duplicate the same code tested by the central e2e test

podkidyshev:
@rutayan-nv could you please take a look? (it's not urgent)

src/cloudai/workloads/common/nixl.py
```diff
@@ -254,9 +265,15 @@ def _cleanup_targets(self) -> list[Path]:
     @property
     def final_env_vars(self) -> dict[str, str | list[str]]:
         env_vars = super().final_env_vars
-        env_vars["NIXL_ETCD_NAMESPACE"] = "/nixl/kvbench/$(uuidgen)"
-        env_vars["NIXL_ETCD_ENDPOINTS"] = '"$SLURM_JOB_MASTER_NODE:2379"'
+        tdef = cast(NIXLBaseTestDefinition[NIXLBaseCmdArgs], self.test_run.test)
+        if tdef.uses_etcd:
+            env_vars["NIXL_ETCD_NAMESPACE"] = "/nixl/kvbench/$(uuidgen)"
+            env_vars["NIXL_ETCD_ENDPOINTS"] = '"$SLURM_JOB_MASTER_NODE:2379"'
```
srivatsankrishnan, line 271: Are these ports always same? Do we need this to be configured as variables and passed here?

rutayan-nv (dismissed):
LGTM.

Two non-blocking notes:

1. ASIO process startup relies on a bare `sleep 4`, while the ETCD path gets a real readiness loop (`until curl ... /health`). If the first process isn't listening within 4s, the second fails. Was 4 measured, or is a readiness check worth adding here too?

2. The null-runtime mode (`etcd_endpoints = ""`) is documented as storage-only, but nothing enforces it. Combined with UCX it yields two processes with no `--etcd-endpoints` and no managed ETCD, so there's nothing to coordinate them. A validator would make that a config-time error.

src/cloudai/workloads/nixl_bench/nixl_bench.py
```diff
@@ -52,6 +68,17 @@ def cmd_args_dict(self) -> dict[str, str | list[str]]:
             },
             exclude_none=True,
         )
+        if self.cmd_args.runtime_type == "ETCD":
+            # ETCD is NIXLBench's default runtime, so keep existing commands concise.
```
srivatsankrishnan: Can remove these comments?

src/cloudai/workloads/nixl_bench/nixl_bench.py
```diff
@@ -52,6 +68,17 @@ def cmd_args_dict(self) -> dict[str, str | list[str]]:
...
+            cmd_args.pop("runtime_type")
+            cmd_args.pop("asio_address")
+            cmd_args.pop("asio_port")
+            if not self.cmd_args.etcd_endpoints:
+                cmd_args.pop("etcd_endpoints")
+        else:
+            # ASIO performs direct peer-to-peer coordination and does not use ETCD endpoints.
```
srivatsankrishnan: Same here.

src/cloudai/workloads/common/nixl.py
podkidyshev: yes, these are the standard ETCD client (2379) and peer (2380) ports. they were already fixed before this PR; this change only makes the existing endpoint export conditional for ASIO

podkidyshev:
> LGTM.
> 
> Two non-blocking notes:
> 
> 1. ASIO process startup relies on a bare `sleep 4`, while the ETCD path gets a real readiness loop (`until curl ... /health`). If the first process isn't listening within 4s, the second fails. Was 4 measured, or is a readiness check worth adding here too?
> 2. The null-runtime mode (`etcd_endpoints = ""`) is documented as storage-only, but nothing enforces it. Combined with UCX it yields two processes with no `--etcd-endpoints` and no managed ETCD, so there's nothing to coordinate them. A validator would make that a config-time error.

1. The 4-second delay matches NIXLBench’s upstream test launcher. I tried 15 seconds like ETCD, but that collided with ASIO’s receive timeout in the two-node case; 4 seconds passed both placements. ASIO has no health endpoint, and a TCP probe would consume its single expected peer connection, so I kept the NIXL's startup approach.

2. I'm not sure this is right to maintain a list of network and storage backends to validate against, those change. Let users fail! :D

src/cloudai/workloads/nixl_bench/nixl_bench.py
podkidyshev: resolved

src/cloudai/workloads/nixl_bench/nixl_bench.py
podkidyshev: resolved

rutayan-nv (approved):
LGTM.

PR #998, opened 2026-08-13
[Reporting] JUnit

podkidyshev:
## Summary
Added JUnit XML generation

## Test Plan
- Automated CI
- Manual runs

## Additional Notes
-

### DIFF
--- a/doc/reporting.rst
+++ b/doc/reporting.rst
@@ -58,6 +58,12 @@ Enabling or disabling a report needs to be done in the system configuration:
    [reports]
    per_test = { enable = false }
    status = { enable = true }
+   junit = { enable = true }
+
+The ``junit`` scenario reporter writes ``junit.xml`` in the scenario results directory. It emits one test case for
+every regular test iteration and every DSE step, including pass/fail status, failure details, scheduler duration when
+available, and the contents of ``stdout.txt`` and ``stderr.txt``. The artifact can be consumed directly by Jenkins,
+GitLab, GitHub Actions, and other CI systems that support JUnit XML.
 
 Speed-of-Light comparisons
 --------------------------
--- a/doc/reporting.rst
+++ b/doc/reporting.rst
@@ -60,10 +60,10 @@ Enabling or disabling a report needs to be done in the system configuration:
    status = { enable = true }
    junit = { enable = true }
 
-The ``junit`` scenario reporter writes ``junit.xml`` in the scenario results directory. It emits one test case for
-every regular test iteration and every DSE step, including pass/fail status, failure details, scheduler duration when
-available, and the contents of ``stdout.txt`` and ``stderr.txt``. The artifact can be consumed directly by Jenkins,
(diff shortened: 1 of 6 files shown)


doc/reporting.rst
```diff
@@ -58,6 +58,12 @@ Enabling or disabling a report needs to be done in the system configuration:
    [reports]
    per_test = { enable = false }
    status = { enable = true }
+   junit = { enable = true }
+
+The ``junit`` scenario reporter writes ``junit.xml`` in the scenario results directory. It emits one test case for
+every regular test iteration and every DSE step, including pass/fail status, failure details, scheduler duration when
+available, and the contents of ``stdout.txt`` and ``stderr.txt``. The artifact can be consumed directly by Jenkins,
+GitLab, GitHub Actions, and other CI systems that support JUnit XML.
```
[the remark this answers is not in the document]
podkidyshev: wording nitpick

src/cloudai/reporter.py
```diff
@@ -134,6 +135,90 @@ def print_summary(self) -> None:
...
+        try:
+            duration = toml.load(metadata_path).get("elapsed_time_sec")
+            return float(duration) if duration is not None else None
+        except (OSError, TypeError, ValueError, toml.TomlDecodeError) as exc:
+            logging.debug("Could not read execution duration from %s: %s", metadata_path, exc)
+            return None
+
+    @staticmethod
+    def _format_duration(duration: float) -> str:
+        return f"{duration:g}"
```
rutayan-nv: nitpick - `:g` switches to scientific notation past 6 significant digits, so a cumulative suite time of 1,000,000s renders as `time="1e+06"`. Fine for parsers using `float()`, but stricter decimal parsers may reject it. `:.3f` sidesteps it.

src/cloudai/reporter.py
```diff
@@ -134,6 +135,90 @@ def print_summary(self) -> None:
...
+
+    @classmethod
+    def _add_log(cls, testcase: ET.Element, tag: str, path: Path) -> None:
+        if not path.is_file():
+            return
+        try:
+            content = path.read_text(errors="replace")
```
rutayan-nv, line 203: Question rather than a request: this embeds all of `stdout.txt`/`stderr.txt` per test case, and for megatron/nemo runs that is the full training log (the same file the report generators parse for per-iteration metrics), so it can be very large. Since the report is enabled by default, have you tried it on a training scenario? Wondering whether a size cap or attaching logs only for failed cases is worth it, or if it is fine in practice.

src/cloudai/reporter.py
podkidyshev: good point, changed to `:.3f`

src/cloudai/reporter.py
podkidyshev: valid concern. stdout/stderr was explicitly requested thus capping/omitting is not the right way. in the end I just disabled junit by default as it was a need from one team only

rutayan-nv (approved):
lgtm.

PR #999, opened 2026-08-13
feat(dse): add pin_nodes_to_first_step option to lock DSE steps to sa…

saivishal1999:


srivatsankrishnan (commented):
We are basically only restricting it for DSE jobs. We can start with this to unblock ourselves. 

But I think even benchmarking or chaining jobs in test scenario can also benefit from this?
@podkidyshev @saivishal1999

podkidyshev:
@saivishal1999 please provide PR description:

* how does it change behavior, in what cases
* how configs will look like
* how did you test it

PR #1001, opened 2026-08-14
[LLM] Custom models

podkidyshev:
## Summary
Allow custom models for vLLM and SGLang

## Test Plan
- Automated CI
- Manual runs

## Additional Notes
-

### DIFF
--- a/doc/workloads/vllm.rst
+++ b/doc/workloads/vllm.rst
@@ -70,6 +70,30 @@ supported under ``[[Tests]]`` in a test scenario. Define them in a test definiti
 ``test_name`` when custom benchmark or semantic-evaluation arguments are needed.
 
 
+Local Models
+------------
+Set ``cmd_args.model_path`` to load an existing model directory instead of downloading ``cmd_args.model`` from
+Hugging Face. The path must be absolute and visible inside the container. Use ``extra_container_mounts`` to mount a
+host or shared-filesystem directory. ``model`` remains the name exposed by the server and used by benchmark clients.
+
+.. code-block:: toml
+   :caption: test.toml (local model)
+
+   name = "vllm_local_model"
+   description = "vLLM with a local model"
+   test_template_name = "vllm"
+   extra_container_mounts = [
+     "/lustre/models/custom:/models/custom:ro",
+   ]
+
+   [cmd_args]
+   docker_image_url = "vllm/vllm-openai:v0.14.0-cu130"
+   model = "custom-model"
+   model_path = "/models/custom"
+
+The mounted directory must contain a model, configuration, and tokenizer supported by vLLM.
+
+
 Semantic Validation
 -------------------
 To run GSM8K semantic validation after the serving benchmark, add ``semantic_eval_cmd_args``. CloudAI reports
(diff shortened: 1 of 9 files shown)


src/cloudai/workloads/common/llm_serving.py
```diff
@@ -190,7 +190,10 @@ def extra_installables(self) -> list[Installable]:
 
     @property
     def installables(self) -> list[Installable]:
-        return [*self.git_repos, self.docker_image, self.hf_model, *self.extra_installables]
+        installables: list[Installable] = [*self.git_repos, self.docker_image]
+        if not Path(self.cmd_args.model).is_absolute():
```
rutayan-nv, line 194: Nice-to-have: validate that an absolute `model` is covered by `extra_container_mounts`. A missing mount currently surfaces as a serve failure mid-job. `megatron_run._check_path_is_mounted()` does this, but matches the mount destination exactly — here the model usually sits under it (`/lustre/models:/models` + `model = "/models/custom"`), so it would need a prefix match.

rutayan-nv, line 194: Nice-to-have: `./models/custom`, `~/models/custom`, and `$MODELS/custom` are not absolute, so they fall through to the HF download and fail with a repo-id error. A check on the first character (`.`, `~`, `$`) would catch it — cannot key on `/`, since repo ids contain one.

rutayan-nv (approved):
I did agent based review since I am not very familiar with the code. The suggestions seemed reasonable but not blockers for this PR.

doc/workloads/vllm.rst
```diff
@@ -70,6 +70,28 @@ supported under ``[[Tests]]`` in a test scenario. Define them in a test definiti
...
+   name = "vllm_local_model"
+   description = "vLLM with a local model"
+   test_template_name = "vllm"
+   extra_container_mounts = [
+     "/lustre/models/custom:/models/custom:ro",
+   ]
+
+   [cmd_args]
+   docker_image_url = "vllm/vllm-openai:v0.14.0-cu130"
+   model = "/models/custom"
```
srivatsankrishnan, line 90: What exactly is the ask here? How do we validate or check if the model is functionally correct? Or we own no responsibility and use judgement that user is doing the right thing?

podkidyshev: 1.

> What exactly is the ask here?

> Imagine I have a model already somewhere in /lustre/fsw/network_software_cloudai/dariab/my_custom_beautiful_model/
I can mount it to the container, but I cannot make neither vLLM nor SGLang to run over it. That's the functionality I'm requesting - the ability to run over some local model, not downloaded/precached from huggingface.

2. 

> Or we own no responsibility and use judgement that user is doing the right thing?

this. workload will fail early if something is wrong

src/cloudai/workloads/common/llm_serving.py
podkidyshev: this is somewhat fragile. on some clusters certain folders could be mounted by default, I don't want to assume `extra_contrainer_mounts` + `model` as path

podkidyshev: both your suggestions (with validate extra_continer_mounts) conflict each other - if a user really wants to go for not-abs model path, then checking for prefix does not work anymore

* dot- or tilda-based path are unreliable, I doubt we should explicitly support them
* env variable could work, but they could also specify model name as env variable liek `model = $MODEL_NAME` and have `[extra_env_vars] MODEL_NAME = "Qwen3"`

so I feel that this extensive path validation is also fragile

PR #1002, opened 2026-08-15
[Slurm, Reporting] pin-nodes + proper nodes storage

podkidyshev:
## Summary
1. `pin_nodes` option to keep re-using the same nodes within a given test case (sweeping or iterations)
2. store used nodes per test run
3. training report: use correct nodes count + nodes used

## Test Plan
- Automated CI
- Manual runs:
  - `pin_nodes` with iterations + DSE
  - negative `pin_nodes` test when swept against num_nodes
  - MBridge run to check number of nodes used in the training report

## Additional Notes
N/A

### DIFF
--- a/src/cloudai/models/scenario.py
+++ b/src/cloudai/models/scenario.py
@@ -74,6 +74,7 @@ class TestRunModel(BaseModel):
     test_name: Optional[str] = None
     num_nodes: int | list[int] | None = None
     nodes: list[str] = Field(default_factory=list)
+    pin_nodes: bool = False
     exclude_nodes: list[str] = Field(
         default_factory=list,
         description=(
@@ -111,6 +112,12 @@ class TestRunModel(BaseModel):
         if isinstance(value, dict):
             return cloudai.metrics.parse_sol_spec(value)
         return value
+
+    @model_validator(mode="after")
+    def validate_pin_nodes(self) -> Self:
+        if self.pin_nodes and isinstance(self.num_nodes, list):
+            raise ValueError("pin_nodes cannot be enabled when num_nodes is swept")
+        return self
 
     def tdef_model_dump(self, by_alias: bool) -> dict:
         """Return a dictionary with non-None values that correspond to the test definition fields."""
--- a/src/cloudai/systems/slurm/slurm_metadata.py
+++ b/src/cloudai/systems/slurm/slurm_metadata.py
@@ -81,7 +81,6 @@ class SlurmJobMetadata(_SlurmStepMetadataBase):
     job_root: Path
     job_steps: list[SlurmStepMetadata]
     nodes: list[str] = Field(default_factory=list)
-    num_nodes: int | None = None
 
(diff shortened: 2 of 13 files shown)


src/cloudai/systems/slurm/slurm_metadata.py
```diff
@@ -80,6 +80,8 @@ class SlurmJobMetadata(_SlurmStepMetadataBase):
     is_single_sbatch: bool = False
     job_root: Path
     job_steps: list[SlurmStepMetadata]
+    nodes: list[str] = Field(default_factory=list)
+    num_nodes: int | None = None
```
rutayan-nv: `num_nodes` is always `len(nodes)` at both call sites (`slurm_runner.py`, `single_sbatch_runner.py`), and the training parser recomputes `len(metadata.nodes)` rather than reading it. Suggest dropping the field before it gets baked into `slurm-job.toml`.

rutayan-nv (dismissed):
LGTM.

Non-blocking question: `pinned_nodes` is never cleared or refreshed, so if one of the pinned nodes drains mid-scenario, later iterations get `--nodelist` with an unavailable node and sit pending. Is that a realistic risk on your clusters, or is restarting the scenario the expected recovery?

src/cloudai/models/scenario.py
```diff
@@ -112,6 +113,12 @@ def parse_sol(cls, value: Any) -> float | cloudai.metrics.MetricSOLConfig | None
             return cloudai.metrics.parse_sol_spec(value)
         return value
 
+    @model_validator(mode="after")
+    def validate_pin_nodes(self) -> Self:
```
srivatsankrishnan, line 117: Is this a valid case? What benchmark currently sweeps nodes? I know technically it can be done, but we have any usecase yet where we sweep the nodes?

srivatsankrishnan (dismissed):
LGTM. For other comments,you can resolve it since not blocking this PR.

src/cloudai/models/scenario.py
podkidyshev: sweeping over num-nodes is explicitly supported thus it makes sense to cover this case

I think grid search over different node combinations is a valid case when users want to, e.g., verify that increasing num nodes increases performance of their workload

podkidyshev:
> LGTM.
> 
> Non-blocking question: `pinned_nodes` is never cleared or refreshed, so if one of the pinned nodes drains mid-scenario, later iterations get `--nodelist` with an unavailable node and sit pending. Is that a realistic risk on your clusters, or is restarting the scenario the expected recovery?

I believe CloudAI cannot recover from this situation and shouldn't as changing nodes would violate determinism contract

src/cloudai/systems/slurm/slurm_metadata.py
podkidyshev: resolved

rutayan-nv (approved):
lgtm.

PR #1003, opened 2026-08-19
fix(megatron_bridge): forward static extra_env_vars via -E

dshirron3:
Register TOML extra_env_vars in NeMo-Run container_env so consumers like nsys GPU_METRICS_NODES see them at container start. Keep -cb only for values that need job-shell expansion.

## Summary
- Forward Megatron-Bridge TOML extra_env_vars via -E into NeMo-Run container_env
  (needed for nsys GPU_METRICS_NODES and similar early consumers).
- Keep -cb only for job-shell expansion ($SLURM_*, arithmetic, ${PYTHONPATH}, etc.).
- Avoid -ce because comma-splitting breaks values like 0,1,2.

## Test Plan
- [x] Unit tests for -E / -cb / container-runtime env forwarding
- [ ] Profiled run with GPU_METRICS_NODES in extra_env_vars (no MBridge patch)
- [ ] Confirm PYTHONPATH=${PYTHONPATH} patterns still expand via -cb

## Additional Notes
My original problem was populating GPU_METRICS_NODES but i preffred a more wholistic solution that just handling GPU_METRICS_NODES

### DIFF
--- a/doc/workloads/megatron_bridge.rst
+++ b/doc/workloads/megatron_bridge.rst
@@ -84,6 +84,27 @@ Test-in-Scenario example:
      domain = "llm"
      compute_dtype = "fp8_mx"
 
+Executor Environment Variables
+------------------------------
+
+By default, CloudAI forwards ``extra_env_vars`` as ``-cb export KEY=VALUE``
+commands. This preserves compatibility with Megatron-Bridge releases such as
+v0.3.0 that do not support ``-E/--env`` and allows values to reference job-shell
+variables such as ``$SLURM_PROCID``.
+
+For a newer Megatron-Bridge version, variables that must be registered in the
+NeMo-Run executor's ``env_vars`` and ``container_env`` can be explicitly set
+under ``cmd_args.executor_env_vars``:
+
+.. code-block:: toml
+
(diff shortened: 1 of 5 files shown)


podkidyshev:
Hi Dan! Thanks for the contribution

the `-E` option didn't exist in MBridge 0.3.0 which we claim to support. I'm clarifying right now if we could bump minimum version supported to 0.4.0 (which includes -E support)

podkidyshev:
I'm afraid we need to keep supporting 0.3.0 where `-E` does not exist

as an alternative design, I'd suggest adding `cmd_args.executor_env_vars` into MegatronBridgeCmdArgs and rendering it as `-E` args. Would this work for you?

dshirron3:
> I'm afraid we need to keep supporting 0.3.0 where `-E` does not exist
> 
> as an alternative design, I'd suggest adding `cmd_args.executor_env_vars` into MegatronBridgeCmdArgs and rendering it as `-E` args. Would this work for you?

Yes, that a good idea. Is this something you can add in near time?

dshirron3:
@podkidyshev , I will update this PR according to your suggestion.

podkidyshev (approved):

PR #1009, opened 2026-08-21
[Installables] configurable workloads dependencies

podkidyshev:
## Summary
Recent [issue](https://github.com/NVIDIA/cloudai/pull/1004) proved that we must keep everything configurable for users. This PR aims to address other potentially problematic parts of CloudAI, making sure users can choose versions of components used by workloads.

## Test Plan
- Automated CI
- Manual runs of affected workloads
  - MBridge: 0.3.0 and 0.4.0
  - DynamoMocker
  - Other workloads were not tested as they don't alter default behavior anyhow

## Additional Notes
N/A

### DIFF
--- a/src/cloudai/workloads/aiconfig/aiconfigurator.py
+++ b/src/cloudai/workloads/aiconfig/aiconfigurator.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 from typing import List, Optional, Union
 
-from pydantic import BaseModel, ConfigDict, model_validator
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from cloudai.core import CmdArgs, Installable, PythonEnvironment, TestDefinition
 
@@ -61,6 +61,10 @@ class Disagg(BaseModel):
 class AiconfiguratorCmdArgs(CmdArgs):
     """Command arguments for Aiconfigurator workload with nested agg/disagg configs."""
 
+    requirements: str = Field(
+        default="aiconfigurator~=0.5.0",
+        description="Space-separated Python requirements installed into the Aiconfigurator environment.",
+    )
     model_name: str
     system: str
     backend: str = "trtllm"
@@ -91,7 +95,7 @@ class AiconfiguratorTestDefinition(TestDefinition):
         return PythonEnvironment(
             name="aiconfigurator",
             python_version="3.10",
-            requirements=["aiconfigurator~=0.5.0"],
+            requirements=self.cmd_args.requirements.split(),
         )
 
     @property
--- a/src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
+++ b/src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
@@ -90,6 +90,7 @@ class MegatronBridgeSlurmCommandGenStrategy(SlurmCommandGenStrategy):
             " ".join(parts),
             launcher_python,
             args.wandb_version,
+            args.numpy_version,
             pre_hook_sbatch_path=pre_hook_sbatch_path,
             base_slurm_params=base_slurm_params,
             capture_nodelist=capture_nodelist,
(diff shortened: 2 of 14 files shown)


src/cloudai/workloads/megatron_bridge/slurm_command_gen_strategy.py
```diff
@@ -388,9 +390,9 @@ def _wrap_launcher_for_job_id_and_quiet_output(
             *pre_hook_lines,
             ': >"$LOG"',
             "WANDB_INSTALL_RC=0",
-            f'{shlex.quote(launcher_python)} -m pip install wandb=={wandb_version} numpy==1.26.4 >>"$LOG" 2>&1 || WANDB_INSTALL_RC=$?',  # noqa: E501
+            f'{shlex.quote(launcher_python)} -m pip install wandb=={wandb_version} numpy=={numpy_version} >>"$LOG" 2>&1 || WANDB_INSTALL_RC=$?',  # noqa: E501
             'if [ "${WANDB_INSTALL_RC}" -ne 0 ]; then',
-            f'  echo "Failed to install runtime deps (wandb=={wandb_version}, numpy==1.26.4) in launcher venv (exit ${{WANDB_INSTALL_RC}})." >&2',  # noqa: E501
+            f'  echo "Failed to install runtime deps (wandb=={wandb_version}, numpy=={numpy_version}) in launcher venv (exit ${{WANDB_INSTALL_RC}})." >&2',  # noqa: E501
```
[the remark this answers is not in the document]
podkidyshev: we trust user inputs. CloudAI operates in the same environment as the user + CloudAI doesn't host any register of configs

src/cloudai/workloads/aiconfig/aiconfigurator.py
```diff
@@ -61,6 +61,10 @@ class Disagg(BaseModel):
 class AiconfiguratorCmdArgs(CmdArgs):
     """Command arguments for Aiconfigurator workload with nested agg/disagg configs."""
 
+    requirements: str = Field(
+        default="aiconfigurator~=0.5.0",
+        description="Space-separated Python requirements installed into the Aiconfigurator environment.",
+    )
```
jj10306, line 67: any reason to use a `str` here instead of `List[str]`? There are two potential issues with `str`:
1. afaik, `"aiconfigurator >= 0.5.0"` is valid (despite spaces around the operator) - if this is the case, this would break this logic
2. an empty string would fail at runtime because "" gets split into [] which would cause an empty venv to be created and lead to imports failing at runtime.

I believe using a List here would naturally be spported by TOML arrays and you can enforce the non-empty requirement via `min_length`

```suggestion
    requirements: List[str] = Field(
        default=["aiconfigurator~=0.5.0"],
        min_length=1,
        description="Space-separated Python requirements installed into the Aiconfigurator environment.",
    )
```

jj10306 (changes requested):

src/cloudai/workloads/aiconfig/aiconfigurator.py
podkidyshev: any list object in our TOMLs is by default a DSE sweep. to avoid that, one must also set `dse_excluded_args` param to include this one which I thought won't be the best UX

long story short, this `str` typing is a trade-off

jj10306: got it. in this case, should the str splitting logic be updated in order to address the 2 potential issues listed above?

jj10306 (changes requested):

src/cloudai/workloads/aiconfig/aiconfigurator.py
podkidyshev: 1. to address the first - the split logic would be rather complicated (iterating chunks manually, checking for characters, etc.) yet I don't see much profit
2. well, user's run will fail early anyway if that's the case. or maybe user's environment already has all the reqs installed so they actually don't need any reqs installed at all

in general I try to trust users and don't limit them with what they want to do. we have so many config options for so many workloads, ain't no way we verify everything....

jj10306 (approved):
lgtm!

PR #1018, opened 2026-09-02
Add Slurm multi-node and WideEP support for AI Dynamo

juntaowww:
## Summary
CloudAI previously treated num-nodes as independent backend workers and could not group multiple nodes into one distributed vLLM or SGLang worker.

This PR adds the optional `nodes-per-worker` setting, derives backend ranks and rendezvous arguments, supports multi-node TP/PP and WideEP layouts, infers disaggregation roles, and validates topology constraints. Omitting `nodes-per-worker` preserves the existing one-worker-per-node behavior.

It also updates AI Dynamo to 1.3.1, adds dedicated examples and documentation, preserves backend address overrides, and rejects AIPerf reports with zero successful requests.

## Test Plan
manual runs + tests.

## Additional Notes
- Slurm is supported; Kubernetes multi-node workers remain out of scope.
- Ray-backed multi-node vLLM workers are unsupported.


src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -418,7 +459,7 @@ class AIDynamoCmdArgs(CmdArgs):
     model_config = ConfigDict(extra="forbid")
 
     dynamo_version: str = Field(
-        default="f7e468c7e8ff0d1426db987564e60572167e8464",
+        default=DYNAMO_COMMIT,
```
podkidyshev, line 462: please don't change this one. we must support already working configs on users' side. or you're 100% bumped commit will remain backwards compatible?

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
```diff
@@ -341,13 +552,20 @@ _compute_worker_allocation_vllm() {
...
 
   if [[ ${prefill_config["gpus-per-worker"]} -eq 0 ]] || [[ ${decode_config["gpus-per-worker"]} -eq 0 ]]; then
     log "ERROR: Invalid TP/PP configuration"
     exit 1
   fi
+  if [[ ${prefill_config["gpus-per-worker"]} -gt $num_gpus ]] \
+    || [[ ${decode_config["gpus-per-worker"]} -gt $num_gpus ]]; then
+    log "ERROR: Worker GPU footprint exceeds the ${num_gpus} GPU(s) available on this node"
+    exit 1
+  fi
```
podkidyshev, line 568: I wonder if this isn't already a part of some validation in ai_dynamo.py or slurm_command_gen_strategy.py?

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
podkidyshev: looks like `deepep_path` should be configurable as well then

podkidyshev:
@juntaowww I need some time to properly review the changes, will try to finish it until end of the week

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
```diff
@@ -211,6 +395,15 @@ _parse_cli_pairs() {
   done
 }
 
+_set_worker_disaggregation_modes() {
```
podkidyshev, line 398: This introduces a backward-compatibility regression for existing vLLM configurations. Before this PR, the default prefill command was:

```bash
cmd = "python3 -m dynamo.vllm --is-prefill-worker"
```

This function now also adds `--disaggregation-mode prefill`, producing a command containing both role selectors. Dynamo exits before engine initialization with:

```
ValueError: Cannot combine --is-prefill-worker/--is-decode-worker with --disaggregation-mode. Use only --disaggregation-mode.
```

I reproduced this with the unchanged pre-PR configurations. All vLLM Slurm cases failed identically: single-node, shared-node, multinode, and LMCache. Equivalent SGLang cases passed because their existing disaggregation-mode value is stored in the argument map and gets overwritten rather than duplicated.

Updating the checked-in configuration to remove `--is-prefill-worker` migrates that configuration, but does not preserve compatibility for existing user configurations. The tests also replace the legacy command, so they do not cover this upgrade path.

Please preserve legacy role selectors: when cmd already contains `--is-prefill-worker` or `--is-decode-worker`, do not inject a conflicting `--disaggregation-mode`, or normalize the command safely. Please also add a regression test using the unchanged pre-PR vLLM configuration.

This failure occurs with `nodes-per-worker=1`, so it breaks existing workloads unrelated to the new multi-node-worker functionality.

conf/experimental/ai_dynamo/test/vllm.toml
```diff
@@ -21,7 +21,7 @@ extra_container_mounts = ["/run/udev:/run/udev"]
 dse_excluded_args = ["cmd_args.aiperf_phases"]
 
 [cmd_args]
-docker_image_url = "nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.1.1"
+docker_image_url = "nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.3.1"
```
podkidyshev, line 24: please keep the existing configs as is. only add new scenarios pointing to the existing test tomls, overriding specific values

podkidyshev (changes requested):

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
```diff
@@ -1259,14 +1585,14 @@ function render_lmcache_config()
...
-  local frontend_ip="$(_resolve_host_ip "$frontend_node")"
+  local frontend_address="$frontend_node"
   local storage_cache_dir="$(lmcache_storage_cache_dir)"
   mkdir -p "$storage_cache_dir"
   chmod 755 "$storage_cache_dir"
 
   local rendered_config="${LMCACHE_CONFIG_FILE}.tmp.${SLURM_NODEID:-0}"
   if ! FRONTEND_NODE="$frontend_node" \
-    FRONTEND_IP="$frontend_ip" \
+    FRONTEND_IP="$frontend_address" \
```
podkidyshev, line 1595: what is the justification for this change? now `FRONTEND_IP` is not an ip but hostname

src/cloudai/workloads/ai_dynamo/slurm_command_gen_strategy.py
```diff
@@ -59,6 +60,7 @@ def _container_mounts(self) -> list[str]:
     @property
     def final_env_vars(self) -> dict[str, str | list[str]]:
         env_vars = super().final_env_vars
+        env_vars["DYNAMO_NODELIST"] = "$(scontrol show hostname $SLURM_JOB_NODELIST | tr -s '\\n' ',' | sed 's/,$//')"
```
podkidyshev, line 63: could you please make this env var respects the value provided by user if any (and if empty - keep the definition you added)

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
juntaowww: reverted to make sure be compatible

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
juntaowww: indeed, removed as in python file the checks are already done

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
juntaowww: updated to ensure backward compatibility

conf/experimental/ai_dynamo/test/vllm.toml
juntaowww: reverted and the bump only appear on scenario files

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
juntaowww: reverted, as the changes were not necessary

src/cloudai/workloads/ai_dynamo/slurm_command_gen_strategy.py
juntaowww: added a conditional check and make sure it respect the users settings

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -708,6 +797,21 @@ def _parse_count_value(value: str | int | float | None) -> float | None:
         return None
 
 
+def parse_aiperf_request_count(report_path: Path) -> float | None:
```
podkidyshev: nit, but `float` for `request_count` looks confusing to me

src/cloudai/workloads/ai_dynamo/ai_dynamo.py
```diff
@@ -250,7 +290,7 @@ def validate_connector(cls, v: str | list[str] | None) -> str | list[str] | None
         worker_initialized_regex="VllmWorker.*has.been.initialized",
     )
     prefill_worker: WorkerConfig = WorkerConfig(
-        cmd="python3 -m dynamo.vllm --is-prefill-worker",
+        cmd="python3 -m dynamo.vllm",
```
podkidyshev: please restore this default

src/cloudai/workloads/ai_dynamo/slurm_command_gen_strategy.py
```diff
@@ -158,6 +164,24 @@ def _get_nested_toml_args(self, base_model: BaseModel, prefix: str, exclude: Lis
...
         return result
 
+    @staticmethod
+    def _uses_legacy_role_selector(worker: WorkerConfig) -> bool:
+        cmd_tokens = shlex.split(worker.cmd or "")
+        return any(token.partition("=")[0] in LEGACY_WORKER_ROLE_SELECTORS for token in cmd_tokens) or any(
+            worker.has_extra_arg(selector) for selector in LEGACY_WORKER_ROLE_SELECTORS
+        )
+
+    def _get_worker_script_args(self, role: str, worker: WorkerConfig, mode: str | None) -> List[str]:
```
podkidyshev, line 181: There's a problem with this function - it blocks previously possible (I know a team that uses this setup) agg placement (config below). It seems like this function discourages user-provided placement and forces inferred mode

<details>
  <summary>Agg config</summary>

```toml
# SPDX-FileCopyrightText: NVIDIA CORPORATION & AFFILIATES
# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0

# Regression experiment: preserve SGLang's explicit aggregate mode ("null").
# Run this identical scenario with the PR base and PR head, without overrides.
name = "dynamo-sglang-explicit-aggregate-compat"
job_status_check = false

[[Tests]]
id = "test.explicit-aggregate"
name = "sglang-explicit-aggregate"
description = "One SGLang aggregate worker with an explicitly configured backend mode."
test_template_name = "AIDynamo"
time_limit = "00:15:00"
extra_container_mounts = ["/run/udev:/run/udev"]

  [Tests.cmd_args]
  docker_image_url = "nvcr.io/nvidia/ai-dynamo/sglang-runtime:1.1.1"
  dynamo_version = "f7e468c7e8ff0d1426db987564e60572167e8464"
  workloads = "aiperf.sh"

    [Tests.cmd_args.dynamo]
    backend = "sglang"
    model = "Qwen/Qwen3-0.6B"
    endpoint = "v1/chat/completions"
    ingress-cmd = "python -m dynamo.frontend --router-mode round-robin"
    node-setup-cmd = "hostname"

      [Tests.cmd_args.dynamo.prefill_worker]
      num-nodes = 0

      [Tests.cmd_args.dynamo.decode_worker]
      num-nodes = 1
      cmd = "python3 -m dynamo.sglang"
      worker-initialized-regex = "register._register_model_with_runtime_config:.Successfully.registered.LLM.with.runtime.config"
      extra-args = "--skip-tokenizer-init --enable-metrics --disable-cuda-graph --disable-piecewise-cuda-graph"

        [Tests.cmd_args.dynamo.decode_worker.args]
        tensor-parallel-size = 1
        pipeline-parallel-size = 1
        disaggregation-mode = "null"
        mem-fraction-static = 0.5
        context-length = 4096
        host = "0.0.0.0"

    [Tests.cmd_args.aiperf]
    setup-cmd = "python -m pip install --break-system-packages --ignore-installed blinker==1.9.0 && python -m pip install --break-system-packages --upgrade aiperf==0.8.0"

      [Tests.cmd_args.aiperf.args]
      concurrency = 1
      request-count = 10
      synthetic-input-tokens-mean = 64
      output-tokens-mean = 32
      extra-inputs = '{"min_tokens":32,"max_tokens":32,"ignore_eos":true}'

  [Tests.extra_env_vars]
  DYNAMO_NODELIST = "$(scontrol show hostname $SLURM_JOB_NODELIST | tr -s '\\n' ',')"
  UCX_LOG_LEVEL = "warn"
```

</details>

src/cloudai/workloads/ai_dynamo/ai_dynamo.sh
```diff
@@ -109,31 +110,224 @@ _gpus_per_node() {
...
+    role_args["--nnodes"]="${role_config["nodes-per-worker"]}"
+    role_args["--node-rank"]="$node_rank"
+  fi
 }
 
 _apply_sglang_dsr1_section_args() {
   local self="$(_current_node_name)"
   local gpn="$(_gpus_per_node)"
-  local deepep_path="${dynamo_args["repo"]}/components/backends/sglang/configs/deepseek_r1/wideep/deepep.json"
+  local deepep_path="${dynamo_args["repo"]}/recipes/deepseek-r1/sglang/deepep.json"
```
podkidyshev: could we de-hardcode it (make configurable) and leave the previous path as the default one?

podkidyshev (changes requested):

PR #1023, opened 2026-09-08
[Slurm] Rest API

podkidyshev:
## Summary
- Add opt-in Slurm REST API support to `SlurmSystem` using the Slurm v0.0.38 API supported by Slurm 22.05.8.
- Route node discovery, GPU detection, sbatch submission, job status, completion, allocated-node lookup, and cancellation through `slurmrestd` when `[slurm_api]` is configured. Existing Slurm CLI behavior remains the default.
- Translate CloudAI-generated `#SBATCH` directives into the REST submission payload while preserving the original script body, authentication headers, retries, and certificate configuration.

## Test Plan

- Verified end to end on a virtual Slurm cluster using `slurmrestd` v0.0.38: submitted and completed workloads, observed cancellation, exercised required/excluded node placement, and verified compute-node execution through generated artifacts and job metadata.

## Additional Notes
- REST mode requires a reachable `slurmrestd` service exposing the `slurm` v0.0.38 endpoints.
- Workloads must produce an sbatch script. Workloads that invoke a separate Slurm launcher which submits its own jobs remain unsupported in REST mode.


podkidyshev:
@coderabbitai full review

src/cloudai/systems/slurm/slurm_node.py
```diff
@@ -14,11 +14,49 @@
...
+    nodes = []
+    if not node_list:
+        return []
+
+    components = re.split(r",\s*(?![^[]*\])", node_list)
+    for component in components:
+        if "[" not in component:
+            nodes.append(component)
+        else:
+            header, node_number = component.split("[")
```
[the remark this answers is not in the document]
podkidyshev: Out of scope: this parser was moved unchanged; multi-range hostlist support belongs in a separate PR.

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -158,6 +129,25 @@ def _reject_blank_transient_patterns(cls, value: list[str]) -> list[str]:
...
+        if self.slurm_api is None:
+            raise RuntimeError("Slurm REST API is not configured.")
+        return SlurmRestClient(self.slurm_api, self.status_retry_pause_seconds)
+
+    @staticmethod
+    def _job_id(job: BaseJob) -> int:
+        """Return numeric Slurm ID; e.g. `SlurmJob(..., id=42)` returns `42`."""
+        if not isinstance(job.id, int):
+            raise TypeError(f"Slurm job ID must be an integer, got {type(job.id).__name__}.")
+        return job.id
```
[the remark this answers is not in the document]
podkidyshev: Rejected: CloudAI creates integer Slurm IDs, and existing CLI cancellation already required an integer.

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+class SlurmAPIConfig(pydantic.BaseModel):
+    """Connection details for a Slurm REST API endpoint."""
+
+    model_config = pydantic.ConfigDict(extra="forbid")
+
+    url: str
+    headers: dict[str, str] = pydantic.Field(default_factory=dict)
+    verify_certs: bool = True
```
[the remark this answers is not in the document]
podkidyshev: Rejected: transport security is deployment policy; HTTP remains valid for trusted local or TLS-terminated endpoints.

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,437 @@
...
+            stripped = line.strip()
+            if not stripped or (stripped.startswith("#") and not stripped.startswith("#SBATCH")):
+                continue
+            if not stripped.startswith("#SBATCH"):
+                break
+            for option, value in self._parse_sbatch_line(stripped.removeprefix("#SBATCH").strip()):
+                self._apply_sbatch_directive(job, option, value)
+
+        job.setdefault("current_working_directory", str(script_path.parent.absolute()))
+        job["environment"] = {"PATH": os.environ.get("PATH", "/usr/local/bin:/usr/bin:/bin")}
```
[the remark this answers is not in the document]
podkidyshev: Rejected: Slurm 22.05.8 v0.0.38 explicitly requires environment to be a dictionary.

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,437 @@
...
+        "--error": "standard_error",
+        "-e": "standard_error",
+        "--partition": "partition",
+        "-p": "partition",
+        "--account": "account",
+        "-A": "account",
+        "--reservation": "reservation",
+        "--distribution": "distribution",
+        "--nodelist": "nodelist",
+        "--exclude": "exclude_nodes",
```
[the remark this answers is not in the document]
podkidyshev: Fixed in ef1b1aff: node constraints now use canonical required_nodes and excluded_nodes fields.

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -158,6 +129,25 @@ def _reject_blank_transient_patterns(cls, value: list[str]) -> list[str]:
...
+    def uses_slurm_api(self) -> bool:
+        """Whether Slurm communication uses slurmrestd instead of local CLI tools."""
+        return self.slurm_api is not None
+
+    @property
+    def _rest_client(self) -> SlurmRestClient:
+        """Build a client from REST config, or fail if REST mode is disabled."""
+        if self.slurm_api is None:
+            raise RuntimeError("Slurm REST API is not configured.")
+        return SlurmRestClient(self.slurm_api, self.status_retry_pause_seconds)
```
jj10306, line 142: High level design thought / suggestion here:

With these changes it appears the `SlurmSystem` object has to check in every relevant method whether or not CLI or API is used. Could it make more sense to define a `SlurmTransport` interface/ABC and have the REST API and CLI objects implement this? See below:

```
class SlurmTransport(ABC):
    def get_nodes(self) -> list[SlurmNode]: ...
    def get_allocated_nodes(self) -> list[SlurmNode]: ...
    def submit_job(self, submission_command: str, test_name: str) -> int: ...
    # ... the other ~9 methods
```
and then `SlurmSystem` would have an instance of this object as an attribute that is instantiated in `model_post_init`:

```
    def model_post_init(self, __context: Any) -> None:
        self._transport = (
            SlurmRestClient(self.slurm_api, self.status_retry_pause_seconds)
            if self.slurm_api is not None
            else SlurmCliTransport(self.cmd_shell)
        )

```

The reason I suggest this is due to the following

 1. The current structure requires that future developers must remember to add this `if self.uses_slurm_api ...` check. If they don't it would silently fallback to the CLI code and thus crash in cases where the CLI isn't supported - the interface approach would remove the mental burden of having to remember this
 2. The interface approach enables ease of testing the CLI and REST transport logic separately
 3. Minor code duplication of `if self.uses_slurm_api ...` currently

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,437 @@
...
+            stripped = line.strip()
+            if not stripped or (stripped.startswith("#") and not stripped.startswith("#SBATCH")):
+                continue
+            if not stripped.startswith("#SBATCH"):
+                break
+            for option, value in self._parse_sbatch_line(stripped.removeprefix("#SBATCH").strip()):
+                self._apply_sbatch_directive(job, option, value)
+
+        job.setdefault("current_working_directory", str(script_path.parent.absolute()))
+        job["environment"] = {"PATH": os.environ.get("PATH", "/usr/local/bin:/usr/bin:/bin")}
```
saivishal1999: this will export the process's PATH, and any other env vars in the TOML are also picked up later. But will any session environment variables also get picked up/exported? i.e, say during invocations like HF_TOKEN=... cloudai run .. will the prepended env vars get picked up? Should we handle/accommodate for this case?

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -308,6 +336,13 @@ def submit_job(self, submission_command: str, test_name: str) -> int:
 
     def validate_install_environment(self) -> None:
         """Validate that the configured Slurm environment can run CloudAI workloads."""
+        if self.uses_slurm_api:
+            try:
+                self._rest_client.validate()
```
saivishal1999, line 352: _REQUIRED_BINARIES in line 346 has 'git' in its list of things to be checked. Do we have to make sure it gets checked inside this path as well?

src/cloudai/systems/slurm/slurm_system.py
podkidyshev: yes I considered this approach. I decided between a broader refactor like you suggested (or a separate System class that inherits/wraps SlurmSystem) VS uncertainty of the overall thing (whether it will live long or not; at the moment slurm rest api is needed just for one internal initiative)

eventually thought about going simpler for now. if there will be more development - the refactor will be straight-forward to do. for now I wanted a smaller diff

src/cloudai/systems/slurm/slurm_rest_client.py
podkidyshev: resolved, using the full env right now, very good catch!

src/cloudai/systems/slurm/slurm_system.py
podkidyshev: one more good catch! yes, git is needed (for local installations of repos)

added it above so non-slurm binaries (git only now) are checked for both slurm-cli and slurm-rest

src/cloudai/systems/slurm/slurm_system.py
```diff
@@ -315,7 +343,18 @@ def submit_job(self, submission_command: str, test_name: str, *, check_return_co
...
         """Validate that the configured Slurm environment can run CloudAI workloads."""
-        for binary in self._REQUIRED_BINARIES:
+        for binary in self._REQUIRED_LOCAL_BINARIES:
+            if shutil.which(binary) is None:
+                raise EnvironmentError(f"Required binary '{binary}' is not installed.")
+
+        if self.uses_slurm_api:
+            try:
+                self._rest_client.validate()
+            except RuntimeError as exc:
```
jj10306, line 353: is there a specific exception we can catch here that would indicate the REST API is not setup / available on the cluster vs. just a transient connection issue?

currently a transient connection issue here would cause the rest of the execution to believe the REST client isn't available even if it actually

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,439 @@
...
+            stop=tenacity.stop_after_attempt(retry_threshold),
+            wait=tenacity.wait_fixed(self._retry_pause_seconds),
+            retry=tenacity.retry_if_exception_type((requests.RequestException, ValueError, RuntimeError)),
+            before_sleep=tenacity.before_sleep_log(logger, logging.WARNING),
+            reraise=True,
+        )
+        try:
+            return retrying(self._request_once, method, service, path, payload)
+        except (requests.RequestException, ValueError, RuntimeError) as exc:
+            raise RuntimeError(f"Slurm API request failed after {retry_threshold} attempt(s): {exc}") from exc
```
jj10306, line 179: Why are we catching several unique exception types and repackaging them as the generic runtime error?

It seems like each of these unique exceptions communicate a different failure cause. So would be good for downstream callers to be able to See what actual exception was thrown so they can act accordingly.

Specifically, this is relevant to the question I asked about The validate REST API method, generally catching runtime exception and declaring that the REST API is not available regardless of the specific details of the error

Given that we are re-throwing a runtime error, why not just let these exceptions pass through as is?

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,439 @@
...
+        flags = {flag.upper() for flag in state_flags}
+
+        if "NOT_RESPONDING" in flags:
+            return SlurmNodeState.NOT_RESPONDING
+        if "DRAIN" in flags:
+            if state in {"ALLOCATED", "MIXED"} or "COMPLETING" in flags:
+                return SlurmNodeState.DRAINING
+            return SlurmNodeState.DRAINED
+        if "FAIL" in flags:
+            return SlurmNodeState.FAILING if state == "ALLOCATED" or "COMPLETING" in flags else SlurmNodeState.FAIL
```
jj10306, line 302: Is this logic specific to the REST client response parsing or is this the same logic that's used for the CLI? If it's the latter, should we We refactor this to a common method, such that this can be used across different transports?

src/cloudai/systems/slurm/slurm_rest_client.py
```diff
@@ -0,0 +1,439 @@
...
+            "DEADLINE",
+            "FAILED",
+            "NODE_FAIL",
+            "OUT_OF_MEMORY",
+            "PREEMPTED",
+            "REVOKED",
+            "SPECIAL_EXIT",
+            "TIMEOUT",
+        }
+    )
```
jj10306, line 72: nit: are these constants specific to the REST API? if not, should this live somewhere more general (i.e. not in the rest client)?

jj10306 (approved):
lgtm - nice work @podkidyshev 🙂

left a couple minor comments, feel free to address here or in a follow up

jj10306:
btw - @podkidyshev can you provide more detailed test instructions (i.e. what command, what files did you edit, etc) on how to use the REST API feature?

podkidyshev:
> btw - @podkidyshev can you provide more detailed test instructions (i.e. what command, what files did you edit, etc) on how to use the REST API feature?

please refer to the changes in doc/USER_GUIDE.rst

src/cloudai/systems/slurm/slurm_rest_client.py
podkidyshev: yes, it's only for REST. the base Slurm states are only `"COMPLETED", "FAILED", "CANCELLED", "TIMEOUT", "CANCELLED+"` and those are declared separately in the base slurm system class

src/cloudai/systems/slurm/slurm_system.py
podkidyshev: I'm not sure I get the question... I could just say, that REST API method is defined by configuration and it doesn't fallback to CLI. so if CloudAI cannot detect rest api available - it aborts immediately

src/cloudai/systems/slurm/slurm_rest_client.py
podkidyshev: CLI and REST are different here:

* CLI path receives a compact values such as `alloc`, `drng`, or `idle*` (search for `state_abbreviations` mapping) as a **single variable** `state`
* for REST we need **two variables**: `state` and `state_flags`

src/cloudai/systems/slurm/slurm_rest_client.py
podkidyshev: agreed. I double checked slurm system exceptions handling... I can't say there's a very consistent strategy for it, neither I can confirm this list of 3 particular exceptions is actually a complete one 

so I just propagated them clearly to the slurm system and it handles them there transforming into `JobIdRetrievalError` or just as a warning (when checking for gpu directives support)

src/cloudai/systems/slurm/slurm_system.py
jj10306: I was asking about catching `RuntimeError` here - is there a more specific exception we should catch.

podkidyshev: yes so now I'm catching a broader exception (well, just `Exception`)

PR #1030, opened 2026-09-11
[Reporting] Unified output for standalone scenarios

podkidyshev:
## Summary

- Add a structured `experiment.json` at the scenario results root, with Pydantic models for experiment metadata, tests, runs, status, timing, and metrics.
- Record standalone runs at submission and completion, preserving each iteration before its mutable state advances. Workload success comes from `was_run_successful()`; canonical metrics come from `metric_observations()`.
- Create output for every runner, including dry runs, and finalize it when execution finishes or raises an exception. Replace snapshots atomically; metric extraction and write failures warn without changing benchmark behavior.

## Test Plan

- Automated CI.
- Manual standalone execution on Linux: ran two AIConfigurator prediction cases with output sequence lengths of 150 and 300. Downloaded the artifacts and inspected them locally. Both cases produced prediction reports, and `experiment.json` contains two completed tests with one completed run each, UTC start/finish timestamps, and positive durations. 

## Additional Notes

This PR provides shared output infrastructure and standalone run capture. Slurm run capture, DSE result selection, and single-sbatch support are covered by subsequent PRs. `Test.metrics` remains empty in this PR; canonical measurements are retained per run.

### DIFF
--- a/doc/reporting.rst
+++ b/doc/reporting.rst
@@ -31,6 +31,21 @@ Per-test reports are linked to a particular workload type (e.g. ``NcclTest``). A
 To list all available reports, users can use ``cloudai list-reports``. Use verbose output to also print report configurations.
 
 
+Unified experiment output
+-------------------------
+
+Ordinary Slurm scenarios write ``experiment.json`` in the scenario results directory when execution finishes or fails.
+The file contains experiment metadata, test cases, submitted runs, statuses, timing, and canonical metrics from
+``TestDefinition.metric_observations()``. NCCL and NIXLBench provide these metrics; other workloads have empty metric lists.
+Test-level metrics are arithmetic means of successful iterations at matching metric and dimension points. Per-run
+measurements retain their original values, and missing measurements are not treated as zero.
+
+Experiment timing covers scenario execution, including gaps between jobs. Run timestamps without timezone information
+are null. Metric extraction and output-write failures produce warnings without changing execution behavior. The file is
+replaced atomically and is independent of reporter configuration. Dry runs, DSE, and single-sbatch execution do not produce
+this artifact.
+
+
 .. _general-flow:
 
 General Flow
--- a/src/cloudai/metrics.py
+++ b/src/cloudai/metrics.py
@@ -44,6 +44,8 @@ class DimensionDefinition:
     key: str
     label: str
     value_type: Any
+    unit: str = ""
+    is_x: bool = False
 
     def validate(self, value: Any) -> MetricValue:
         """Validate one configured or observed dimension value."""
@@ -97,7 +99,7 @@ class MetricCatalog:
         return {key: cls.get_dimension(key).validate(value) for key, value in values.items()}
 
 
-SIZE_BYTES = DimensionDefinition("size_bytes", "Size", Annotated[int, Field(strict=True, ge=0)])
+SIZE_BYTES = DimensionDefinition("size_bytes", "Size", Annotated[int, Field(strict=True, ge=0)], unit="B", is_x=True)
 BATCH_SIZE = DimensionDefinition("batch_size", "Batch size", Annotated[int, Field(strict=True, gt=0)])
 OPERATION = DimensionDefinition("operation", "Operation", Annotated[str, Field(strict=True, min_length=1)])
 PLACEMENT = DimensionDefinition("placement", "Placement", Literal["in_place", "out_of_place"])
--- a/src/cloudai/systems/slurm/slurm_runner.py
+++ b/src/cloudai/systems/slurm/slurm_runner.py
@@ -21,7 +21,8 @@ from typing import cast
 
 import toml
 
-from cloudai.core import BaseJob, BaseRunner, JobIdRetrievalError, System, TestRun, TestScenario
+from cloudai.core import BaseJob, BaseRunner, JobIdRetrievalError, JobStatusResult, System, TestRun, TestScenario
+from cloudai.unified_output import ExperimentOutput
 from cloudai.util import CommandShell
 
 from .slurm_command_gen_strategy import SlurmCommandGenStrategy
@@ -43,6 +44,34 @@ class SlurmRunner(BaseRunner):
         self.system = cast(SlurmSystem, system)
         self.cmd_shell = CommandShell()
         self.pinned_nodes: dict[str, list[str]] = {}
+        self._experiment: ExperimentOutput | None = None
+
+    def run(self) -> None:
+        if self.mode != "run" or any(tr.is_dse_job or tr.step > 0 for tr in self.test_scenario.test_runs):
+            super().run()
+            return
+
+        try:
+            self._experiment = ExperimentOutput(self.test_scenario, self.scenario_root)
+        except Exception as exc:
+            logging.warning("Cannot initialize unified experiment output: %s", exc)
+        completed = False
+        try:
+            super().run()
+            completed = True
+        finally:
+            if self._experiment is not None:
+                self._experiment.finish(self.system, self.jobs, completed)
+                self._experiment = None
+
+    def get_job_status(self, job: BaseJob) -> JobStatusResult:
+        result = super().get_job_status(job)
+        if self._experiment is not None:
+            try:
+                self._experiment.capture(self.system, job, result)
+            except Exception as exc:
+                logging.warning("Cannot capture unified output for job %s: %s", job.id, exc)
+        return result
 
     def submit_test(self, tr: TestRun) -> None:
         if tr.pin_nodes and tr.name in self.pinned_nodes:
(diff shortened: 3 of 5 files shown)


podkidyshev:
@coderabbitai full review

doc/reporting.rst
```diff
@@ -31,6 +31,21 @@ Per-test reports are linked to a particular workload type (e.g. ``NcclTest``). A
...
+``TestDefinition.metric_observations()``. NCCL and NIXLBench provide these metrics; other workloads have empty metric lists.
+Test-level metrics are arithmetic means of successful iterations at matching metric and dimension points. Per-run
+measurements retain their original values, and missing measurements are not treated as zero.
+
+Experiment timing covers scenario execution, including gaps between jobs. Run timestamps without timezone information
+are null. Metric extraction and output-write failures produce warnings without changing execution behavior. The file is
+replaced atomically and is independent of reporter configuration. Dry runs, DSE, and single-sbatch execution do not produce
+this artifact.
+
+
```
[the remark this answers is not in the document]
podkidyshev: resolved

src/cloudai/output.py
```diff
@@ -0,0 +1,123 @@
...
+        temporary_path: pathlib.Path | None = None
+        try:
+            content = self.snapshot().model_dump_json(indent=2)
+            self.output_path.mkdir(parents=True, exist_ok=True)
+            with tempfile.NamedTemporaryFile(
+                mode="w", encoding="utf-8", dir=self.output_path, prefix=".experiment.json.", delete=False
+            ) as temporary:
+                temporary_path = pathlib.Path(temporary.name)
+                temporary.write(content + "\n")
+            temporary_path.replace(self.output_path / "experiment.json")
```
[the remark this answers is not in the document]
podkidyshev: over-engineering for now

src/cloudai/_core/runner.py
```diff
@@ -79,13 +79,15 @@ def create_runner(self, mode: str, system: System, test_scenario: TestScenario)
 
         return runner_class(mode, system, test_scenario, results_root)
 
-    def run(self):
+    def run(self) -> bool:
         """Run the test scenario using the instantiated runner."""
         try:
             self.runner.run()
             logging.debug("All jobs finished successfully.")
+            return True
```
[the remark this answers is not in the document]
podkidyshev: fixed in subsequent PR

tests/test_acceptance.py
```diff
@@ -170,6 +187,37 @@ def test_details_is_dumped_and_valid(self, do_dry_run: tuple[Path, dict]) -> Non
...
+
+        assert experiment.model_dump() == {
+            "id": results_output.name,
+            "name": toml.load(scenario["path"])["name"],
+            "description": None,
+            "status": "completed",
+            "path": str(results_output.absolute()),
+            "start": experiment.start,
+            "finish": experiment.finish,
+            "duration": experiment.duration,
```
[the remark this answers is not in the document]
podkidyshev: not relevant for tests, let it be

rutayan-nv:
A few suggestions.

**`_update_timing`** — a naive datetime becomes `None` rather than assumed-UTC, so the timestamp and the derived duration drop silently:

```
before: start=2026-09-21 10:00:00  finish=2026-09-21 10:05:00
after : start=None  finish=None  duration=None
```

Latent today since every producer passes aware datetimes, but `Run.start`/`finish` accept naive ones and the Slurm PR adds producers. `value.replace(tzinfo=utc)` or a warning seems safer than a silent drop. Related: `finish()` runs this against the live experiment while `snapshot()` works on a copy.

**Dry run vs real run** — a dry run writes `status: "completed"` with every test completed and `runs: []`. Since `SlurmRunner` has no `get_run_output` yet, a real Slurm run produces the same shape. A `mode` field would make the artifact self-describing; `doc/reporting.rst` mentions dry runs but the JSON carries no marker.

**`finally: finish_output(successful)`** — the mixed DSE/non-DSE `return 1` marks the experiment `"failed"` though nothing ran. And `handle_non_dse_job` returns success now, but that branch still returns 0, so the CLI can exit 0 while `experiment.json` says failed. Intentional for compatibility?

podkidyshev:
> A few suggestions.
> 
> **`_update_timing`** — a naive datetime becomes `None` rather than assumed-UTC, so the timestamp and the derived duration drop silently:
> 
> ```
> before: start=2026-09-21 10:00:00  finish=2026-09-21 10:05:00
> after : start=None  finish=None  duration=None
> ```
> 
> Latent today since every producer passes aware datetimes, but `Run.start`/`finish` accept naive ones and the Slurm PR adds producers. `value.replace(tzinfo=utc)` or a warning seems safer than a silent drop. Related: `finish()` runs this against the live experiment while `snapshot()` works on a copy.
> 
> **Dry run vs real run** — a dry run writes `status: "completed"` with every test completed and `runs: []`. Since `SlurmRunner` has no `get_run_output` yet, a real Slurm run produces the same shape. A `mode` field would make the artifact self-describing; `doc/reporting.rst` mentions dry runs but the JSON carries no marker.
> 
> **`finally: finish_output(successful)`** — the mixed DSE/non-DSE `return 1` marks the experiment `"failed"` though nothing ran. And `handle_non_dse_job` returns success now, but that branch still returns 0, so the CLI can exit 0 while `experiment.json` says failed. Intentional for compatibility?

1. timezone issue - it's rather hypothetical. later PRs produce explicit TZ-aware datetime objects. In other words, there's no producer of these DT objects that can provide unaware DT
2. dry-run vs real run for Slurm - addressed in the next PRs in the stack. this PR focuses on overall feature and Standalone mode
3. mixing DSE/non-DSE is prohibited by design. It's an actuall error despite CLI returns 0. So `failed` is absolutely correct. CLI can be fixed but it's out of scope here. Related features will need some touch of handlers code - perhaps I'll address it there

podkidyshev:
@blugassi please take a look as well (notice that it's a stack, support for slurm is in the 2nd pr)

rutayan-nv (approved):

doc/reporting.rst
```diff
@@ -31,6 +31,22 @@ Per-test reports are linked to a particular workload type (e.g. ``NcclTest``). A
 To list all available reports, users can use ``cloudai list-reports``. Use verbose output to also print report configurations.
 
 
+Unified experiment output
```
jj10306, line 34: can we include an actual experiment.json file from a real run somewhere in the repo? I know that the tests show the structure with mock data, but would be helpful to have a real example to reference

src/cloudai/models/output.py
```diff
@@ -0,0 +1,89 @@
...
+    name: str
+    description: str | None = None
+    status: Status = "pending"
+    path: str
+    metrics: list[Metric] = pydantic.Field(default_factory=list)
+    runs: list[Run] = pydantic.Field(default_factory=list)
+    dse: DSE | None = None
+
+
+class Experiment(pydantic.BaseModel):
```
jj10306, line 78: for S3 publishing, we need the `system.name` (i.e. cluster) from the system toml in order to construct the s3 key. can we add this info here?

either can be a direct field of Experiment (i.e. `system_name: str`) or we can create something like `class SystemInfo` class that holds all the system level info (cluster, scheduler, etc) since there may be more fields we need to add in the future

src/cloudai/models/output.py
```diff
@@ -0,0 +1,89 @@
...
+    """Full snapshot spanning the entire scenario, including all DSE trials."""
+
+    id: str
+    name: str
```
jj10306, line 82: what's id vs name?

src/cloudai/systems/slurm/single_sbatch_runner.py
```diff
@@ -190,6 +190,7 @@ def run(self):
         tr = self.test_scenario.test_runs[0]
         job = self._submit_test(tr)
         self.jobs.append(job)
+        self.update_run_output(job)
```
jj10306, line 193: if this PR is for "standalone sceanarios", why are we modifying slurm files here?

src/cloudai/models/output.py
podkidyshev: I'd say `system.name` is unreliable as it's a user input. For Slurm and Standalone jobs that run on a cluster we could use slurm's cluster name conf option (already retrieved by cloudai)

I'll take a look on what I could do here, maybe will make it a separate PR

src/cloudai/models/output.py
podkidyshev: * name is TestScenario.name (e.g. `vllm-regression` or `nccl-dse-gb200-eos`
* id is a unique id of a single execution. At the moment I simply use the output folder name which is `<scenario.name>-<dt>`

src/cloudai/systems/slurm/single_sbatch_runner.py
podkidyshev: I worked on the whole stack at a time so some stubs are in-place. This PR focuses on standalone but other systems generate experiment outputs with at least some basic info anyway

PR #1035, opened 2026-09-17
feat(env_params): add a log encoding for ordinal env_params

rutayan-nv:
Adds a second `env_params` observation encoding. No behaviour change for existing configs — `CategoricalEncoding` stays the default.

## Why

`CategoricalEncoding` observes a drawn value as its **index** into the candidate list. That is right for an arbitrary set, and its docstring says so. It is wrong when the candidates are ordinal, because the index throws away magnitude.

`drop_rate ∈ {0.0, 0.001, 0.01}` is monotone severity. A policy told *"index 2"* cannot generalize from 0.001 to 0.01; a policy told *"-2 on a log scale"* can, and can extrapolate past the candidates it trained on.

This was cut from the original `env_params` series deliberately, for want of a use case. The use case has arrived: the MRC-PRT RL policy randomizes `drop_rate` and `msg_size` over decades and needs their magnitude to transfer across regimes.

## What

```toml
[env_params.drop_rate]
encoding = { type = "log" }
```

`LogEncoding` observes `[is_zero, log10(value)]` as `ObsLeafDescriptor(kind="box", dim=2)`.

Two dimensions because `log10(0)` is undefined **and** an exact zero is not merely a small value — "no drops at all" is a qualitatively different regime. The indicator carries that case and the log slot stays `0.0`, so a zero draw lands on the flag instead of becoming an extreme outlier that dominates the observation's scale.

`EnvParamSpec.encoding` becomes `Union[CategoricalEncoding, LogEncoding]` discriminated on `type`. The field was previously annotated as the one concrete class, which closed the extension point the `Encoding` protocol's own docstring describes ("a new strategy implements this pair without touching `EnvParam` or the adapter"). Discriminating on the name also means a typo is rejected with the valid options rather than silently falling back to the default.

## Secondary benefit, downstream

An RL connector stack that normalizes observations with a running mean/std filter is built for continuous leaves. Applied to a one-hot categorical it de-means the one-hot dimensions and leaves a category that has not yet been drawn with zero variance, so that dimension is divided by roughly the filter's epsilon and can dominate the input the first time it fires. With every leaf currently `Discrete`, that is the default path for any env declaring `env_params`. A continuous leaf removes it.

## Scope

Encoding only. `LogUniformSampling` — continuous sampling rather than a candidate list — is **not** part of this; the candidate list stays the single source of truth for values, and the use case is a discrete list.

## Gates

`ruff check`, `ruff format --check`, `pyright` (0 errors) all pass. `pytest`: 1952 passed, 5 skipped. Eight new tests in `tests/test_env_params.py` covering the descriptor width, magnitude vs position, the exact-zero indicator, negative input, candidate-list independence, discriminated selection by name, the unchanged categorical default, rejection of an unknown name, and the `EnvParam` delegation path.


rutayan-nv:
Closing recommendation from the author: **this has no consumer.**

I opened it to close a divergence between an experiments branch and `main`, without first checking whether anything that is actually being upstreamed consumes the encoding. It does not.

`[is_zero, log10]` appears in 26 files on that experiments branch — every `mrc_prt_stage3_*.toml` staging config, about 16 analysis and experiment scripts, and the branch's own `rl_ppo.py`. None of those are in the check-in scope. The configs being upstreamed declare the default categorical form, and the policy they feed sees a one-hot over three regimes, which is an ordinary contextual-bandit representation and is well-scaled after normalization (Bernoulli(1/3) per dimension → about +1.4 set, −0.7 unset).

The arguments I put in the original description do not hold up:

- **Generalizing across drop rates** only pays if behaviour is needed at values outside the candidate list. The regimes *are* the list.
- **The one-hot `MeanStdFilter` interaction** is a Ray connector concern. Changing observation semantics to work around it is the wrong layer.
- **Checkpoint portability across differing candidate lists** is real but applies to offline policy-inspection scripts that are not being upstreamed.

What is actually left is reproducing the conditions one historical measurement was taken under, which is a reason to reproduce that measurement, not a reason for core to grow a capability.

This was cut from the original `env_params` series for want of a use case. That judgement was right, and still is. The one thing that would revive it is a Stage-3 staging config being upstreamed with `encoding = { type = "log" }`; if that happens I will come back with the config as the motivating case.

The implementation and its eight tests are on the branch if it is ever wanted.

podkidyshev:
@rutayan-nv 

> Closing recommendation from the author: this has no consumer.

so are you going to push it for review/merge or no? if no pls mark it as draft or even close the PR so it doesn't bloat the repo 🙏

PR #1044, opened 2026-09-19
Make File.is_installed binary-safe

### LINKED ISSUE #782: Improve File installation check to handle binary files and content-mismatch errors
## Description
The current implementation of `is_installed_one` for `File` items in both Slurm and Kubernetes installers uses `read_text()` to compare file contents, which:
1. Raises `UnicodeDecodeError` for binary files
2. Returns misleading error messages when a file exists but contents differ (says 'does not exist')

## Proposed Enhancement
- Use `read_bytes()` instead of `read_text()` for binary-safe comparison
- Distinguish between 'file does not exist' and 'file exists but contents differ' in error messages

## References
- PR #781: 
- Comment: https://github.com/NVIDIA/cloudai/pull/781#discussion_r2708354205

## Affected Files
- `src/cloudai/systems/kubernetes/kubernetes_installer.py`
- `src/cloudai/systems/slurm/slurm_installer.py`

/cc @amaslenn

MaxFreedomPollard:
## Summary
Fixes #782.

- `File.is_installed` compared the installed copy with the source using `read_text()`, which raises `UnicodeDecodeError` for files that are not valid UTF-8. It now compares bytes, which matches `install()`, where the file is copied with `shutil.copyfile`.
- A file that exists with different contents was reported as "does not exist". That case now has its own message: `File <installed path> differs from <src>`.

The issue lists the Slurm and Kubernetes installers, but since #885 the check lives in `File.is_installed`, so that is the only place that needs the change.

## Test Plan
- Added `test_is_installed_binary_file` and `test_is_installed_missing_file`, and the existing content-mismatch test now checks the message. Without the fix, the binary test fails with `UnicodeDecodeError` and the mismatch test gets the "does not exist" message.
- `uv run --locked --extra dev pytest`: 1960 passed, 5 skipped.
- `uv run --locked --extra dev pre-commit run --files` on the two changed files: all hooks pass.

## Additional Notes
`read_text()` also normalizes line endings, so two files that differed only in CRLF vs LF used to count as the same. With a byte comparison they no longer do, which seems right for checking whether the copy made by `copyfile` is still identical to the source.


podkidyshev (approved):

podkidyshev:
@MaxFreedomPollard thank you for contribution, it's a good change 👍

